Issue #6, opened 2020-10-20
Set up Azure Blob storage accounts, containers


ryanhofdotgov:
Using IaC approach determined by #1, set up Azure Blob storage accounts, one per participating state.

PR #32, opened 2020-10-28
ARM template, script for creating per-state blob storage

ryanhofdotgov:
Addresses #6 – barebones implementation; does not address authentication, network restrictions, private networks, etc.


kingcomma (approved):
👍

Issue #7, opened 2020-10-20
Set up per-state databases


ryanhofdotgov:
Per IaC approach in #1, create a database for each participating state.

PR #42, opened 2020-11-02
IaC for per-state databases 

ryanhofdotgov:
Addresses #7. ARM templates for key vault, PostgreSQL server. Logical databases have to be established outside ARM; use re-runnable PostgreSQL DDL for role and database creation.


kingcomma (approved):
Maybe add the psql client application to the list of prerequisites? It sounds like the installation process can be [a bit more involved on Windows](https://www.compose.com/articles/postgresql-tips-installing-the-postgresql-client/#windows10) than macOS/Linux.

Otherwise, looks 👍

ryanhofdotgov:
Good catch! Right now the script would only run on Windows via WSL, so they would be able to install it there with `apt`. At some point, we'll do the [more complex thing in this ARM example](https://github.com/Azure/azure-postgresql/tree/master/arm-templates/ExampleWithStartupScript), where ARM is directed to spin up a temporary Ubuntu VM to apply the DDL.

And then maybe just migrating to PowerShell :/

Issue #58, opened 2020-11-03
Hello world app


cjhskippy:
Webservice that is automatically deployed that *could* represent metrics application (but has not content for now)

- [x] Website exists that can be browsed to
- [ ] Incorporated into CI (updating the master branch changes website content)

PR #77, opened 2020-11-10
Barebones dashboard app

kingcomma:
First step of #58. Barebones "hello, world" implementation — manual deployment, no deployment slots or CI/CD process.


doc/dashboard.md
```diff
@@ -0,0 +1,64 @@
...
+```
+    az webapp list --query "[?starts_with(name,'piipan-dashboard')].name" --output tsv
+```
+
+Then, get the app-level deployment user:
+```
+    az webapp deployment list-publishing-credentials -n <app_name> -g piipan-resources --query publishingUserName --output tsv
+```
+
+With that information, set the remote URL on the repo. Since app-level usernames being with a dollar sign, use single quotes or a backslash to escape:
```
ryanhofdotgov, line 44:
```suggestion
With that information, set the remote URL on the repo. Since app-level usernames begin with a dollar sign, use single quotes or a backslash to escape: 
```

ryanhofdotgov (changes requested):
Nice! One thing we should do before moving on is configure something to prevent public access per the [before-you-ship guide](https://before-you-ship.18f.gov/infrastructure/sandbox/). Either restrict to the GSA network block or set up some rudimentary access control.

kingcomma:
I updated the ARM template for the dashboard app to add an access restriction for the GSA network block. This restricts traffic to the public URL for the app.

A useful note from [Microsoft's docs](https://docs.microsoft.com/en-us/azure/app-service/app-service-ip-restrictions):

> When there are one or more [access restriction] entries, there is then an implicit "deny all" that exists at the end of the list.

ryanhofdotgov (approved):
👍

PR #78, opened 2020-11-12
Create per-state managed identities and db roles

ryanhofdotgov:
- Set up PostgreSQL AD admin group
- Create managed identity for each state
- Create managed roles for each state db
- Grant managed role required privileges
- Managed role will need to use SET ROLE <DBOWNER> to apply schema changes

### DIFF
--- a/iac/create-databases.bash
+++ b/iac/create-databases.bash
@@ -1,7 +1,9 @@
 #!/bin/bash
 set -e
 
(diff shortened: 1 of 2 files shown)


ryanhofdotgov:
Result of research done for #9

kingcomma (approved):
👍

PR #99, opened 2020-11-23
Refine CI/CD process and documentation

kingcomma:
- Creates a service principal account as part of IaC, only if one does not already exist
- Moves basic `siteConfig` property for App Service ARM template to a variable
- Clarifies current git workflow
- Documents CircleCI configuration and environment variables

Would appreciate a close look at the service principal creation in `iac/create-resources.bash`. Running the create command on an existing service principal will perform a "patch" operation that resets the password even if no changes are made. In that case, the password would need to be updated in the CI/CD tool every time the IaC script is run. The script avoids this by first checking if the service principal exists. If it does exist, the script does nothing — implicitly assuming the correct role/access has been configured.

Would also appreciate a gut check on how key vault is used to store the service principal credentials.


ryanhofdotgov:
Here's a few thoughts:
* It might be good to keep any CircleCI-specific/18F-specific setup separate from what is going to be applicable to USDA.
* We probably don't need to actually persist the service provider credentials in Key Vault. If a human has to be involved to set up CircleCI with it, you could just have that (manual) process snag the password at creation (or patch). 
* Similarly, the PostgreSQL password I set in KeyVault now currently can be removed in the script after PostgreSQL set up concludes. I had just left in there to help debug things before I had figured out AD managed identities. (I'll remove this PostgreSQL secret in a forthcoming PR.)
* Another reason for pulling it out of main IaC, is that at some point, an admin would need to rotate deployment credentials separate from re-deploying all the infrastructure.
* Depending on what we decide are next steps, this might be moot, but [passing sensitive data/secrets as command-line parameters should be avoided](https://unix.stackexchange.com/questions/439497/is-there-a-way-to-pass-sensitive-data-in-bash-using-a-prompt-for-any-command/439519), as other processes can see them in the process listing when the script is executing. I do [some trickery here in the existing use of Key Vault](https://github.com/18F/piipan/blob/48ba9506768328b221e1fcacd1b45dbace802997/iac/create-resources.bash#L72) to avoid that. That's the main reason I had Key Vault involved at all for the PostgreSQL set up is that if the secret is there, the ARM template deployer can access it itself. No need to pass it as a parameter as a CLI argument or figure out how to get it into a parameter.json to pass.

kingcomma:
Thanks! Here's my plan:

- Create a new `create-service-principal.bash` script that handles service principal creation and is called from the main IaC. This script will still only create the service principal if it does not already exist. The goal being to avoid the need for an admin to update any integration credentials every time the infrastructure is updated.
- Remove the use of key vault for service principal credentials entirely.
- Add a new doc for service principals that addresses how they are used, how to manually create one using the bash script, and how to grab/rotate credentials.
- Clean up any other USDA/18F documentation overlap I introduced with this PR.

I think that should address everything. What do you think?

ryanhofdotgov:
Kind of an odd question, but is there a use case/motivation for calling `create-service-principal` from `create-resources`?

kingcomma:
I think the main advantage is the ability to keep our service principal(s) in sync with the infrastructure without a lot of manual fiddling. For example, if we expand the infrastructure to include an additional resource group like `piipan-functions`, we may want to create that group and add it to the scope of our service principal(s) in one go.

I also think there is a small advantage when setting up a new environment from scratch. The benefit here is admittedly limited by the need to manually snag credentials before setting up/adjusting any integrations.

All that said, those advantages are probably only gained if/when we have multiple service principals in play. Since we only have the need for a single service principal at this point I may be thinking too far down the road.

ryanhofdotgov:
OK. Sounds good. We haven't really tried to wrangle all the output the script sends to stdout, and really just expect that if it doesn't exit with error code, everything is good. So trying to find one line of output (e.g., service principal password) is fairly hard.

If I was creating everything from scratch, I'd run the giant script to create resources (ignore all its stdout, just look for successful exit code), then just manually run the SP script by itself, and easily grab the (newly reset) password to set up CI/CD. So in a sense, putting it in the main script is just to centrally self-document that this credential exists.

I have gotten the impression that our partner agency has no CI/CD solution for us to use and getting one in place will take many, many months. If that end's up being the case, then we should rip it out of the central script so we aren't leaving unused service principals lying around.

kingcomma:
Updated with the following edits:

- Moved service principal creation to `create-service-principal`, which can be called manually
- Added service principal documentation
- Kept service principal creation in `create-resources`, but reduced to two lines that can be easily ripped out later in a new environment

ryanhofdotgov (commented):
Nice! Just a couple of very minor things:
1. We can rip out the references to Key Vault in `ci-cd.md` and link to `service-principals.md` instead, right?
1. Missing newline at end of `create-service-principal.bash` (won't break anything, but GitHub is kind enough to visually flag it)

Thanks for adding the functions resource group to the scope – I'll modify CI/CD stuff for our Functions after #8

kingcomma:
Good catch, thanks!

ryanhofdotgov (approved):
Awesome!

PR #100, opened 2020-11-24
IaC for Azure Function, etc. for ETL

ryanhofdotgov:
Partially addresses #8 

Sets up new resource group for Function Apps, enables Event Grid for the subscription. For each state, set up:
* Storage account for a Function App
* a Function App (with managed identity)
* Function for ETL
* system topic for blob creation (e.g., PII upload)
* event subscription to connect topic events to Function


ryanhofdotgov:
@kingcomma `create-resources` is in [need of some reorganization I think](https://github.com/18F/piipan/issues/101), but I'd like to defer until the next sprint.

ryanhofdotgov:
Yesterday's test was (eventually) successful:
1. Deleted all our resource groups
1. Ran `create-resources`, but forgot to connect to VPN, and it stopped with error on PostgreSQL config.
1. Ran script again, it stopped with error on publishing Function to Function App for 2nd example state. It appears that `az functionapp create` returns before the resource is fully provisioned. Based on my frequent work with that command this week, it seems rare enough event to just address it in our docs and re-run the script if it ever occurs.
1. Ran `create-resources` again; it completed successfully.

etl/host.json
```diff
@@ -0,0 +1,11 @@
+{
+    "version": "2.0",
+    "logging": {
+        "applicationInsights": {
+            "samplingExcludedTypes": "Request",
```
kingcomma, line 5: Just a heads up, according to Azure/azure-functions-core-tools#1900 looks like this setting should be under `sampleSettings` and named `excludedTypes`. This is the default output by `func init` and doesn't seem to be an active issue — I'm mostly just flagging so we know functions core tools may be a little buggy in places.

kingcomma (approved):
Nice work!

ryanhofdotgov:
Thanks! You probably saw it alluded to in the Giant Engineering Slack thread, but will be followed up with #103 at some point.

Issue #108, opened 2020-12-01
Continuous integration: verify ETL successfully loads per-state information


cjhskippy:
Follow-up from #31

- [x] Setup mock databases for per-state API that are built with each commit / test suite run
- [x] Unit tests automatically run with each commit to master/main branch
- [x] Unit test covers at least 90% of ETL code

PR #119, opened 2020-12-04
Initial and basic ETL process

ryanhofdotgov:
Stripped this down to get this out the door; there's a fair amount of follow-up:
- no incorporated tests, no CI/CD, need to fold in some manual steps into IaC
- no managed identity

Will address in #108 and in other issues TBD.


ryanhofdotgov:
@kingcomma And we have our first codeclimate complexity error. Is there a way to make this a warning instead of a hard fail until we get a sense of what the config should be?

ryanhofdotgov:
I couldn't find a way to get Code Climate to do what I want (run checks, but don't fail the PRR if the checks find maintainability "issues"), but I only dug around a little bit. Not sure what we want to do as a process going forward, but for this PR, I navigated to the the 2 Code Climate complexity issues and marked them as `wontfix`.

kingcomma:
I also can't figure out a clear way to get Code Climate to do what we want. [This documentation](https://docs.codeclimate.com/docs/workflow#status-updates) suggests you can configure which maintainability checks "can fail your pull requests," but links off to [more documentation](https://docs.codeclimate.com/docs/advanced-configuration#section-disabling-maintainability-checks) that suggests checks can only be enabled or disabled.

I don't want these checks to hold us up until we identify agreed upon thresholds. My thinking is I should (1) create a new PR that adds a `.codeclimate.yml` file which disables all checks and (2) add an issue for following up on the checks when we have a better handle of things.

Regardless, I'll push forward with reviewing this PR.

ryanhofdotgov:
Perfect – thanks!

doc/etl.md
```diff
@@ -0,0 +1,55 @@
...
+```
+
+Set the blob storage connection string in the `BlobStorageConnectionString` environment variable, on a per-state basis, for the Function App. Use an access key indicated by the portal for the state storage account (e.g., specific-storage-account → Settings → Access keys → key1)
+```
+az functionapp config appsettings set --name function-app-name --resource-group piipan-functions --settings BlobStorageConnectionString="…"
+```
+
+Set the database connection string in the `DatabaseConnectionString` environment variable, on a per-state basis, for the Function App. Use the ADO.NET value indicated by the portal for the `participants-records` cluster, modifying the string to reflect the specific state database (e.g, `ea`, `eb`, `ec`, etc.).
+```
+az functionapp config appsettings set --name function-app-name --resource-group piipan-functions --settings BlobStorageConnectionString="…"
```
kingcomma, line 42:
```suggestion
az functionapp config appsettings set --name function-app-name --resource-group piipan-functions --settings DatabaseConnectionString="…"
```

kingcomma (commented):
Looks great — just one small tweak in the documentation.

ryanhofdotgov:
Good catch!

kingcomma (approved):
👍

Issue #120, opened 2020-12-04
Enable code maintainability checks


kingcomma:
As software maintainers, we want to enable automatic code maintainability analysis on our repository to ensure our software is as maintainable, extendable, and fixable as possible and to improve developer experience. Code Climate is already in use and offers this analysis, but individual checks need to be manually enabled in `.codeclimate.yml`. Enable relevant checks and, if possible, configure any checks which should display warnings but not cause PRs to fail.

Related to conversation in #119.

We will have to revisit when we migrate to a wholly partner-hosted environment.

PR #123, opened 2020-12-09
Disable Code Climate maintainability checks

kingcomma:
Disables Code Climate maintainability checks per conversation in #119. Will be readdressed with #120.


ryanhofdotgov (approved):
👍

PR #124, opened 2020-12-09
Restructure `etl` directory

ryanhofdotgov:
Moving towards something like this [widely adopted .NET project structure](https://gist.github.com/davidfowl/ed7564297c61fe9ab814) as we refine our monorepo approach. `bin` is in the .gitignore for the default .NET project, so moving to `tools`.

Deferring restructuring of the top-level `doc` directory for now, but I'm imagining moving subsystem docs to their folders and creating a top-level README to point down to them.

I figure we can get consensus between us for this subsystem, then circle back on `dashboard`, etc.


kingcomma:
This approach sounds great to me. Having a structure that's both fairly well formalized and not intensely opinionated feels like the right mix.

A couple questions:

- Specifically for this PR — I think the paths in `create-resources` and `dependabot.yml` need to be updated to account for this new structure?

- More generally — it seems like we're establishing a namespace convention of `Piipan.{Subsystem}`. That seems obvious and good to me, but wanted to say it out loud in case I'm drawing an incorrect conclusion.

ryanhofdotgov:
Great catch! Fixed.

Yeah, I figured `Piipan` was unique enough to start our namespaces with, and more esthetically pleasing than  `Gsa` or `Tts`. 😄 

And after browsing a lot of [Microsoft's .NET projects](https://github.com/dotnet) and the [CsvHelper library ](https://github.com/JoshClose/CsvHelper), it looks like folks derive their directory names under `src` and `tests` from their namespaces. 

If our subsystems start having multiple packages, then we will definitely need to start looking at those .sln files. Also hoping dependabot, etc. will be able to parse those files instead of having to specify every package-level directory.

kingcomma (approved):
👍

ryanhofdotgov:
Also updated Snyk config, though across all my projects I used it on, it does not have a good signal-to-noise ratio. And I just discovered that if a package file (e.g., `.csproj`) is moved, Snyk requires its corresponding project to be deleted and created new, losing any history.

PR #127, opened 2020-12-10
Open API specifications for matching APIs

kingcomma:
@ryanhofdotgov wanted to get a draft of the API specs in front of you to make sure we have the same expectations before moving forward with a full review.

This adds specs for both the per-state matching API (`match/docs/state_api/openapi.yaml`) and the orchestrator API (`match/docs/match_api/openapi.yaml`). Since these tie together so closely it made sense to me to be able to evaluate each in the context of the other.

There are more than a few of my own assumptions at play:

- For security reasons, it's better to use `POST` rather than `GET` to avoid logging any query parameters with PII.
- Matches need to return the full `PiiRecord` rather than a simple "yes"/"no" so that states have enough information for follow-up communication.
- The response format in the state API spec is designed to return a _single_ match (or no match), based on the assumed exact match algorithm.
- The structure and syntax of the request and response formats will need to change as we gather more information through research. For now I've tried to rough in some of the concepts we've seen, like nesting request query data within its own object.
- Requests will send the individual name components and the backend will handle the logic for constructing a "full name."
- The `dob` value uses ISO 8601 for formatting per the [OpenAPI spec](https://swagger.io/docs/specification/data-models/data-types/#format) and [18F's API standards](https://github.com/18F/api-standards#use-a-consistent-date-format). This is different than the specified date format for bulk uploads in `doc/csv/import-schema.json`.


ryanhofdotgov:
1. 👍 for `POST`
1. Yeah, I may have said it was an oracle (i.e., yes/no results) at times as over simplification. If only exact matches were permitted and never anything different, we *might* be able to get away with that, but the API needs to allow for change along that dimension.
1. I think the state-level and orchestrator-level both need to support returning one or more results, just in case the matching algorithms change, or search restrictions are lifted for certain roles (e.g., investigator).
1. That makes sense; it will be important for the orchestrator API to be backwards compatible for as long as it can be once it's in production use. The per-state API is less sensitive, but applying the same approach makes upgrades easier.
1. Yes, that gives us the most flexibility. 
1. Yes, using ISO 8601 makes sense. I initially used it for the CSV format as well, but then Excel would always translate it to the MM/DD/YYYY format on save if you didn't take explicit steps on import (at least for workstations configured with the English US locale default). The CSV file is for machine-to-machine, but thought it might be a useful state user accommodation to deviate from ISO 8601. The CSV format is definitely worth revisiting as we get farther along with the state user research.

kingcomma:
Thanks for weighing in! I'm marking this ready for review with the following edits:

- State-level API now returns a list of one or more matches.
- Repeated components across the two specs have been moved to separate files and referenced within the specs themselves.

ryanhofdotgov:
2 popular OpenAPI tools, [Postman](https://www.postman.com/downloads/) and [openapi-generator](https://github.com/OpenAPITools/openapi-generator) don't seem to like the relative path reference (e.g., `../models/pii-record.yaml`) in `match-api/openapi.yaml`. Even moving those references to the same folder didn't work.

Not sure if I've ever tried to split a spec across multiple files before. Quick google finds work-arounds like: https://github.com/javanile/yamlinc

kingcomma:
You're right — I made a false assumption that if the [OpenAPI extension for VS Code](https://github.com/42Crunch/vscode-openapi) could handle the external references, others could too.

It looks like a common solution is using [swagger-cli](https://github.com/APIDevTools/swagger-cli) to bundle external references into a single file. The tool also has a `validate` command that revealed a few issues with the YAML structure. Those have been fixed.

I've also incorporated several approaches from [this post](https://mux.com/blog/an-adventure-in-openapi-v3-api-code-generation/) that should improve our experience, especially if we use these specs to generate code.

This structure may be overkill for two APIs with only a single path each, but I think it will help avoid errors if/when the spec evolves.

ryanhofdotgov (approved):
Cool! Was able to get artifacts generated by both Postman and openapi-generator-cli 👍

PR #128, opened 2020-12-10
Configure CircleCI to require successful testing before deploy

kingcomma:
Adjusts `.circleci/config.yml` to include the `test` job in the `requires` list for each deploy job. Fixes an issue with my original implementation that ran the test job but did not require it for before starting a deploy job.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -92,18 +92,21 @@ workflows:
       - build
       - deploy-dev:
           requires:
+            - test
             - build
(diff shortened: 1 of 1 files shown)


kingcomma:
@ryanhofdotgov I know you are also doing some CI/testing work. Feel free to defer this until after you've wrapped your work up, or incorporate it if you'd prefer.

ryanhofdotgov (approved):

PR #133, opened 2020-12-14
Complete automated unit testing of etl code

ryanhofdotgov:
- Reactor system under test to permit dependency injection of database
- 98% code coverage, though probably not worth the effort. Better to get 90%+ coverage through functional testing (i.e., with real PostgreSQL instance, etc.).

### DIFF
--- a/etl/src/Piipan.Etl/BulkUpload.cs
+++ b/etl/src/Piipan.Etl/BulkUpload.cs
@@ -25,13 +25,6 @@ namespace Piipan.Etl
     /// </summary>
     public static class BulkUpload
     {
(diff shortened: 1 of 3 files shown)


kingcomma (commented):
Looks great! Just one thing: could the `Azure.Storage.Blobs` and `Newtonsoft.Json.Linq` imports can be removed from `BulkUpload.cs` (and perhaps from the project references too) now that you've dropped the logging?

kingcomma (approved):

ryanhofdotgov:
Good catch – a casualty of editing outside of Visual Studio Code... At some point, we can look into extending our CI/CD process #135

PR #134, opened 2020-12-14
Remove redundant CSVHelper package

ryanhofdotgov:
Since we include Piipan.Etl, don't need to re-specify CSVHelper as a dependency here. Doing so can cause mismatches, Dependabot failures.


kingcomma (approved):
👍

PR #137, opened 2020-12-15
Restructure dashboard and update deployment

kingcomma:
Some cleanup to reflect new approaches and clarify them for future maintainers.

- Restructures `dashboard` according to project structure discussed in #124.
- Removes unused `staging` and `develop` deployment slots in dashboard ARM template.
- Removes unused `deploy-staging` and `deploy-dev` deployments from CircleCI config.
- Updates dashboard documentation to reflect current deployment approach.

Propose we defer the failed code coverage diff check on this until dashboard development resumes.


ryanhofdotgov (approved):
Nice!

ryanhofdotgov:
I just remembered that we need to fix paths in dependabot, Synk. I have a directory to add too, so I'll do that in a separate PR now. And I'll open up a ticket to write up a doc on all the things we have to configure manually when we add/change a source or test directory.

PR #158, opened 2020-12-15
Enable repeatable package restores via lock file

### LINKED ISSUE #153: Pin/lock versions of dependencies
As best practice to permit build reproducibility across local environments and CircleCI, pin/lock the dependencies we are using in our sub-systems:
https://devblogs.microsoft.com/nuget/enable-repeatable-package-restores-using-a-lock-file/

ryanhofdotgov:
Closes #153.

References:
- [Enable repeatable package restores using a lock file](https://devblogs.microsoft.com/nuget/Enable-repeatable-package-restores-using-a-lock-file/)
- [Using a NuGet Lock File for Reproducible Builds](http://blog.ctaggart.com/2019/03/using-nuget-lock-file-for-reproducible.html)

### DIFF
--- a/dashboard/src/Piipan.Dashboard/Piipan.Dashboard.csproj
+++ b/dashboard/src/Piipan.Dashboard/Piipan.Dashboard.csproj
@@ -3,7 +3,7 @@
   <PropertyGroup>
     <TargetFramework>netcoreapp3.1</TargetFramework>
     <RestorePackagesWithLockFile>true</RestorePackagesWithLockFile>
-    <RestoreLockedMode>true</RestoreLockedMode>
+    <RestoreLockedMode Condition=<"'$(ContinuousIntegrationBuild)' == 'true'">true</RestoreLockedMode>
   </PropertyGroup>
 
 </Project>
--- a/dashboard/src/Piipan.Dashboard/Piipan.Dashboard.csproj
+++ b/dashboard/src/Piipan.Dashboard/Piipan.Dashboard.csproj
@@ -2,6 +2,8 @@
 
   <PropertyGroup>
     <TargetFramework>netcoreapp3.1</TargetFramework>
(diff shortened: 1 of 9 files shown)


dashboard/src/Piipan.Dashboard/Piipan.Dashboard.csproj
```diff
@@ -2,6 +2,8 @@
 
   <PropertyGroup>
     <TargetFramework>netcoreapp3.1</TargetFramework>
+    <RestorePackagesWithLockFile>true</RestorePackagesWithLockFile>
+    <RestoreLockedMode>true</RestoreLockedMode>
```
kingcomma, line 6: Sorry, saw #159 before this and added a [relevant comment](https://github.com/18F/piipan/pull/159#discussion_r544351469) there.

My same issue applies, but the steps to recreate the issue on this branch are simpler:

```
cd dashboard/src/Piipan.Dashboard

# Any package would cause this same issue
dotnet add package Newtonsoft.Json
```

Changing this line to `<RestoreLockedMode Condition="'$(ContinuousIntegrationBuild)' == 'true'">true</RestoreLockedMode>` avoids the error, but would require configuring `ContinuousIntegrationBuild` in CircleCI.

ryanhofdotgov: Ah... thanks for catching this. I'll dig into it. This whole feature is implemented strangely in NuGet/dotnet CLI as they didn't have the feature from the get-go in NuGet, and rolled it in as part of a minor version update.

The feature announcement blog post also seemed to have a weird perspective – that in your local dev environment you wouldn't want pinned versions; i.e., devs would be happy having the library versions change underneath them behind the scenes.

kingcomma (approved):

Issue #167, opened 2020-12-29
Add Managed Identities to the per-state API


cjhskippy:
Utilize Azure's managed identities to improve the security of the per-state match API.

Note: the other system APIs may still need updates to incorporate managed identities.

AC:
- [ ] Match API uses managed identities

PR #168, opened 2020-12-29
Initial per-state API functionality

### LINKED ISSUE #26: Per-state API to test if a person is an existing enrollee
As a state user, I want to see if a person is already enrolled in SNAP in another state.

Interacts with #95 to orchestrate and call for each state.

Note for user workflow: Need to do before making certification decisions, but want to wait until confident in the name, SSN, DOB

AC:
 - [x] #140 - Create detailed specifications for API
 - [ ] Provide name, DOB, SSN, find out if these match against any existing enrollees.
 - [ ] Returns a "yes" if an existing match, defined at this stage as EXACT match of first name, last name, middle name (all separate fields), SSN, DOB.
 - [ ] Returns "no" if the person isn't already existing

kingcomma:
Closes #26 and #167.

Adds per-state match API functionality:
- IaC (ARM templates) for creating state-specific Function apps
- Function app for matching API (`match/src/Piipan.Match.State`)
- Unit tests for API app (`match/tests/Piipan.Match.State.Tests`)
- Instructions for running locally and publishing (`doc/state-match.md`)

Will be expanded later to include:
- Event publishing #170
- Error/validation messages returned with 400 responses #171
- Validation against OpenAPI spec
- Viable local development option, presumably using a docker-based postgres image


doc/state-match.md
```diff
@@ -0,0 +1,95 @@
...
+1. Run `func start` or, if hot reloading is desired, `dotnet watch msbuild /t:RunFunctions`.
+
+## Manual deployment
+
+These instructions assume that the [piipan infrastructure](iac.md) has been established in the Azure subscription. Running the IaC will set up a Function app for each participating state. Each app is associated with a storage account, an application insights instance, a state-specific managed identity, and a state-specific database. All settings necessary for connecting the app to the state database are automatically stored in the app's configuration.
+
+### Database setup
+
+Before the managed identity can execute commands against the state database the `per-state.sql` DDL must be run while using the state-specific database owner role. The [bulk upload](etl.md) process currently uses the `postgres` user to run the DDL, creating tables in the `public` schema that are not accessible to the state-specific admin roles (e.g., `eaadmin`). Running the DDL as the state-specific database owner is a temporary step until the ETL process is updated to use a managed identity.
```
ryanhofdotgov, line 35: Let's brainstorm on how this should work.

Originally I had envisioned several db roles tied to corresponding managed identities: admin, reader, writer. I got bogged down establishing managed identities and stopped at the one (which may be a good place to stop anyhow). When the DDL was loaded, explicit read and/or write privs were to be granted to these roles on the specific tables, etc. The admin role would still `SET ROLE...` to do superuser-ish things, like altering tables, etc – kind of like `su`. You wouldn't set the role to do more basic operations.

Some options:
* Add CI/CD account to piipan-admin group, which is the PostgreSQL AD admin. I don't like resorting to the cluster-level superuser for state-specific actions and giving the service principal more privileges, but it may be CI/CD may be moot once we get to the partner environment.
* Some type of scheme using an Azure Function to run DDL as the state-specific admin.
* Lazily update database DDL in the ETL function. This would be a finding under DoD compliance rules, probably avoid it here if we can eventually get to a place where functions don't connect with role that can alter the DDL.

doc/state-match.md
```diff
@@ -0,0 +1,95 @@
...
+### App deployment
+
+Before deploying the app for the first time, switch to the project directory and fetch settings from Azure:
+
+```
+cd match/src/Piipan.Match.State
+func azure functionapp fetch-app-settings {app-name}
+```
+
+This will create a file named `local.settings.json` containing app settings. Some settings are required by the deployment process (e.g., `FUNCTIONS_WORKER_RUNTIME`).
```
ryanhofdotgov, line 78: It's surprising that we have to discover and stumble over what should be basic functionality – I think the correct thing is to pass `--dotnet` or `--csharp` to the publish command. I was originally tripped up on this because `--dotnet` wasn't in the list of suggested parameters, but I think that is a bug. Here's [the relevant source code](https://github.com/Azure/azure-functions-core-tools/blob/2f5369afbdeb3e9ded70dccc93388f2b186b2093/src/Azure.Functions.Cli/Helpers/GlobalCoreToolsSettings.cs#L44). Error message that omits `--dotnet` is a few lines above that.

match/src/Piipan.Match.State/Api.cs
```diff
@@ -0,0 +1,149 @@
...
+            var sql = "SELECT upload_id, first, last, middle, dob, ssn, exception FROM participants " +
+                        "WHERE ssn=@ssn AND dob=@dob AND last=@last " +
+                        "AND " + (p.first == null ? "first IS NULL" : "first=@first") + " " +
+                        "AND " + (p.middle == null ? "middle IS NULL" : "middle=@middle") + " " +
+                        "AND upload_id=(SELECT id FROM uploads WHERE created_at = (SELECT MAX(created_at) FROM uploads))";
+
+            return (sql, p);
+        }
+
+        internal async static Task<string> ConnectionString(ILogger log)
```
ryanhofdotgov, line 113: I'll work on some reusable code that can be swapped in here. Ideally we want all of the connection string elements to be set at publish time, except the password.

I'm also wondering if all of these URLs we see in the Azure tutorials are exactly the same in Azure Government.

match/src/Piipan.Match.State/Converters.cs
```diff
@@ -0,0 +1,33 @@
+using System;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Converters;
+
+namespace Piipan.Match.State {
+    public class DateTimeConverter : IsoDateTimeConverter
```
ryanhofdotgov, line 6: Even for these tiny, pretty obvious classes we should try and be consistent about having at least class-level summary comment. If only for the generated docs that don't include source.

match/src/Piipan.Match.State/MatchRequest.cs
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    public class MatchQuery
+    {
+        [JsonProperty("last", Required = Required.Always)]
+        public string Last { get; set; }
+
+        [JsonProperty("first", NullValueHandling = NullValueHandling.Ignore)]
+        [JsonConverter(typeof(NullConverter))]
+        public string First { get; set; }
```
ryanhofdotgov, line 24: Did [nullable reference types come up](https://docs.microsoft.com/en-us/dotnet/csharp/nullable-references) in the context of the JSON library? Only asking because I was lead down that path following the CsvHelper tutorial and these classes are very similar.

.circleci/config.yml
```diff
@@ -25,11 +25,16 @@ jobs:
...
+      - run:
+          working_directory: match/tests/Piipan.Match.State.Tests
+          name: Test State Match
+          command: dotnet test $MSBUILD_PROPS --collect:"XPlat Code Coverage" -- DataCollectionRunSettings.DataCollectors.DataCollector.Configuration.Format=lcov
       - run:
           name: Upload Code Climate coverage report
           command: |
             ./cc-test-reporter -d format-coverage ./dashboard/tests/Piipan.Dashboard.Tests/TestResults/**/coverage.info --input-type lcov -p /root/project -o /tmp/codeclimate.dashboard.json 
             ./cc-test-reporter -d format-coverage ./etl/tests/Piipan.Etl.Tests/TestResults/**/coverage.info --input-type lcov -p /root/project -o /tmp/codeclimate.etl.json
+            ./cc-test-reporter -d format-coverage ./match/tests/Piipan.Match.State.Tests/TestResults/**/coverage.info --input-type lcov -p /root/project -o /tmp/codeclimate.match.state.json
```
ryanhofdotgov, line 37: Let's brainstorm on next steps to facilitate automated builds and deployments. I need to do this for ETL as well.

iac/arm-templates/function-state-match.json
```diff
@@ -0,0 +1,162 @@
...
+        "functionWorkerRuntime": "dotnet",
+        "storageAccountType": "Standard_LRS",
+        "applicationInsightsTag": {
+            "[concat('hidden-link:', resourceId('Microsoft.Web/sites', variables('functionAppName')))]": "Resource"
+        },
+        "identityName": "[concat(parameters('stateAbbr'), 'admin')]"
+    },
+    "resources": [
+        {
+            "type": "Microsoft.Storage/storageAccounts",
```
ryanhofdotgov, line 41: Can we reuse `arm-templates/function-storage.json`?

iac/arm-templates/function-state-match.json
```diff
@@ -0,0 +1,162 @@
...
+                        {
+                            "ipAddress": "159.142.0.0/16",
+                            "action": "Allow",
+                            "name": "GSA IP range",
+                            "priority": 100
+                        }
+                    ],
+                    "ftpsState": "Disabled",
+                    "appSettings": [
+                        {
```
ryanhofdotgov, line 96: There seems to be a fair amount of setting construction that I somehow avoided using the non-ARM approach. Do all of these have to be set explicitly; i.e., are these the defaults that would be set if these weren't present?

ryanhofdotgov (approved):
Nice job! A lot of moving parts coming together 👍 

None of my review comments hold up merging this in, but let's meet up to discuss the brainstorm items, and specific follow on tickets that will fully connect this subsystem to the ETL subsystem with fully automated database updates and function deployments.

PR #183, opened 2021-01-06
Update dependencies

ryanhofdotgov:
* for each tree, `dotnet list package --outdated`
* then for each, `dotnet add package ...`

Edit: Still had to run ` dotnet restore --force-evaluate` afterwards


kingcomma (approved):
I also found every approach I tried still required `dotnet restore --force-evaluate` after updating a package. Not relevant here, but _adding_ a new package with `dotnet add` appears to update the lock file without any subsequent commands.

Anyway, looks good 👍

Issue #150, opened 2020-12-15
Incorporate managed identities into ETL


ryanhofdotgov:
Initial ETL work in #119 deferred incorporation of managed identities – the Function is configured with the top-level database and blob credentials. 

Instead, use the per-state managed identity to fetch the blob and connect to the database.

ryanhofdotgov:
It ends up that Azure Storage Accounts, don't support retrieval based on a managed identity. To implement, you have to use an indirect method and generate a SAS URL. That work has been split off into https://github.com/18F/piipan/issues/213, which can be deferred to a later date, possibly after our MVP milesone.

PR #199, opened 2021-01-10
Bind database connection string at provision time

ryanhofdotgov:
Partially addresses #150.

For the ETL functions, bind the corresponding database connection string to the function via the DatabaseConnectionString environment variable. The Password placeholder will be replaced at runtime by the ETL function code with the a service token in a follow-on PR.


kingcomma (approved):
👍

PR #200, opened 2021-01-11
Add hint for soft delete error

ryanhofdotgov:
Even when you delete a key vault, Azure will keep it around for a while unless you forcefully purge it ahead of its retention schedule.


kingcomma (approved):
👍

PR #205, opened 2021-01-11
Fix local.settings.json error, use --dotnet

ryanhofdotgov:
Should resolve need to have a local.settings.json file in the working directory when publishing initial Azure function in IaC.

Still need to make similar changes to CI/CD script, etc.


kingcomma (approved):
👍

PR #207, opened 2021-01-11
fix broken link to docs

echappen:
Link to table schema docs 404'ed. This may be the intended link.


kingcomma (approved):
Good catch — thanks!

Issue #198, opened 2021-01-10
Remove unnecessary SET ROLE directive


ryanhofdotgov:
With changes done in PR #197, this `SET ROLE` directive should be unnecessary:
https://github.com/18F/piipan/blob/5620c35abd850e09e8f57a33a749323ad7f54ba0/match/src/Piipan.Match.State/Api.cs#L138

Remove and verify.

PR #210, opened 2021-01-11
Follow up to initial state match API work

kingcomma:
Adds some follow up work from #168:
- Cleans up documentation
- Removes unneeded `SET ROLE` directive (#198)
- Brings DB connection approach into sync with #199

### DIFF
--- a/iac/arm-templates/function-state-match.json
+++ b/iac/arm-templates/function-state-match.json
@@ -93,10 +93,33 @@
                     ],
                     "ftpsState": "Disabled",
                     "appSettings": [
+                        /*
+                            The following settings are required for Function apps per the MS documentation:
+                            - AzureWebJobsStorage
+                            - FUNCTIONS_WORKER_RUNTIME
+                            - FUNCTIONS_EXTENSION_VERSION
+
+                            See: https://docs.microsoft.com/en-us/azure/azure-functions/functions-infrastructure-as-code#function-app)
+                        */
                         {
                             "name": "AzureWebJobsStorage",
                             "value": "[concat('DefaultEndpointsProtocol=https;AccountName=', variables('storageAccountName'), ';EndpointSuffix=', environment().suffixes.storage, ';AccountKey=',listKeys(resourceId('Microsoft.Storage/storageAccounts', variables('storageAccountName')), '2019-06-01').keys[0].value)]"
                         },
+                        {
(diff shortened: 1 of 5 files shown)


iac/arm-templates/function-state-match.json
```diff
@@ -121,21 +151,13 @@
                             "name": "StateAbbr",
                             "value": "[parameters('stateAbbr')]"
                         },
-                        {
-                            "name": "ServerName",
-                            "value": "[parameters('serverName')]"
-                        },
                         {
                             "name": "AzureServicesAuthConnectionString",
```
ryanhofdotgov, line 155: I just discovered a few minutes ago (when I *finally* got managed identity working with ETL) that `AzureServicesAuthConnectionString` is required when using AzureServiceTokenProvider.GetAccessTokenAsync. Otherwise, you get an error, excerpt below:
```
...Tried the following 3 methods to get an access token, but none of them worked...
...Exception Message: Tried to get token using Managed Service Identity. Access token could not be acquired. Received a non-retryable error. MSI ResponseCode: BadRequest, Response: {"StatusCode":400,"Message":"SystemAssigned identity is not enabled for this site"...
```

kingcomma: Ah, yeah, it would have been helpful if I documented that a little better 😦. You can also pass `AzureServiceTokenProvider` a [connection string](https://docs.microsoft.com/en-us/dotnet/api/overview/azure/service-to-service-authentication#connection-string-support) at runtime, which potentially allows different connection approaches in different environments (e.g., passing `RunAs=Developer; DeveloperTool=AzureCli` to authenticate as your CLI user). That's what led me to thinking of this more as a custom setting rather than a _required_ one.

iac/create-resources.bash
```diff
@@ -397,10 +400,11 @@ while IFS=, read -r abbr name ; do
         resourceTags="$RESOURCE_TAGS" \
         identityGroup=$RESOURCE_GROUP \
         location=$LOCATION \
-        identityClientId=$client_id \
-        serverName=$PG_SERVER_NAME \
+        azAuthConnectionString=$az_serv_str \
```
ryanhofdotgov, line 403: Your template approach may be ultimately best; let's revisit when I migrate my Function app work to use your ARM template approach.

ryanhofdotgov (approved):
👍

PR #212, opened 2021-01-12
Use managed identities in BulkUpload

ryanhofdotgov:
Partially addresses #150 – blob access via managed identities in separate PR.

### DIFF
--- a/etl/src/Piipan.Etl/BulkUpload.cs
+++ b/etl/src/Piipan.Etl/BulkUpload.cs
@@ -6,9 +6,11 @@ using System.Data;
 using System.Data.Common;
 using System.Globalization;
 using System.IO;
+using System.Threading.Tasks;
 using CsvHelper;
 using CsvHelper.Configuration;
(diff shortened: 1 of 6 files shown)


kingcomma (approved):
👍 Code looks great!

Do you have a preference on resolving/deferring the snyk and code climate issues before merging?

ryanhofdotgov:
Let's defer the Code Climate issue until #192 – that's the 20 lines of code that resulted in a 0.8% decrease in code coverage. I had some unit test code that ran it, but wasn't easily able to write a test for anything meaningful. I want to avoid test code that just runs through the lines, without purpose. I discovered there's a significant amount of variability about what is the syntax of an "ADO.NET" connection string. Host vs Server, User Id vs Username, Port as a field or not, so it wasn't trivial to test that the fixture string round-tripped correctly through the method.

As far as the Snyk one goes, I've concluded Snyk's alerts are mostly unactionable by us. It's the same dependency in the Match API, and it is already at the latest version.

PR #233, opened 2021-01-20
Clarify use of .NET Core 3.1 for Piipan

ryanhofdotgov:
Supplements PR #232


echappen:
@ryanhofdotgov Glad we're using ADR's!

echappen (approved):

Issue #243, opened 2021-01-20
Figure out what CLI extensions the IaC script needs


ryanhofdotgov:
The AZ CLI will prompt users to install extension dependencies, when they are needed. For example, `az postgres` needs the `db-up` extension.

The problem is that the create-resources.bash script will be invoked by users who may not have those extensions installed. The AZ CLI may not be able to prompt for install in those circumstances. And that's not really the preferred state anyhow – we should have an explicit list of AZ CLI extensions and explicitly install them.

See [Slack thread](https://gsa-tts.slack.com/archives/C018P1ZJPEC/p1611180855045000?thread_ts=1611174743.028200&cid=C018P1ZJPEC) for more details, but the relevant error is:
```
The command requires the extension db-up.
 Unable to prompt for extension install confirmation as no tty available. Run 'az config set extension.use_dynamic_install=yes_without_prompt' to allow installing extensions without prompt.
```

- [x] Update create-resources.bash with explicit CLI extensions required
- [x] Add an explicit success message at the end of the script
- [x] Determine if `az postgres` is correctly returning an error code in this scenario
- [x] Re-run create-resources.bash and get Azure dev environment back to a known good state

PR #255, opened 2021-01-22
Better script error handling and ADR for scripts

ryanhofdotgov:
- -e flag doesn't on its own work with command substitutions, so several
  options were added to allow scripts to always immediately exit on error
- add explicit success message to end of IaC scripts

Foundational work for #243

### DIFF
--- a/docs/adr/0010-use-bash-for-scripting.md
+++ b/docs/adr/0010-use-bash-for-scripting.md
@@ -1,0 +1,26 @@
+# 10. Use bash for scripting
+
+Date: 2021-01-22
+
+## Status
+
+Accepted
+
+## Context
+
+We need a basic scripting tool to assist with deployment of code, automate tasks in CI/CD, and perform manual integration tests against our deployment environments.
+
+PowerShell and sh/bash are our primary options. bash is installed by default on macOS and in Linux distributions. sh/bash is available on Windows via the Windows Subsystem for Linux (WSL) and Cygwin. PowerShell ships with Windows 10 and is available for macOS and Linux.
+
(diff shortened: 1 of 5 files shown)


docs/adr/0010-use-bash-for-scripting.md
```diff
@@ -0,0 +1,26 @@
...
+
+PowerShell and sh/bash are our primary options. bash is installed by default on macOS and in Linux distributions. sh/bash is available on Windows via the Windows Subsystem for Linux (WSL) and Cygwin. PowerShell ships with Windows 10 and is available for macOS and Linux.
+
+In general, our implementation team does not have much experience with PowerShell, while sh/bash scripts are pervasive across 18F projects and our CI/CD tools (e.g., CircleCI, GitHub Actions).
+
+In some environments, `/bin/sh` will be a soft-link to `/bin/bash`. Other times it will be linked to other implementations (e.g., dash or zsh). It is often the case that non-trivial shell scripts that describe themselves as POSIX sh scripts will incorporate useful "bash-isms" which can fail when `/bin/sh` is not actually bash.
+
+## Decision
+
+For now, use bash (explicitly as `/bin/bash`, not `/bin/sh`), as our primary scripting language. Bash scripts should be written to support bash 3.2, which is the latest version of bash on macOS and
```
echappen, line 21: Maybe a typo at end of sentence

echappen (approved):

docs/adr/0010-use-bash-for-scripting.md
ryanhofdotgov: Good catch – fixed.

PR #257, opened 2021-01-22
Update resource group list that CI/CD can update

ryanhofdotgov:
- Add match API resouce group to CI/CD service principal
- Move invocation of create-service-principal to be closer
  to creation of resource groups


kingcomma (approved):
👍

Issue #214, opened 2021-01-12
Spike: State system authentication to Orchestrator API


cjhskippy:
Research spike to decide how to ensure that the orchestrator API can be accessed by state eligibility systems and not by unauthorized users.

- [x] Research + decide on recommended approach to authenticate Orchestrator API
- [x] Recommendations should support authentication mechanism that meets the AC in #28

PR #266, opened 2021-01-26
Initial orchestrator API implementation

### LINKED ISSUE #95: Orchestrator to test if a person is an existing enrollee
As a state user, I want to see if a person is already enrolled in SNAP in another state.

This story is to build the orchestrator - endpoint that client systems will hit that then passes on the requests to each state's API.

AC:
 - [ ] Provide name, DOB, SSN, find out if these match against any existing enrollees.
 - [ ] Returns a "yes" if an existing match
 - [ ] Returns "no" if the person isn't already existing
 - [ ] Returns the state(s) that match
   - [ ] If matches with more than one state, return each
 - [ ] Include some form of authentication control (does not have to be the final one)

kingcomma:
Initial implementation of the orchestrator API as an Azure Function App and IaC. Closely follows the approach of `Piipan.Match.State`. Closes #95.

Notes:
- Follows the existing [OpenAPI specs](https://github.com/18F/piipan/blob/main/match/docs/openapi.md).
- Does not address authorization of the orchestrator API with the per-state APIs.
- Authorization/access control for querying the orchestrator API will be addressed in  #214.


kingcomma:
You'll notice a good amount of code duplication between `Piipan.Match.State` and `Piipan.Match.Orchestrator` (e.g., `MatchResponse.cs`, `MatchRequest.cs`). This feels like a code smell, but it's the product of attempting to preserve the two APIs as decoupled microservices. I'm curious to have this approach and/or my thinking challenged.

match/docs/orchestrator-match.md
```diff
@@ -0,0 +1,45 @@
...
+- [Azure Functions Core Tools](https://docs.microsoft.com/en-us/azure/azure-functions/functions-run-local)
+
+## Summary
+
+An initial API for matching PII data across all participating states.
+1. JSON `POST` request that conforms to the [OpenApi spec](openapi.md) is sent to the orchestrator API endpoint.
+1. The `POST` event triggers a function named `Query` in the orchestrator Function App.
+    - If the request is not valid (malformed, missing required data, etc), the function returns a 400 response. Currently no error messaging is included in the response.
+    - If the request is valid, the function queries the per-state APIs for matches. A 200 response is returned containing any matching records.
+    - If there is an issue connecting to or querying any of the per-state APIs, the orchestrator returns a 500 response.
```
ryanhofdotgov, line 15: Not for this PR, but I think we will want to consider whether or not a failure of a state endpoint should actually result in a 500; i.e., What's the worse thing that can happen if 1 or 2 states out of 53 fail on a call? I'd probably set some failure threshold. Likewise, if a state fails to upload new PII for several days and the data is stale, should that cause this API to fail, or just discard any stale matches.

match/src/Piipan.Match.Orchestrator/Api.cs
```diff
@@ -0,0 +1,135 @@
...
+        }
+
+        internal static IEnumerable<string> ApiEndpoints()
+        {
+            const string StateApiEndpointStrings = "StateApiEndpointStrings";
+
+            IEnumerable<string> endpoints = JsonConvert.DeserializeObject<IEnumerable<string>>(
+                Environment.GetEnvironmentVariable(StateApiEndpointStrings));
+
+            return endpoints;
```
ryanhofdotgov, line 106: Is there a way to defensively parse these as a URI object at this layer of responsibility that we could pass later to MatchState? i.e., MatchState should operate on URIs not raw strings.

Not for this PR, but just thinking generally across our system – can we meaningfully validate service binding strings passed from the environment? e.g., reject strings for components outside our subsystem. Technically, it is input that we don't validate before using.

match/src/Piipan.Match.Orchestrator/Converters.cs
```diff
@@ -0,0 +1,56 @@
+using System;
```
ryanhofdotgov, line 1: You know this already, but just flagging this explicitly as good candidates for a commons library, similar to the managed identity logic ticket in https://github.com/18F/piipan/issues/192.

match/src/Piipan.Match.Orchestrator/MatchRequest.cs
```diff
@@ -0,0 +1,37 @@
...
+
+        public string ToJson()
+        {
+            return JsonConvert.SerializeObject(this, Newtonsoft.Json.Formatting.Indented);
+        }
+    }
+
+    public class MatchQuery
+    {
+        [JsonProperty("last", Required = Required.Always)]
```
ryanhofdotgov, line 19: Now that we've been thinking about it longer... I think there are a few API additions we'll need at some point:
* user identifier; i.e., on whose behalf is the state system or query tool is executing the API call for
* match identifier will need to be in the response

match/src/Piipan.Match.Orchestrator/Api.cs
```diff
@@ -0,0 +1,135 @@
...
+        /// <remarks>
+        /// This function is expected to be executing as a resource with query
+        /// access to the individual per-state API resources.
+        /// </remarks>
+        [FunctionName("query")]
+        public static async Task<IActionResult> Query(
+            [HttpTrigger(AuthorizationLevel.Anonymous, "post", Route = null)] HttpRequest req,
+            ILogger log)
+        {
+            var incoming = await new StreamReader(req.Body).ReadToEndAsync();
```
ryanhofdotgov, line 37: Farther down the road, we will need to log:
* what state is making the request
* what state user is making the request
* perhaps what system is making the call on the user's behalf

match/docs/orchestrator-match.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Binding to state APIs
+
+The orchestrator treats per-state APIs as backing services. When running the [IaC](../../docs/iac.md), per-state endpoints are compiled into a JSON list and saved as an environment variable for the orchestrator Function App.
```
ryanhofdotgov, line 19: Could you explicitly document that environment variable here and add a row to the list in iac.md? We should add a 3rd column to the iac.md table where we list what subsystems use what environment variables.

I'll go back and update our other subsystem docs.

ryanhofdotgov (commented):
Nice work! 👍 👍 

I've left a few thinking-out-loud/future looking type comments – the only ones that would be great to include in this PR are:
* Link to new doc in the subsystem README.md
* Documentation of environment variable in the new matching doc and in the iac.md file

match/src/Piipan.Match.Orchestrator/Api.cs
kingcomma: Great point. It's possible/likely that these URIs will change with #275 (e.g., the base host will be needed for authentication and `MatchState` will need to build the query endpoint URI). If so, I'll fold an initial implementation of this approach into that work.

ryanhofdotgov (approved):

Issue #215, opened 2021-01-12
Hello World for NAC Query app


cjhskippy:
NAC Query

- App can display "hello world"
- Build automatically / integrated in to CI / CD
- Runs a unit test on checkin

PR #271, opened 2021-01-26
Create "Hello, World" version of the query tool

heymatthenry:
This PR does two main things:

1. Creates a basic "Hello, World" version of the NAC query tool (per #215), which is just a Razor Pages form that doesn't yet do anything, along with a basic unit test to ensure the test harness is set up correctly.
2. Integrates the new app into the CI/CD pipeline.


query-tool/src/Piipan.QueryTool/Piipan.QueryTool.csproj
```diff
@@ -0,0 +1,8 @@
+<Project Sdk="Microsoft.NET.Sdk.Web">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <RootNamespace>Piipan.QueryTool</RootNamespace>
```
kingcomma, line 5: This tweak should enable repeatable package restores per [source tree checklist](https://github.com/18F/piipan/blob/main/docs/source-guidelines.md#source-tree-checklist)

```suggestion
    <RootNamespace>Piipan.QueryTool</RootNamespace>
    <RestorePackagesWithLockFile>true</RestorePackagesWithLockFile> and 
    <RestoreLockedMode Condition="'$(ContinuousIntegrationBuild)' == 'true'">true</RestoreLockedMode>
```

query-tool/src/Piipan.QueryTool/obj/project.assets.json
```diff
@@ -0,0 +1,66 @@
+{
```
kingcomma, line 1: Looks like `obj/` and some of the build files contained within snuck in on c4c0776ad26dfca2f581524e11b82a64e46c37eb.

query-tool/tests/Piipan.QueryTool.Tests/Piipan.QueryTool.Tests.csproj
```diff
@@ -0,0 +1,21 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
+
+    <IsPackable>false</IsPackable>
```
kingcomma, line 6: Same suggestion as with `Piipan.QueryTool.csproj`:
```suggestion
    <IsPackable>false</IsPackable>
    <RestorePackagesWithLockFile>true</RestorePackagesWithLockFile> 
    <RestoreLockedMode Condition="'$(ContinuousIntegrationBuild)' == 'true'">true</RestoreLockedMode>
```

iac/create-resources.bash
```diff
@@ -267,6 +270,18 @@ az deployment group create \
     appName=$DASHBOARD_APP_NAME \
     servicePlan=$APP_SERVICE_PLAN
 
+# Create App Service resources for dashboard app
```
kingcomma, line 273:
```suggestion
# Create App Service resources for query tool app
```

kingcomma:
Nice! I left a few suggestions in comments.

A couple other notes:
- It would be good to add a sub-system level README.md and a barebones query tool specific doc, similar to [`etl/`](https://github.com/18F/piipan/blob/main/etl/README.md) or [`dashboard/`](https://github.com/18F/piipan/blob/main/dashboard/README.md) (though note the updates correcting my rush job to the dashboard doc in #232)
- I'm on the fence on whether or not we should reuse a single ARM template for both the dashboard and query tool apps. I assume they will likely diverge in the future because of differences in, for example, `siteConfig`, but it looks like they are identical right now. We haven't established this as a practice yet (e.g., there is a function storage template in its own file and nested within the `function-state-match.json`), so I don't think it should hold up merging.

ryanhofdotgov:
🥳  on the first PR of many!

Just a nitpick: We unofficially use [2-space idents for bash files](https://google.github.io/styleguide/shellguide.html#s5.1-indentation), but have been using [OmniSharp defaults for C#](https://github.com/18F/piipan/blob/main/docs/source-guidelines.md#source-code-conventions), which I think defaults to 4-spaces.

query-tool/src/Piipan.QueryTool/obj/project.assets.json
heymatthenry: Whoops! I shuffled things around a little bit in the process of adding the `.gitignore` and somewhere in there this one dir got past the velvet rope. Gone now.

heymatthenry:
@ryanhofdotgov D'oh! That's what I get for bouncing between vim and VScode. I probably just need to stick with VSCode on this project.

heymatthenry:
@kingcomma Docs added. Ty!

Re: the ARM template, I figured they would end up diverging sooner or later—probably very soon since we're about to take up auth for the Query Tool—so I thought it was worth starting with a separate template, even if it's currently identical to the dashboard.

kingcomma:
@heymatthenry Thanks!

Were you able to get OmniSharp setup in VSCode? It looks like the C# files are still using 2 spaces for indentation (good catch @ryanhofdotgov!), and I'm just thinking it would be good to format the files in this PR so future PRs limit indentation-only changes.

Also, looks like I caused a conflict in `.circleci/config.yml` with #266 — sorry! 😬

heymatthenry:
@kingcomma Ok, should be all set with the formatting issues and Omnisharp is (I think) configured correctly to keep the whitespace groomed going forward.

kingcomma (approved):
👍

Issue #213, opened 2021-01-12
Shared access signatures (SAS) for per-state ETL storage accounts


ryanhofdotgov:
As we want to minimize the number of credentials the system operators have to maintain and update, implement Shared access signatures (SAS) for per-state storage accounts.

We've used managed identities directly in other parts of Piipan, but Azure Storage doesn't directly support that mechanism. Instead, you have to use the managed identity to indirectly generate a signed SAS URL that can be used to access the storage account.

AC:
- [ ] IaC no longer binds storage account access keys to ETL Function App
- [ ] Storage accounts access control (IAM) has role assignment of Storage Account Contributor to corresponding managed identity, in IaC
- [ ] BulkUpload function updated to generate SAS URL, which it uses itself to fetch the blob specified in the event

ryanhofdotgov:
Related URLs:
* https://docs.microsoft.com/en-us/samples/azure-samples/functions-storage-managed-identity/using-managed-identity-between-azure-functions-and-azure-storage/
* https://github.com/Azure-Samples/functions-storage-managed-identity
* https://docs.microsoft.com/en-us/azure/storage/common/storage-sas-overview
* https://docs.microsoft.com/en-us/azure/marketplace/azure-vm-get-sas-uri
* https://docs.microsoft.com/en-us/azure/storage/blobs/sas-service-create?tabs=dotnet
* https://docs.microsoft.com/en-us/rest/api/storageservices/create-account-sas

PR #274, opened 2021-01-26
Automate blob binding, update docs

### LINKED ISSUE #242: Update ETL docs
In the last sprint, most of the deployment pieces were automated, so we no longer need the explicit manual deployment steps.

Also, clarify use of grant-blob.bash, and the Azure delay in propagating out privilege changes.

ryanhofdotgov:
Closes #242

- Move blob connection string binding to Function app from manual process to IaC
- Remove manual documentation for setting service bindings
- Update IaC docs to reflect standard environment variable name for blob
  connection strings
- Add note about grant-blob taking several minutes to replicate out

### DIFF
--- a/docs/iac.md
+++ b/docs/iac.md
@@ -26,11 +26,12 @@ To (re)create the Azure resources that `piipan` uses:
 
 ### Environment variables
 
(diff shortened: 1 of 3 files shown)


ryanhofdotgov:
The work the comment block on `blob_connection_string` alludes to is being tracked in https://github.com/18F/piipan/issues/213

kingcomma (approved):
👍

Issue #97, opened 2020-11-18
Evaluate deployment strategies


kingcomma:
Building on the work in #93, research deployment strategies and ways of achieving zero downtime deploys in the Azure environment.

References:
- MS documentation for App Service [Deployment Best Practices](https://docs.microsoft.com/en-us/azure/app-service/deploy-best-practices)

PR #278, opened 2021-01-26
ADR for Razor Pages

### LINKED ISSUE #57: ADR for the app deployment approach
Provide rationale for the technical choice for app deployment strategy.

- [ ] Documentation in GitHub is updated

ryanhofdotgov:
Closes #57 –  retroactively documenting the Razor Pages decision we made a couple of months ago

### DIFF
--- a/docs/adr/0011-use-razor-pages-for-web-apps.md
+++ b/docs/adr/0011-use-razor-pages-for-web-apps.md
@@ -1,0 +1,31 @@
+# 11. Use Razor pages for web-apps
+
+Date: 2021-01-26
+
+## Status
+
+Accepted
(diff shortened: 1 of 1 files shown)


kingcomma (approved):
👍 Thanks for documenting this. I had been thinking about #57 as researching and deciding on a blue-green-style deployment approach for App Service apps / deployment slots. That led me to overlooking the missing ADR for this decision.

ryanhofdotgov:
Ah, I see that the description aligns with what you thought. But, we have https://github.com/18F/piipan/issues/97 to address the blue/green deployment model and that should result in an ADR.

PR #312, opened 2021-02-02
Hard code app name & resource group

heymatthenry:
The Azure app name & resource group were previously pulled in from CI environment variables. Because this isn't sensitive information, it's easier to just include it directly in the configuration.


kingcomma (approved):
👍 Thanks for picking this up!

Issue #281, opened 2021-01-26
Authentication mechanism for NAC Query Website


cjhskippy:
Provide a means for limiting access to the NAC Orchestrator API but allowing the Query Website to do its work.

- [x] Identify how internal tools should be able to have access to the NAC Orchestrator
- [x] Implement the access restrictions, preventing other internal systems in the Azure environment from having access to the Orchestrator API.
- [x] Query website can authenticate with the Orchestrator API

Issue #280, opened 2021-01-26
NAC Query Website contacts API, displays results


cjhskippy:
Website accepts Full name, DOB, SSN in form.

- Display response from API on the website.  Formatting not important.

PR #316, opened 2021-02-03
Implement API requests via Query Tool Form

heymatthenry:
This PR implements the basic functionality of the Query Tool, `POST`ing form data to a remote API and then rendering the response from the API back to the page, per #280.

I ended up merging in the commits from the USWDS work into this branch so it would be there for the demo, but as a result there's a fair bit of noise in the PR (sorry 😬). The real action is in:

1. [`OrchestratorApiRequest.cs`](https://github.com/18F/piipan/pull/316/files?file-filters%5B%5D=.cs&file-filters%5B%5D=.cshtml#diff-81d26e5b37f123ad309b333789b66ccc3763b1d4356fb89a0cf890593c619373R1) (and [`OrchestratorApiRequestTest.cs`](https://github.com/18F/piipan/pull/316/files?file-filters%5B%5D=.cs&file-filters%5B%5D=.cshtml#diff-92149c15f4f6c8aee107f0000bfb2d97ae68319646aee10353c7964717bb5e27R1)), which is responsible for querying the orchestrator API;
1. [`PiiRecord.cs`](https://github.com/18F/piipan/pull/316/files?file-filters%5B%5D=.cs&file-filters%5B%5D=.cshtml#diff-4191d0b76787c97292c2cd4bea89e7e02c9b54eaf8383eadf55dc02269bbe518R1), which models the PII as well as backing the query form and providing the keys for JSON serialization;
2. [`Index.cshtml`](https://github.com/18F/piipan/pull/316/files?file-filters%5B%5D=.cs&file-filters%5B%5D=.cshtml#diff-5d85e137b9ebedc14d719989d68b3fe94889a1421dee3c7bf1e0f040270d1954R7), which pulls the JSON from the API into the view;
3. [`Index.cshtml.cs`](https://github.com/18F/piipan/pull/316/files?file-filters%5B%5D=.cs&file-filters%5B%5D=.cshtml#diff-bed51eaffaebaeebd26b01aac8aeaa81cd059b97c195f5d05bc5255d24084318R19), which triggers the API call on `POST`.

A couple of caveats 🚧:

1. This will currently only work locally and on VPN until we have a solution for #281
2. The `onPostAsync` method that triggers the API call is not tested. I struggled with this for _a while_ before punting until we start doing integration testing.


query-tool/src/Piipan.QueryTool/OrchestratorApiRequest.cs
```diff
@@ -0,0 +1,58 @@
...
+
+        public async Task<string> SendQuery(string url, PiiRecord query)
+        {
+            RequestUrl = url;
+            Query = query;
+            _client = new HttpClient();
+            return await QueryOrchestrator();
+        }
+
+        public async Task<string> SendQuery(string url, PiiRecord query, HttpClient client)
```
heymatthenry, line 23: This constructor is overloaded with an extra parameter for an `HttpClient` in order to enable testing. It's not my favorite workaround but it's the best way I could figure out to let the API mock intercept requests.

ryanhofdotgov:
@heymatthenry Do you have an idea of items you want to address before marking this ready for review? Is it just working out the authentication to the orchestrator API?

heymatthenry:
@ryanhofdotgov No, I wasn't worrying about the authentication piece for this PR but I'm just struggling through a couple of testing issues that I'd like to wrap up before marking this ready for review.

query-tool/src/Piipan.QueryTool/OrchestratorApiRequest.cs
```diff
@@ -0,0 +1,53 @@
+using System;
+using System.Threading.Tasks;
+using System.Net.Http;
+using System.Text;
+using System.Text.Json;
+using System.Collections.Generic;
```
kingcomma, line 6: **Nitpick**: Looks like Omnisharp may not be [configured](https://github.com/18F/piipan/blob/main/docs/source-guidelines.md#source-code-conventions) to organize these imports.

query-tool/src/Piipan.QueryTool/gulpfile.js
```diff
@@ -0,0 +1,182 @@
+/*
```
kingcomma, line 1: **Question**: Does local development currently require running `gulp` at any point? If so, could you document the process in `docs/query-tool.md`?

kingcomma (commented):
This is great — so satisfying to see the systems come together with a push of a submit button!

My comments include a few chores/nitpicks to be addressed before merging.

query-tool/src/Piipan.QueryTool/OrchestratorApiRequest.cs
heymatthenry: Fixed in aa54030. I've intermittently had to turn off the autoformatter when editing razor pages since it gets confused by the template code, so I might've just neglected to turn it back on.

query-tool/src/Piipan.QueryTool/Pages/Index.cshtml.cs
```diff
@@ -17,10 +15,29 @@ public IndexModel(ILogger<IndexModel> logger)
...
+        [BindProperty]
+        public PiiRecord Query { get; set; }
+
+        private readonly OrchestratorApiRequest _apiRequest = new OrchestratorApiRequest();
+        public List<PiiRecord> QueryResult { get; private set; } = new List<PiiRecord>();
+
+        public async Task<IActionResult> OnPostAsync(PiiRecord query)
+        {
+            QueryResult = await _apiRequest.SendQuery(
+                "https://ofuncw5wltopux5z7i.azurewebsites.net/api/v1/query",
```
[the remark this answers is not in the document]
heymatthenry: Good call. Handled in add27c2.

query-tool/src/Piipan.QueryTool/gulpfile.js
heymatthenry: It only requires gulp if you're making changes to the styles. You need just to have `gulp watch` running while you're making changes to the SCSS. There's no other automated build tooling yet. This is now documented in e73308b.

heymatthenry:
@kingcomma Okey doke. I believe I've addressed your (helpful!) suggestions, save for the one regarding auth, which I'll tackle once we roll that feature out.

iac/create-resources.bash
```diff
@@ -247,6 +247,14 @@ az deployment group create \
...
 # Create App Service resources for query tool app
 echo "Creating App Service resources for query tool app"
+
+orch_api_uri=$(\
+  az functionapp function show \
+    -g piipan-match \
+    -n ofuncw5wltopux5z7i \
+    --function-name Query \
+    --query invokeUrlTemplate)
+
```
kingcomma, line 257: Thanks for adding this!

It's made me realize I overlooked something on previous review: since the dashboard is created before the orchestrator, this command will result in an empty string (because the orchestrator does not yet exist). So...

**Chore**: Move the dashboard-related IaC to after the orchestrator-related IaC. Also, at that point you should be able to swap out the hardcoded `ofuncw5wltopux5z7i` name with the `$orch_name` variable.

Apologies for overlooking this the first go-around!

kingcomma (commented):
Nice! Thanks for all these updates. Just one chore that resulted from my previous IaC request and would be great to wrap up before merging.

iac/create-resources.bash
heymatthenry: Good call! Thanks for catching that. Fixed in db8513a.

kingcomma (approved):
👍

Issue #101, opened 2020-11-24
Refactor IaC scripts


ryanhofdotgov:
create-resources.bash has become a beast:
* it's getting harder to maintain as most of it is lumped into one giant script
* it's not straight forward to only re-build portions of the infrastructure

- [x] Can be used in the FNS environment.
- [x] Scripts no longer rely on global administrator role to run

cjhskippy:
Flagging as a likely interest for sprint 3.  As we continue to add new skeletons of features, this is likely to only become more of a burden on productivity.  Might as well invest in it sooner if this is an ongoing source of toil.

kingcomma:
This may be out of scope or this issue, but it could be worth researching ways to achieve some of the functionality of [complete mode](https://docs.microsoft.com/en-us/azure/azure-resource-manager/templates/deployment-modes) within our current approach to IaC. Primarily, as a developer, I would like a way for resources that are no longer part of our IaC to be removed whenever `create-resources` is run.

With our current setup where ARM templates are deployed to shared resource groups via several calls to `az deployment group create`, complete mode would be unworkable and removing a resource like an unused deployment slot on app service requires manual work.

ryanhofdotgov:
https://github.com/18F/piipan/issues/101#issuecomment-745349820 spilt off into https://github.com/18F/piipan/issues/544

ryanhofdotgov:
While there is still more refactoring and decomposition to be done, the referenced PRs above address the scope of this issue.

PR #317, opened 2021-02-04
Implement authentication and authorization between orchestrator and per-state APIs

### LINKED ISSUE #275: Authentication between match orchestrator and per-state APIs
As we want the per-state APIs to be internal-facing components, restrict access to the per-state API function apps and implement an authentication mechanism for use by the orchestrator function app. It appears this can be achieved via a system assigned identity for the orchestrator that uses an Azure AD token for authenticating with the per-state APIs.

AC:
- [ ] Per-state APIs are no longer externally accessible via a trusted network
- [ ] Orchestrator identity is provided role-based access to per-state APIs
- [ ] Orchestrator implements mechanism of obtaining an authorization token and passing it to the per-state API

kingcomma:
Closes #275. Details:
- Enables app service authentication (aka Easy Auth) for the per-state match APIs via IaC
- Creates and assigns an "app role" used to authorize authenticated requests to a state API
- Updates the orchestrator API to obtain and pass Easy Auth bearer tokens when querying a state API
- Updates OpenAPI specs to reference new bearer token requirement

Follow up work will include:
- Minimizing the use of the AZ CLI in our IaC in favor of ARM templates or [deployment scripts](https://docs.microsoft.com/en-us/azure/azure-resource-manager/templates/deployment-script-template) (this update is probably the last straw for the existing IaC script) #101
- Migrating the authentication library used by each API subsystem to `Azure.Identity` #318
- [More meaningful validation of service binding strings passed from the environment](https://github.com/18F/piipan/pull/266#discussion_r564908491)
- Integration testing, improved test coverage #154 and #342

### DIFF
--- a/docs/iac.md
+++ b/docs/iac.md
@@ -32,11 +32,9 @@ The following environment variables are pre-configured by the Infrastructure-as-
 |---|---|---|
 | `DatabaseConnectionString` | ADO.NET-formatted database connection string. If `Password` has the value `{password}`; i.e., `password` in curly quotes, then it is a partial connection string indicating the use of managed identities. An access token must be retrieved at run-time (e.g., via [AzureServiceTokenProvider](https://docs.microsoft.com/en-us/dotnet/api/overview/azure/service-to-service-authentication)) to build the full connection string.  | Piipan.Etl, Piipan.Match.State |
 | `BlobStorageConnectionString` | Azure Storage Account connection string for accessing blobs. | Piipan.Etl |
-| `StateApiHostStrings` | Serialized JSON array of valid URI strings for accessing each per-state matching API. | Piipan.Match.Orchestrator |
-| `StateApiEndpointPath` | Relative path for per-state API Query endpoint. | Piipan.Match.Orchestrator |
+| `StateApiUriStrings` | Serialized JSON array of valid URI strings for each per-state matching API's `/query` endpoint. | Piipan.Match.Orchestrator |
 | `StateName` | Name of the state associated with the Function App instance. | Piipan.Match.State |
 | `StateAbbr` | Abbreviation of the state associated with the Function App instance. | Piipan.Match.State |
-| `AuthorizedRoleName` | Name of the [app role](https://docs.microsoft.com/en-us/azure/active-directory/develop/howto-add-app-roles-in-azure-ad-apps) used to authorize requests to the per-state APIs | Piipan.Match.State |
 
 ## Notes
 - `iac/states.csv` contains the comma-delimited records of participating states/territories. The first field is the [two-letter postal abbreviation](https://pe.usps.com/text/pub28/28apb.htm); the second field is the name of the state/territory.
--- a/docs/iac.md
+++ b/docs/iac.md
@@ -32,9 +32,11 @@ The following environment variables are pre-configured by the Infrastructure-as-
 |---|---|---|
 | `DatabaseConnectionString` | ADO.NET-formatted database connection string. If `Password` has the value `{password}`; i.e., `password` in curly quotes, then it is a partial connection string indicating the use of managed identities. An access token must be retrieved at run-time (e.g., via [AzureServiceTokenProvider](https://docs.microsoft.com/en-us/dotnet/api/overview/azure/service-to-service-authentication)) to build the full connection string.  | Piipan.Etl, Piipan.Match.State |
 | `BlobStorageConnectionString` | Azure Storage Account connection string for accessing blobs. | Piipan.Etl |
-| `StateApiEndpointStrings` | Serialized JSON array of endpoints strings for accessing each per-state matching API. | Piipan.Match.Orchestrator |
+| `StateApiHostStrings` | Serialized JSON array of valid URI strings for accessing each per-state matching API. | Piipan.Match.Orchestrator |
+| `StateApiEndpointPath` | Relative path for per-state API Query endpoint. | Piipan.Match.Orchestrator |
 | `StateName` | Name of the state associated with the Function App instance. | Piipan.Match.State |
 | `StateAbbr` | Abbreviation of the state associated with the Function App instance. | Piipan.Match.State |
+| `AuthorizedRoleName` | Name of the [app role](https://docs.microsoft.com/en-us/azure/active-directory/develop/howto-add-app-roles-in-azure-ad-apps) used to authorize requests to the per-state APIs | Piipan.Match.State |
 
 ## Notes
 - `iac/states.csv` contains the comma-delimited records of participating states/territories. The first field is the [two-letter postal abbreviation](https://pe.usps.com/text/pub28/28apb.htm); the second field is the name of the state/territory.
--- a/iac/create-resources.bash
+++ b/iac/create-resources.bash
@@ -374,12 +374,12 @@ done < states.csv
 # Assumes existence of a managed identity with name `{abbr}admin`.
 
 # Relative path for per-state Query endpoint
-MATCH_API_PATH='/api/v1/query'
+MATCH_API_QUERY_NAME='Query'
 
 # Name of application role authorized to call per-state APIs
 MATCH_API_APP_ROLE='StateApi.Query'
 
-match_api_hosts=''
+match_api_uris=''
 match_func_names=()
 
 while IFS=, read -r abbr name ; do
@@ -413,8 +413,7 @@ while IFS=, read -r abbr name ; do
         stateName="$name" \
         stateAbbr="$abbr" \
         dbConnectionString="$db_conn_str" \
-        dbConnectionStringKey="$DB_CONN_STR_KEY" \
-        authorizedRoleName="$MATCH_API_APP_ROLE")
+        dbConnectionStringKey="$DB_CONN_STR_KEY")
   
   # Store function names for future auth configuration
   match_func_names+=("$func_name")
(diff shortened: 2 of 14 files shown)


ryanhofdotgov:
@kingcomma Wow! This is an impressive set of changes. I'll be able to take a look at it more closely tomorrow and early Monday.

docs/iac.md
```diff
@@ -32,9 +32,11 @@ The following environment variables are pre-configured by the Infrastructure-as-
 |---|---|---|
 | `DatabaseConnectionString` | ADO.NET-formatted database connection string. If `Password` has the value `{password}`; i.e., `password` in curly quotes, then it is a partial connection string indicating the use of managed identities. An access token must be retrieved at run-time (e.g., via [AzureServiceTokenProvider](https://docs.microsoft.com/en-us/dotnet/api/overview/azure/service-to-service-authentication)) to build the full connection string.  | Piipan.Etl, Piipan.Match.State |
 | `BlobStorageConnectionString` | Azure Storage Account connection string for accessing blobs. | Piipan.Etl |
-| `StateApiEndpointStrings` | Serialized JSON array of endpoints strings for accessing each per-state matching API. | Piipan.Match.Orchestrator |
+| `StateApiHostStrings` | Serialized JSON array of valid URI strings for accessing each per-state matching API. | Piipan.Match.Orchestrator |
+| `StateApiEndpointPath` | Relative path for per-state API Query endpoint. | Piipan.Match.Orchestrator |
```
ryanhofdotgov, line 36: What's the motivation for keeping these separate?

Not saying we will do this, but it is theoretically possible that backwards compatible changes to the state API might be rolled out incrementally across the states. For example, update 10% of states to new version of API, see that no errors are occurring over a period of time, then roll out update to 100% of the endpoints.

iac/create-resources.bash
```diff
@@ -451,44 +459,199 @@ while IFS=, read -r abbr name ; do
...
+#
+# For each state:
+#   - Register an Azure Active Directory (AAD) app with an application
+#     role named the value of `MATCH_API_APP_ROLE`
+#   - Create a service principal (SP) for the app registation
+#   - Add the application role to the orchestrator API's identity
+#   - Configure and enable App Service Authentiction (i.e., Easy Auth)
+#     for state's Function app.
+#
+# Note: while all incoming requests to the state API will need to
```
ryanhofdotgov, line 523: Using the security vernacular, this makes the application the policy enforcement point. It's not possible to enforce the policy (e.g., only those with the role can access the API) at a layer above?

match/docs/state-match.md
```diff
@@ -54,13 +55,22 @@ func azure functionapp publish <app_name> --dotnet
...
-1. Send a valid `POST` request to the app's endpoint. The endpoint is in the format `https://{app-name}.azurewebsites.net/api/v1/query`.
+1. The client application requests a token from the app registration
+1. The token is included as an authentication header in the request sent to the API's query endpoint
+1. The API verifies that the authorized user has the `StateApi.Query` role (and returns a `401 unauthorized` error if not)
+1. The API performs the requested operation
+
+## Remote testing
 
-Until the ETL process is updating to insert data into tables in the `piipan` schema, responses will not contain matches unless records have been manually imported into the database.
+With authentication enabled, there is currently no way to access the remote per-state APIs directly. Verify functionality by sending requests through the [orchestrator API](orchestrator-match.md).
```
ryanhofdotgov, line 76: I think we will need an explicit follow on ticket that will grant the CI/CD service principal the role, so it can execute integration tests (at least in sandbox). And then perhaps manual instructions (or short script) for assigning the role to a dev's account?

match/src/Piipan.Match.Orchestrator/Api.cs
```diff
@@ -96,36 +98,73 @@ internal static bool Validate(MatchQueryRequest request, ILogger log)
...
         {
-            var response = await client.PostAsJsonAsync(uri, request);
-            var matchResponse = await response.Content.ReadAsAsync<MatchQueryResponse>();
+            // Retrieve authentication token via Azure AD Easy Auth. Passing an empty
+            // string to `AzureServiceTokenProvider` uses system-assigned identity.
+            var azureServiceTokenProvider = new AzureServiceTokenProvider();
+
+            // Passed URI must match AAD app URI exactly; AAD app URI does not contain
+            // a trailing slash, but `Uri.AbsoluteUri` will add one.
+            var resource = uri.AbsoluteUri.TrimEnd('/');
```
ryanhofdotgov, line 120: So in this case, it looks like Uri was the wrong abstraction. If the code is going to use it as a URL that they actually post/get/connect to, then it makes sense.

If instead the Uri is really just a opaque token used to namespace a request, then using a plain old String is better.

match/src/Piipan.Match.State/Api.cs
```diff
@@ -58,6 +65,32 @@ public static class Api
...
         }
 
+        internal static bool Authorize(ClaimsPrincipal identity, ILogger log)
+        {
+            const string AuthorizedRoleName = "AuthorizedRoleName";
+            Claim roleClaim = identity.FindFirst("roles");
+
+            // Authorized user must be in specified app role
+            var role = Environment.GetEnvironmentVariable(AuthorizedRoleName);
+            if (roleClaim == null || !roleClaim.Value.Split(' ').Contains(role))
```
ryanhofdotgov, line 75: If we must do policy enforcement in the app code itself, then we might need to be more defensive here...
* validate `role` is non-null
* consider tradeoff in hard-coded value in the source?

match/src/Piipan.Match.State/Api.cs
```diff
@@ -58,6 +65,32 @@ public static class Api
...
+            {
+                log.LogError("Caller does not have authorized role.");
+                return false;
+            }
+
+            // Authorized "user" must be an application, in which case
+            // the object ID (`oid`) and subject (`sub`) will match
+            string oid = identity.FindFirst("oid")?.Value;
+            string sub = identity.FindFirst("sub")?.Value;
```
ryanhofdotgov, line 84: Is it possible for `oid` and `sub` to both be null, causing the flow to incorrectly exit with `true`?

ryanhofdotgov (commented):
Nice work pulling together all the moving parts to get this authn flow working!

The main items I'd like to work out before approving are the questions in the `Authorize` method in the state-level `Api.cs`.

iac/create-resources.bash
kingcomma: From my research, there's no way to enforce a _specific_ role assignment when using Easy Auth on a specific Function / App Service app. There is a coarse-grained option to require _any_ app role assignment that I haven't explored fully ([see the "Important" note here](https://docs.microsoft.com/en-us/azure/active-directory/develop/scenario-protected-web-api-app-registration#ensuring-that-azure-ad-issues-tokens-for-your-web-api-to-only-allowed-clients)).

Beyond that, my understanding is we would need to put an API Management instance in front of the function app to allow/enforce role-based policies (e.g., use a [JWT validation policy](https://docs.microsoft.com/en-us/azure/active-directory/develop/scenario-protected-web-api-app-registration#ensuring-that-azure-ad-issues-tokens-for-your-web-api-to-only-allowed-clients)).

Agreed enforcing this policy at the application-level doesn't feel like the ideal solution — my justification for using it is Easy Auth seems like the closest match to our use of managed identities elsewhere and this app role approach is [recommended](https://docs.microsoft.com/en-us/azure/azure-functions/functions-bindings-http-webhook-trigger?tabs=csharp#secure-an-http-endpoint-in-production) in several of MS's docs.

match/src/Piipan.Match.State/Api.cs
kingcomma: Resolved by removing application-level enforcement of this policy in 6d640c9.

match/src/Piipan.Match.State/Api.cs
kingcomma: Resolved by removing application-level enforcement of this policy in 6d640c9.

match/docs/state-match.md
kingcomma: Ticket created — #342.

iac/create-resources.bash
kingcomma: After discussion and more research, I've implemented the "user assignment required" approach referenced in [this doc](https://docs.microsoft.com/en-us/azure/active-directory/develop/scenario-protected-web-api-app-registration#ensuring-that-azure-ad-issues-tokens-for-your-web-api-to-only-allowed-clients). This enforces the application role policy at the AAD layer and removes any enforcement from the application itself.

Implementation details in 6d640c9.

docs/iac.md
kingcomma: I've reverted to using a single environment variable to bind all the state APIs and derive the necessary URI string for requesting an access token from the full endpoint URIs.

match/src/Piipan.Match.Orchestrator/Api.cs
kingcomma: Makes sense, really appreciate the clear description of this implementation. Active Directory refers to the string passed to `AzureServiceTokenProvider.GetAccessTokenAsync()` as an "application URI" and that led me down the path to this approach.

6d640c9 changes the implementation to utillize a single set of URIs (the full API endpoints) which are then used when calling the APIs and for [deriving the "application URI" string](https://github.com/18F/piipan/pull/317/files#diff-b1e5f06978eb6bb97a93fe1ef25f88db4676446d41b9e4496fef696a6185a95fR141).

iac/create-resources.bash
```diff
@@ -495,117 +475,174 @@ app_role_assignment () {
...
+    --aad-allowed-token-audiences $func_host $func_uri \
     --aad-client-id $func_app_reg_id \
     --aad-token-issuer-url "https://sts.windows.net/${TENANT_ID}/" \
     --enabled true \
     --action LoginWithAzureActiveDirectory
+
+  # Any client that attemps authentication must be assigned a role
+  az ad sp update \
+    --id $func_app_sp \
+    --set "appRoleAssignmentRequired=true"
```
ryanhofdotgov, line 645: Just adding a note here for posterity that this is final bit of set up that we will want to replicate between pretty much all apps/batch jobs/etc. and the Function-based internal APIs.

ryanhofdotgov (approved):
Fantastic to see these nitty-gritty internal authn/authz issues worked out. Thanks!

Issue #355, opened 2021-02-17
Shared library for managing sub-system authentication


kingcomma:
We're using App Service authentication to secure our internal HTTP-based APIs. This requires sub-systems to retrieve an authentication token via a managed identity and include it in any API calls. We need a shared library that can abstract away this added complexity, support local and CI/CD environments, and ease unit testing.

Related to #192

PR #356, opened 2021-02-17
Refactor authentication implementation to shared library

kingcomma:
A proof-of-concept for abstracting "[Easy Auth](docs/adr/0012-use-app-service-authentication-between-subsystems.md)" authentication to a shared library:
- Creates a new `Piipan.Shared` project.
- Implements the helpers from `Piipan.Shared.Authentication` in `Piipan.Match.Orchestrator` using [dependency injection](https://docs.microsoft.com/en-us/azure/azure-functions/functions-dotnet-dependency-injection).
- Updates the relevant unit tests to mock the new classes and provide more comprehensive unit testing.

Interested in feedback on the general approach to a shared library, knowing the intention is to eventually incorporate #192.

Also interested in feedback on using dependency injection in a Function App. See `Piipan.Match.Orchestrator/Startup.cs` and `Piipan.Match.Orchestrator/Api.cs`. My intent was to show what this shared library could do by way of facilitating unit testing. I'm A-OK cutting this out — whether that's implementing in a separate PR or ditching entirely.


shared/src/Piipan.Shared/Authentication.cs
```diff
@@ -0,0 +1,95 @@
...
+                    { HttpRequestHeader.Accept.ToString(), _accept }
+                }
+            };
+
+            return httpRequestMessage;
+        }
+
+        public async Task<HttpResponseMessage> PostAsync(Uri uri, StringContent body)
+        {
+            var requestMessage = await PrepareRequest(uri, HttpMethod.Post);
```
echappen, line 82: [non-blocking] Maybe the `PostAsync` method could allow additional headers from the caller, but I don't have a concrete use case for this at the moment.

echappen (approved):
Glad to see an example of dependency injection in a function app! Thanks for doing this.

match/src/Piipan.Match.Orchestrator/Api.cs
```diff
@@ -32,7 +36,7 @@ public static class Api
         /// access to the individual per-state API resources.
         /// </remarks>
         [FunctionName("query")]
-        public static async Task<IActionResult> Query(
+        public async Task<IActionResult> Query(
```
ryanhofdotgov, line 39: @kingcomma From the [docs I read](https://docs.microsoft.com/en-us/azure/azure-functions/functions-dotnet-class-library?tabs=v2%2Ccmd#methods-recognized-as-functions), I was thinking the entry point had to be `static`?

ryanhofdotgov: Hmm... [this comment on an Azure Function issue](https://github.com/Azure/Azure-Functions/issues/1680#issuecomment-691668091) seems to contradict that documentation.

kingcomma: @ryanhofdotgov Agreed — it's a clear inconsistency in the MS docs. The [doc specific to dependency injection in Functions](https://docs.microsoft.com/en-us/azure/azure-functions/functions-dotnet-dependency-injection#use-injected-dependencies) has this to say:

> The use of constructor injection requires that you do not use static classes for injected services or for your function classes.

`static` seems like the natural choice for Functions. Even though the docs say otherwise, it still feels strange to not go that route.

kingcomma:
@echappen and @ryanhofdotgov I've taken this out of draft status and it is now ready for formal review.

With my new work, this PR looks bigger than it is. The bulk of the work that needs review is limited to two files, neither of which have changed since you first took a look:
- [`shared/src/Piipan.Shared/Authentication.cs`](https://github.com/18F/piipan/pull/356/files#diff-61dd4aa78d2f8d6220fac6f5cc66a12675f36586885343ed9729fad96cb210ee)
- [`match/src/Piipan.Match.Orchestrator/Api.cs`](https://github.com/18F/piipan/pull/356/files#diff-b1e5f06978eb6bb97a93fe1ef25f88db4676446d41b9e4496fef696a6185a95f)

The new work is mostly limited to the standard boilerplate associated with new source trees (`.csproj`, package locks, CI/CD) and includes only one file of substance:
- [`shared/tests/Piipan.Shared.Tests/AuthenticationTests.cs`](https://github.com/18F/piipan/pull/356/files#diff-23d112630ecf48ed078df05bf28df10825422e730bd7093a7fb81e56057694b3)

Closes #355. Partially addresses #281 by providing the mechanism for making authenticated API calls.

ryanhofdotgov (approved):
Looks good 👍 – I've opened up this follow on to consider if/how tokens should be cached: https://github.com/18F/piipan/issues/369

PR #365, opened 2021-02-18
feat: Add basic server-side form validation

heymatthenry:
- Adds built-in server-side form validation for the query tool webapp
- Displays validation errors using the USWDS alert component

Currently, all fields are required: 
image

And formatting for SSN & DOB need to match the API's formatting:
image

### DIFF
--- a/query-tool/src/Piipan.QueryTool/PiiRecord.cs
+++ b/query-tool/src/Piipan.QueryTool/PiiRecord.cs
@@ -5,22 +5,31 @@ namespace Piipan.QueryTool
 {
     public class PiiRecord
     {
+        [Required]
         [Display(Name = "First Name")]
         [JsonPropertyName("first")]
         public string FirstName { get; set; }
 
+        [Required]
         [Display(Name = "Middle Name")]
         [JsonPropertyName("middle")]
         public string MiddleName { get; set; }
 
+        [Required]
         [Display(Name = "Last Name")]
         [JsonPropertyName("last")]
         public string LastName { get; set; }
 
+        [Required]
+        [RegularExpression(@"^\d{4}-\d{2}-\d{2}$",
+            ErrorMessage = "Birth date must have the form YYYY-MM-DD")]
         [Display(Name = "Date of Birth")]
         [JsonPropertyName("dob")]
         public string DateOfBirth { get; set; }
 
+        [Required]
+        [RegularExpression(@"^\d{3}-\d{2}-\d{4}$",
+            ErrorMessage = "SSN must have the form 000-00-0000")]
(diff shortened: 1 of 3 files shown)


query-tool/src/Piipan.QueryTool/PiiRecord.cs
```diff
@@ -5,22 +5,31 @@ namespace Piipan.QueryTool
...
         public string MiddleName { get; set; }
 
+        [Required]
         [Display(Name = "Last Name")]
         [JsonPropertyName("last")]
         public string LastName { get; set; }
 
+        [Required]
+        [RegularExpression(@"^\d{4}-\d{2}-\d{2}$",
+            ErrorMessage = "Birth date must have the form YYYY-MM-DD")]
```
kingcomma, line 25: **Question**: Have you done any experimenting with declaring this property as `DateTime`? I'm wondering if parsing it as such and using [DataType attributes](https://docs.microsoft.com/en-us/aspnet/core/tutorials/razor-pages/validation?view=aspnetcore-5.0&tabs=visual-studio#use-datatype-attributes) would provide more robust validation out of the box?

heymatthenry: I hadn't, but I just did and it works great. So now it's using a nice-looking `input type="date"` on the front end and converting to the required `yyyy-MM-dd` on the back end.

kingcomma: Cool — interesting to see Razor pages flexing some muscle.

When I run the this locally, I now see two error messages saying `The value '' is invalid` when I submit the form without a date of birth. Some googling points to [this doc about validating non-nullable types](https://docs.microsoft.com/en-us/aspnet/core/mvc/models/validation?view=aspnetcore-3.1#required-validation-on-the-server).

Screen Shot 2021-02-19 at 9 28 31 AM

heymatthenry: Good catch! My bad for not testing the null case after changing the type 🤦🏻. [The fix](https://github.com/18F/piipan/pull/365/commits/c073977d183f1dd974af3fe09db93dce72d86813#diff-4191d0b76787c97292c2cd4bea89e7e02c9b54eaf8383eadf55dc02269bbe518R29) ended up being to make the `DateTime` type nullable. 

I tried the other solution, which sort of worked, but the generic validation message that didn't include the field name wasn't ideal. And it didn't fix the duplicate message issue you can see in your screenshot. Making the field a nullable `DateTime` did the trick.

kingcomma (approved):
👍

PR #366, opened 2021-02-18
Update dependencies, remove AppAuthentication dependency from PiipanMetricsApi

kingcomma:
@echappen `Microsoft.Azure.Services.AppAuthentication` was flagged as out of date in `PiipanMetricsApi`, but I think it's unused now that the project uses `Azure.Identity`. I removed the dependency and tests / builds run fine, but would appreciate a look in case I missed anything.


echappen (approved):
That's correct—good catch.

PR #384, opened 2021-02-22
feat: Display alert when search returns no results

heymatthenry:
Provide visual feedback to the user that the query completed successfully, but that there were no matches:

image


kingcomma (approved):
👍 

Unit tests to verify these model state details should be straightforward to implement once the shared authentication library is integrated and API requests are easily mockable.

PR #387, opened 2021-02-22
show participant upload metrics in dashboard

echappen:
Wanted to try getting this in before our demo tomorrow. This is a completely un-styled version of showing participant upload metrics in the Dashboard app through server-side API requests.

Creates a new Dashboard page and model `ParticipantUploads` which shows:
   - list of most recent participant uploads from the etl `BulkUpload` actions
    - a search input with a "clear search" button to query by state name
    - prev/next page links

Does not have:
    - frontend validation for search input
    - USWDS styling

I referred a lot to @heymatthenry's work in how to make server-side api requests and how to mock api requests in tests (thanks, Matt).

If you test locally, you need to be on the GSA network in order to make the API request.  

Specifically looking for better ways to test the page models, and to possibly use dependency injection for the Api service used there. I'm also wondering if my naming and breakdown of the Api classes makes sense.

### DIFF
--- a/dashboard/src/Piipan.Dashboard/Pages/Index.cshtml
+++ b/dashboard/src/Piipan.Dashboard/Pages/Index.cshtml
@@ -2,3 +2,5 @@
 @model IndexModel
 
 <p>@Model.Message</p>
+
+<a class="" asp-page="/ParticipantUploads">Participant Uploads</a>
--- a/dashboard/src/Piipan.Dashboard/Pages/Index.cshtml
+++ b/dashboard/src/Piipan.Dashboard/Pages/Index.cshtml
@@ -3,4 +3,4 @@
 
 <p>@Model.Message</p>
 
-<a class="" asp-page="/ParticipantUploads">Participant Uploads</a>
+<a asp-page="/ParticipantUploads">Participant Uploads</a>
--- a/dashboard/src/Piipan.Dashboard/Pages/ParticipantUploads.cshtml.cs
+++ b/dashboard/src/Piipan.Dashboard/Pages/ParticipantUploads.cshtml.cs
@@ -1,10 +1,10 @@
 using System;
 using System.Collections.Generic;
 using System.Net.Http;
-using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.AspNetCore.Mvc.RazorPages;
+using Microsoft.AspNetCore.WebUtilities;
 using Piipan.Dashboard.Api;
 
 #nullable enable
@@ -37,7 +37,8 @@ namespace Piipan.Dashboard.Pages
             if (BaseUrl == null)
                 throw new Exception("BaseUrl is null.");
             StateQuery = Request.Form["state"];
-            var url = $"{BaseUrl}?state={StateQuery}&perPage={PerPageDefault}";
+            var url = QueryHelpers.AddQueryString(BaseUrl, "state", StateQuery);
+            url = QueryHelpers.AddQueryString(url, "perPage", PerPageDefault.ToString());
             var api = new ParticipantUploadRequest(httpClient);
             var response = await api.Get(url);
             ParticipantUploadResults = response.data;
@@ -53,19 +54,7 @@ namespace Piipan.Dashboard.Pages
             var url = BaseUrl + Request.QueryString;
             StateQuery = Request.Query["state"];
             if (String.IsNullOrEmpty(Request.Query["perPage"]))
-            {
-                string perPageDefault = $"perPage={PerPageDefault}";
-                Regex regex = new Regex(@"\?");
-                Match match = regex.Match(url);
-                if (match.Success)
-                {
-                    url += String.Concat("&", perPageDefault);
-                }
-                else
-                {
-                    url += String.Concat("?", perPageDefault);
-                }
-            }
+                url = QueryHelpers.AddQueryString(url, "perPage", PerPageDefault.ToString());
             return url;
         }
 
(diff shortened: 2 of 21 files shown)


metrics/src/Piipan.Metrics/PiipanMetricsApi/GetParticipantUploads.cs
```diff
@@ -88,6 +93,22 @@ public static class GetParticipantUploads
...
+            int perPage,
+            Int64 total)
+        {
+            var newPage = page - 1;
+            if (newPage <= 0) return null;
+
+            var result = "";
+            if (!String.IsNullOrEmpty(state))
+                result += $"&state={state}";
+            result += $"&page={newPage}&perPage={perPage}";
```
heymatthenry, line 108: Down the road, it might be worth considering refactoring this towards using [QueryHelpers](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.webutilities.queryhelpers?view=aspnetcore-5.0).

dashboard/src/Piipan.Dashboard/Pages/ParticipantUploads.cshtml.cs
```diff
@@ -0,0 +1,73 @@
...
+            var api = new ParticipantUploadRequest(httpClient);
+            var response = await api.Get(url);
+            ParticipantUploadResults = response.data;
+            SetPageLinks(response.meta);
+        }
+
+        public async Task<IActionResult> OnPostAsync()
+        {
+            StateQuery = Request.Form["state"];
+            var url = $"{BaseUrl}?state={StateQuery}&perPage={PerPageDefault}";
```
heymatthenry, line 38: Similarly to `GetParticipantUploads.cs`, there might be some maintainability benefits to using a library for building & parsing the query strings.

dashboard/src/Piipan.Dashboard/Pages/ParticipantUploads.cshtml.cs
```diff
@@ -0,0 +1,73 @@
...
+{
+    public class ParticipantUploadsModel : PageModel
+    {
+        public string Title = "Participant Uploads";
+        public List<ParticipantUpload> ParticipantUploadResults { get; private set; } = new List<ParticipantUpload>();
+        public string? NextPageParams { get; private set; }
+        public string? PrevPageParams { get; private set; }
+        public string? StateQuery { get; private set; }
+        public static int PerPageDefault = 10;
+        public static string BaseUrl = "https://piipanmetricsapiztqzsbh432oyw.azurewebsites.net/api/GetParticipantUploads";
```
heymatthenry, line 22: I think going forward we're moving URLs into the IAC rather than hard-coding them. For reference, check out add27c2 where I made that change to how the query tool gets the URL for the orchestrator API.

dashboard/src/Piipan.Dashboard/Pages/Index.cshtml
```diff
@@ -2,3 +2,5 @@
 @model IndexModel
 
 <p>@Model.Message</p>
+
+<a class="" asp-page="/ParticipantUploads">Participant Uploads</a>
```
heymatthenry, line 6: Tiny nitpick: is there a reason for the empty class? Truly NBD if not since the markup will get overhauled once we add USWDS.

heymatthenry (commented):
This is awesome! So cool to see things start to come together. I made a couple of suggestions in there, but I think the most important one is the hard-coded URL, especially since there's a risk that might change when we move to the new Azure environment.

dashboard/src/Piipan.Dashboard/Pages/ParticipantUploads.cshtml.cs
echappen: @heymatthenry Thanks! Do you know if this Env Var would be accessible locally / in dev mode?

heymatthenry: If you just `export YourEnvVarName=https://piipanmetricsapiztqzsbh432oyw.azurewebsites.net/api/GetParticipantUploads` in the shell you launch the app in, it'll work. That's what I've been doing for local dev. @kingcomma looked into this too and there might be a more .NET-y way to achieve this using a [configuration provider](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/?view=aspnetcore-3.1).

dashboard/src/Piipan.Dashboard/Pages/Index.cshtml
echappen: No reason—removed

dashboard/src/Piipan.Dashboard/Pages/ParticipantUploads.cshtml.cs
echappen: Updated [here](https://github.com/18F/piipan/pull/387/commits/2ea464a40d50b1b7e800de97acef5872c0fe3e6c). I had to shift some things around in the iac scripts and change where the Dashboard app was being deployed, now that it relies on Metrics resources. Would welcome thoughts on how to fetch the metrics api function app name independently, but this may be moot after we revisit these scripts in general.

heymatthenry: LGTM, I think the main thing was getting rid of the hard-coded stuff. I'm not sure it's worth going too deep into how those env vars get set since, as you say, the details might change soon.

heymatthenry:
I see you're still tinkering here and it's awesome that you moved toward the DI approach in c801466. Feel free to keep refining but otherwise this is 👍🏻 as far as I'm concerned.

heymatthenry (approved):

echappen:
@heymatthenry Thanks for reviewing! The DI approach helped me mock some features of my page model for testing (which felt great), but not others. Still not sure how to mock `Request` in my page model, which would help test that better. Since we're revisiting test coverage soon, I'll move on to making this pretty with USWDS.

PR #431, opened 2021-02-25
Add local version of USWDS to dashboard app through npm/gulp

echappen:
Lots off static files added here to wwwroot directory. Notable files for review would be:
- package.json
- gulpfile.js
- Shared/_Layout.cshtml where resource links are added


dashboard/src/Piipan.Dashboard/Pages/ParticipantUploads.cshtml
```diff
@@ -1,41 +1,54 @@
...
-                <td>@record.UploadedAt</td>
-            </tr>
-        }
-    </table>
+    @if (Model.ParticipantUploadResults.Any())
+    {
+        <table class="usa-table">
+            <thead>
+                <th>state</th>
+                <th>uploaded at</th>
```
heymatthenry, line 30:
```suggestion
                <th scope="col">state</th>
                <th scope="col">uploaded at</th>
```

heymatthenry: This table isn't complicated, but adding `scope` is still a good idea for a11y.

heymatthenry (changes requested):
Good to go except for that one small suggestion about the table markup.

heymatthenry (approved):
LGTM 👍

Issue #388, opened 2021-02-22
Downgrade piipan-metrics-db in dev to lower price tier


ryanhofdotgov:
Looks like `piipan-metrics-db` in dev is at a ~$150/month tier. Let's bring it down to the same pricing tier as `participant-records`which is ~$35/month – I do not anticipate any impact to dev cycle. 

- [ ] If you go to [Azure cost analysis](https://portal.azure.com/#blade/Microsoft_Azure_GTM/ModernBillingMenuBlade/costAnalysis), expecting > $100 a month reduction in projected costs after change.

Issue #11, opened 2020-10-20
Lock down access to participant database through Azure resources that need it


ryanhofdotgov:
Using IaC approach in #1, establish internal network isolation for the per-state databases and the per-state ETL; see #7 and #8. Remove AllowAllWindowsAzureIps setting in the process.

echappen:
@ryanhofdotgov I'm running into [an issue I had previously](https://github.com/18F/piipan/pull/435#issuecomment-786907159) when attempting to deploy a postgres database on a Basic tier with VNet resources:

```
"code": "FeatureNotSupportedForEdition",
"message": "This feature is not available for the selected edition 'Basic'."
```
I finally found where this is [specified in the docs](https://docs.microsoft.com/en-us/azure/postgresql/concepts-limits#vnet-service-endpoints). We may have to bump the participant-records db server up from Basic to General Purpose.

echappen:
Adding what I found here about Azure Gov's use of VNets and Private Links:
- [Azure post](https://azure.microsoft.com/en-us/updates/privatelinkgagovcloud/) claiming private links are available to Postgres Databases in Azure Gov and have feature parity with Azure Commercial
- [General Availability](https://azure.microsoft.com/en-us/global-infrastructure/services/?products=virtual-network,private-link) of VNets and Private Links across all resources in Azure Gov
- VNets and Private Endpoints were available for creation in the FNS portal

Since this issue turned into a research spike, I'll open another issue to codify our findings into iac.

PR #435, opened 2021-02-26
Downgrading metrics db to Basic tier with other iac improvements

echappen:
Resolves [#338](https://github.com/18F/piipan/issues/388)

Doing this made me realize the virtual network in my metrics.json template was unnecessary. I think this was residual from pulling this example ARM template from Azure docs.

### DIFF
--- a/iac/arm-templates/metrics.json
+++ b/iac/arm-templates/metrics.json
@@ -89,41 +89,6 @@
                             "metadata": {
                                 "description": "Server Name for Azure database for PostgreSQL"
                             }
-                        },
-                        "subnetName": {
-                            "type": "string",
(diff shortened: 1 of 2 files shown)


ryanhofdotgov (approved):
👍 I think we may actually add vnets back in to all dbs, but we will do so with a systematic approach.

ryanhofdotgov:
Related ticket: https://github.com/18F/piipan/issues/11

echappen:
> 👍 I think we may actually add vnets back in to all dbs, but we will do so with a systematic approach.

I kept getting errors like "Service not compatible with Basic tier" on the vnet deployments when attempting to downgrade the db with vnets in place. Not sure if this means vnets are incompatible with the Basic tier.

Issue #425, opened 2021-02-25
Establish the local development story for API auth


kingcomma:
As a developer, I want a coherent approach to local development that depends on Piipan's secured APIs so I can build and test related sub-systems (e.g., query tool, orchestrator API, metrics dashboard) outside of Azure.

This work likely includes:
- [ ] Updating `Piipan.Shared.Authentication` to support retrieving tokens as a developer's Azure user
- [ ] Helper scripts for assigning a developer's Azure user necessary application roles
- [ ] Examples for conditionally using different token providers depending on the environment context

PR #437, opened 2021-03-01
Add local development approach to orchestrator API

kingcomma:
Establishes the approach for #425. Specifically:
- Adds CLI-based token provider
- Adds helper script for assigning app roles
- Updates orchestrator to use appropriate token provider for the current configuration
- Documents local development for orchestrator
- Updates IaC to allow app role assignment to users

All of the above has been done in the context of `Piipan.Match.Orchestrator`, but I'm thinking this same approach applies to any Azure resource acting as a client of one of our internal APIs.

Follow-up work:
- Add a helper script for adding the Azure CLI as an authorized client application to an existing application object (requires more research)

### DIFF
--- a/match/docs/orchestrator-match.md
+++ b/match/docs/orchestrator-match.md
@@ -29,14 +29,19 @@ The orchestrator treats per-state APIs as backing services. When running the [Ia
 - Per-state URIs are compiled into a JSON list and saved as an environment variable
 - The orchestrator's system-assigned identity is given an authorized application role (which will be checked by the state API upon receiving requests)
 
-At runtime, the app requests an authentication token from the state app's Active Directory app registration. The token is then included as an authorization header (`Authorization: Bearer {token}`) in the request sent to the state API.
+At runtime, the app requests an authentication token from the state app's Active Directory application object. The token is then included as an authorization header (`Authorization: Bearer {token}`) in the request sent to the state API. For more detail on this process see [Securing internal APIs](../../docs/securing-internal-apis.md).
 
 ## Local development
 
+Local development is achieved by connecting a locally running instance of the orchestrator API to remote instances of the state APIs. When running locally, `Startup.cs` conditionally configures the `Piipan.Shared.Autentication.AuthorizedJsonApiClient` dependency to use Azure CLI credentials when obtaining access tokens. To make use of this functionality:
+
+1. Adjust the orchestrator's local configuration by adding `"DEVELOPMENT": "true"` to the `Values` object in `local.settings.json`.
+1. Assign your user account the `StateApi.Query` role for each state the orchestrator queries. Assignment can be done using the [`tools/assign-app-role.bash`](../../tools/assign-app-role.bash) script.
+1. Ensure the Azure CLI has been added as an authorized client application to each state API's application object. This can be done via the Azure Portal at Azure Active Directory > Application registrations > {State API's application object} > Expose an API > Add a client application and adding the client ID `04b07795-8ddb-461a-bbee-02f9e1bf7b46` (the global identifier for the Azure CLI) with the `user_impersonation` scope. A helper script for this action does not yet exist.
+
+With the orchestrator running locally (`func start` or `dotnet watch msbuild /t:RunFunctions`), any requests to the local endpoint will now use the user account authorized with Azure CLI to obtain access tokens from the state APIs.
+
 A true local development approach with locally run instances of the per-state APIs and participant records database does not yet exist.
-
-Until then, local development is limited by the need to authenticate with Active Directory before accessing state APIs. The Instance Metadata Service used to retrieve authentication tokens is not available locally. There are [potential solutions](https://docs.microsoft.com/en-us/dotnet/api/overview/azure/service-to-service-authentication#local-development-authentication) using the `Microsoft.Azure.Services.AppAuthentication` library. None have been implemented at this time.
-
 
 ### App deployment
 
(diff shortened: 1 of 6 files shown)


tools/assign-app-role.sh
```diff
@@ -0,0 +1,60 @@
+#!/bin/bash
```
ryanhofdotgov, line 1: suggestion: Right now all the other bash scripts use .bash as the file extension to indicate they really need bash, and aren't POSIX-compliant shell scripts. We can revisit that, but just for consistency for now, I'd rename from .sh to .bash

match/docs/orchestrator-match.md
```diff
@@ -29,14 +29,19 @@ The orchestrator treats per-state APIs as backing services. When running the [Ia
...
+1. Ensure the Azure CLI has been added as an authorized client application to each state API's application object. This can be done via the Azure Portal at Azure Active Directory > Application registrations > {State API's application object} > Expose an API > Add a client application and adding the client ID `04b07795-8ddb-461a-bbee-02f9e1bf7b46` (the global identifier for the Azure CLI) with the `user_impersonation` scope. A helper script for this action does not yet exist.
 
-Until then, local development is limited by the need to authenticate with Active Directory before accessing state APIs. The Instance Metadata Service used to retrieve authentication tokens is not available locally. There are [potential solutions](https://docs.microsoft.com/en-us/dotnet/api/overview/azure/service-to-service-authentication#local-development-authentication) using the `Microsoft.Azure.Services.AppAuthentication` library. None have been implemented at this time.
+With the orchestrator running locally (`func start` or `dotnet watch msbuild /t:RunFunctions`), any requests to the local endpoint will now use the user account authorized with Azure CLI to obtain access tokens from the state APIs.
```
ryanhofdotgov, line 42: I ran into some problems here, late in the day so I will punt back to you since it's been a few weeks since I've looked at this and I've forgotten how to deal with this 🙂 

Context:
I recreated my local.settings.json file from the portal, it looks like this:
```
{
  "IsEncrypted": true,
  "Values": {
    "DEVELOPMENT" : "true",
    "AzureWebJobsStorage": "foo",
    "FUNCTIONS_WORKER_RUNTIME": "foo",
    "FUNCTIONS_EXTENSION_VERSION": "foo",
    "WEBSITE_CONTENTAZUREFILECONNECTIONSTRING": "foo",
    "WEBSITE_CONTENTSHARE": "foo",
    "WEBSITE_RUN_FROM_PACKAGE": "foo",
    "APPINSIGHTS_INSTRUMENTATIONKEY": "foo",
    "StateApiUriStrings": "foo"
  },
  "ConnectionStrings": {}
}
```
where `foo` is an encrypted string.

When I run `func start`, I first get a complaint that `Can't determine project language from files`. Adding  `--dotnet` fixes this, but I thought there was a way to just tell it to do this in the project file?

Moving past that, I get:
```
Failed to decrypt settings. Encrypted settings only be edited through 'func settings add'.
```
which makes sense, but... not sure how we make this easy.

match/docs/orchestrator-match.md
```diff
@@ -29,14 +29,19 @@ The orchestrator treats per-state APIs as backing services. When running the [Ia
...
+At runtime, the app requests an authentication token from the state app's Active Directory application object. The token is then included as an authorization header (`Authorization: Bearer {token}`) in the request sent to the state API. For more detail on this process see [Securing internal APIs](../../docs/securing-internal-apis.md).
 
 ## Local development
 
-A true local development approach with locally run instances of the per-state APIs and participant records database does not yet exist.
+Local development is achieved by connecting a locally running instance of the orchestrator API to remote instances of the state APIs. When running locally, `Startup.cs` conditionally configures the `Piipan.Shared.Autentication.AuthorizedJsonApiClient` dependency to use Azure CLI credentials when obtaining access tokens. To make use of this functionality:
+
+1. Adjust the orchestrator's local configuration by adding `"DEVELOPMENT": "true"` to the `Values` object in `local.settings.json`.
+1. Assign your user account the `StateApi.Query` role for each state the orchestrator queries. Assignment can be done using the [`tools/assign-app-role.bash`](../../tools/assign-app-role.bash) script.
```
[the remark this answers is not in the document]
kingcomma: Agreed — I had originally thought this script could support `<state-abbr>` as an argument instead of `<function-name>`, or even have a wrapper `assign-to-states.bash` script that removed the need to specify individual functions. But I ran into the issue discovering the state functions outside of the IaC, and the only path I could see given our current structure was hardcoding our circumstantial, somewhat arbitrary names.

I wonder if [tags](https://docs.microsoft.com/en-us/azure/cloud-adoption-framework/decision-guides/resource-tagging/) are a potential solution?

tools/assign-app-role.sh
kingcomma: Renamed in a3bde5c. I intended for this to be a .bash file from the get go — must have unknowingly lost a fight with muscle memory.

tools/assign-app-role.sh
```diff
@@ -0,0 +1,60 @@
...
+#
+# Assigns the current CLI user to the specified application role of the
+# specified Azure Function. Used to enable the use of Azure CLI credentials
+# (i.e., `Azure.Identity.AzureCliCredential`) when connecting to remote APIs
+# which have been secured using App Service Authentication. Only intended
+# for development environments.
+#
+# usage: assign-app-role.bash function-name role-name
+
+source common.bash || exit
```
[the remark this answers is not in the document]
kingcomma: Suggestion applied in a3bde5c.

match/docs/orchestrator-match.md
kingcomma: Happy to receive the punt! I clarified the local development instructions in a3bde5c — see the updated [`orchestrator-match.md`](https://github.com/18F/piipan/pull/437/commits/a3bde5cb469585977419b2255e732445cd8475e5#diff-69bc1db9de815f1e2fec90cc8d940ef24ac3afa5518f74ae012f6ebef4b0a9dc).

When I start from scratch and recreate `local.settings.json` using `func azure functionapp fetch-app-settings`, I'm able to get running locally using just `func start`. I.e., I did not run into the runtime error you saw. The `FUNCTIONS_WORKER_RUNTIME` setting that comes down with `fetch-app-settings` should already be set to `dotnet` and that should make it unnecessary to use `--dotnet` every time you start up the function. I'm not sure what would have caused you to get that error.

I did run into the same encryption issue, however. There are probably multiple ways to handle this, including using `func settings decrypt` to convert all the encrypted strings and allow manual edits. The approach I specified in the updated docs is using `func settings add`. It's a single step that will work whether or not the settings are encrypted. 

Do you mind stepping through this again and seeing if the issues are resolved?

ryanhofdotgov: Thanks – this worked! I didn't need the `--dotnet` parameter this go around. I suspect some local state is being altered in one of the files that we don't check into GitHub. This quirk is particularly annoying since the error message doesn't actually include the option that is correct.

ryanhofdotgov (approved):
Nice work!

PR #440, opened 2021-03-02
feat: Add authentication to Query Tool

heymatthenry:
Integrate an `AuthorizedJsonApiClient` into the Query Tool

### DIFF
--- a/query-tool/src/Piipan.QueryTool/Startup.cs
+++ b/query-tool/src/Piipan.QueryTool/Startup.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Net.Http;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting;
@@ -8,6 +9,7 @@ using Microsoft.AspNetCore.HttpsPolicy;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Hosting;
+using Piipan.Shared.Authentication;
 
 namespace Piipan.QueryTool
 {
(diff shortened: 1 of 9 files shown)


query-tool/tests/Piipan.QueryTool.Tests/OrchestratorApiRequestTest.cs
```diff
@@ -43,12 +58,13 @@ public async void TestQueryOrchestrator()
             }";
             var handlerMock = MockHttpMessageHandler(mockResponse);
             var httpClient = new HttpClient(handlerMock.Object);
-
+            var token = "|token|";
+            var tokenProvider = MockTokenProvider(token);
```
kingcomma, line 62: **Suggestion** (not blocking): Tying my other two comments together — If `AuthorizedJsonApiClient` becomes the only dependency for the classes in this test, it should be possible to use a only a single mocked instance of the api client that is setup to return a JSON response.

The orchestrator's unit tests probably should have done this, but, similar to here, the bootstrapping was already in place to mock all the `HttpClient` bits and it was quicker to keep things as is and defer optimization to a later sprint.

kingcomma (commented):
Nice! Especially since you had to fly blind without an established local development process.

My comments all center on utilizing dependency injection and moving the various services (`EasyAuthTokenProvider`, `AuthorizedJsonApiClient`, `HttpClient`) to a single entry point. Since this PR works as is, my suggestions may not be required strictly speaking. But, if you agree, I think it may be worth addressing sooner than later.

heymatthenry:
@kingcomma This has been a journey, but a worthwhile one! Per your comments, the `AuthorizedJsonApiClient` is now coming in through DI, and in implementing that, I was able to get rid of that gnarly overloaded method.

As to the small refactor you suggested, I think you're right but have held off for now until there are some more tests in there to inform the refactor (and to be able to get your 👀 on this PR again sooner).

query-tool/src/Piipan.QueryTool/Startup.cs
```diff
@@ -1,13 +1,15 @@
...
 using System.Linq;
+using System.Net.Http;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.AspNetCore.HttpsPolicy;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Hosting;
+using Piipan.Shared.Authentication;
```
kingcomma, line 12: **Nitpick** (non-blocking): looks like a few of these imports can be cleaned out the next time you're working here.

kingcomma (approved):
🎉 Thanks for taking the journey — looks great!

I left a small nitpick for the next time you're in `Startup.cs` but it shouldn't hold up merging.

Issue #408, opened 2021-02-24
Rename Metrics subsystems to align with naming convention


echappen:
Now that we have multiple function apps for metrics (for example, the api), we should rename the original function app that collects metrics from other systems to something more specific, as well as other components of the metrics system.

This will entail updating the ARM templates and other scripts that create these resources, and recreating them in the Azure dev environment. 

- Align with naming convention pattern, coordinate with @ryanhofdotgov

echappen:
@ryanhofdotgov Was searching for our written comms about naming conventions and found this in our running meeting notes with FNS: 

> Use global FNS prefix (“FN”) on everything. Can also use “RG-FNS” for resource groups. That’s how USDA incident responders easily identify components as belonging to FNS


Is there documentation elsewhere I should also be referencing?

echappen:
Also referencing https://docs.microsoft.com/en-us/azure/cloud-adoption-framework/ready/azure-best-practices/resource-naming

ryanhofdotgov:
I'm tracking the larger renaming effort in https://github.com/18F/piipan/issues/252, which also links:
https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/resource-name-rules

My [super early work in progress is here](https://docs.google.com/spreadsheets/d/1pYdvVoICjQjx9RvyYhuoV8D0xCrVRRQzYNhpdNvFsyc/edit#gid=1351188795), where I'm working out what resources need to be globally unique, and what have restrictive length limits/char restrictions.

I'm not a fan of the agency prefix the partner has asked to have on all resources; for our TTS sandbox I think we won't use them.

echappen:
@ryanhofdotgov Here's what I'm thinking for Metrics resource names. How does this look to you?

| purpose | current name | final name |
| ------- | ------------ | ---------- |
| resource group | piipan-metrics | fns-rg-piipan-metrics-dev-westus-001 |
| secret keeper | metrics-secret-keeper | fns-kv-metrics-dev-westus-001 |
| database | piipan-metrics-db | fns-db-metrics-dev-westus-001 |
| Metrics API function app | PiipanMetricsApi[unique-string] | fns-func-metricsapi-dev-westus-001 |
| Metrics Collections function app | PiipanMetricsFunctions[unique-string] | fns-func-metricscollect-dev-westus-001 |
| Application Insights for API app | PiipanMetricsApi[unique-string] | fns-insights-metricsapi-dev-westus-001 |
[...]

Issue #252, opened 2021-01-21
Revisit Azure resource naming conventions


ryanhofdotgov:
We need to support multiple, non-production deployment environments within a single subscription. Revisit our resource naming conventions, group names, access control, etc. to support multiple deployment environments.

As we revisit how we identify our resources, keep [Microsoft's recommended practices in mind](https://docs.microsoft.com/en-us/azure/cloud-adoption-framework/ready/azure-best-practices/resource-naming).

ryanhofdotgov:
Also: adhere to [partner agency decision/discussion here](https://docs.google.com/document/d/1m4esX7gtH2SXrpUz0j_JKJRLSoOwjLoxJlBokXWkWHU/edit#).

ryanhofdotgov:
Also see:
https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/resource-name-rules

echappen:
@ryanhofdotgov  Are we dropping the name "piipan" once we move into the partner's Azure environment? (referencing [this spreadsheet](https://docs.google.com/spreadsheets/d/1pYdvVoICjQjx9RvyYhuoV8D0xCrVRRQzYNhpdNvFsyc/edit#gid=1351188795))

For example: `piipan-metrics` would become `fns-rg-metrics-dev`

ryanhofdotgov:
That's right – the goal is to keep Piipan the project somewhat separate from its corresponding instance at the partner agency in order to help maximize reuse in the future for other benefit programs.

I think that means:
* allowing a configurable prefix for resources
* allowing our resource groups to be named differently, and configured external to the main scripts
* dropping "piipan" when we don't need it for disambiguation to keep names short; i.e., everything in the subscription is piipan.

So I am anticipating a set of infrastructure config files for TTS dev/TTS test/partner dev/partner test/partner preprod/partner prod. We may just keep the partner config files just in the partner source code control system.

ryanhofdotgov:
Good chunk of this work was completed in #442, #408 and the yet-to-be-merged PR #451.

Don't finalize until we incorporate lessons from deploying to Azure Government.

PR #442, opened 2021-03-02
Renames metrics resources

echappen:
Partially addresses #408

Renames the following for clarity and consistency with other subsystems:
PiipanMetricsFunctions -> Piipan.Metrics.Collect
PiipanMetricsApi -> Piipan.Metrics.Api

Also renames metrics resources to fit new naming convention:
fns-[resource_type]-[app_name]-[env]-[azure_location]

Some resources don't allow hyphens (KeyVault, Storage Accounts) so hyphens are just removed from regular naming convention for those. 

Also adds a new environment variable for key vault name to both Metrics.Collect and Metrics.Api apps, since that name will be dynamic going forward. 

Lots of the files added here have to do with the renaming.

Files relevant for review:
- create-metrics-resources.bash
- iac-common.bash
- function-metrics.json

Note: I need a better way of constructing the resource names to account for no prefix ("fns-"). For now I kept it on there as a configurable item.


echappen:
@ryanhofdotgov I reverted the deployment of Metrics.Collect back to cli, since the arm template was preventing the event subscriptions from firing in the app for some reason I still can't figure out.

iac/create-metrics-resources.bash
```diff
@@ -13,18 +13,22 @@ source $(dirname "$0")/iac-common.bash || exit
...
+DB_SERVER_NAME=${PREFIX}-db-metrics-${ENV}-${LOCATION}
 DB_ADMIN_NAME=piipanadmin
 DB_NAME=metrics
 DB_TABLE_NAME=participant_uploads
+# Needed for both function apps
+DB_CONN_STR=`pg_connection_string $DB_SERVER_NAME $DB_NAME $DB_ADMIN_NAME`
 # Name of Key Vault
-VAULT_NAME=metrics-secret-keeper
+VAULT_NAME_KEY=KeyVaultName
+VAULT_NAME=${PREFIX}kvmetrics${ENV}${LOCATION} # vault names can't use hyphens even though the docs say they can
```
ryanhofdotgov, line 24: I think the problem might be something else, as the other KeyVault we use is named `secret-keeper`? The limit is 24 characters though.

echappen: mmm true. I think I was confused by their error message —it said something like "must use alphanumerics" instead of "alphanumerics and hyphens" like in the docs.

ryanhofdotgov:
Thinking about it a bit, maybe the best way for us to get our PRs in a compatible state is for:
* to merge in https://github.com/18F/piipan/pull/447 to main
* merge in main into this
* add the prefix/env to the externalized config files
* temporarily define our TTS sandbox to use the agency prefix for the metric sub-system (i.e., match what is deployed right now).

I'll create a PR for this to merge into `408-metrics-iac-refactors` for you to look at in the morning.

Then in follow-up actions:
* make PREFIX optional (i.e., it could be an empty string), remove it from our TTS deployment configs. It seems like it would be line noise in most Azure environments.
* finalize https://github.com/18F/piipan/pull/451
* come back and tweak the names of resources in metrics sub-system (if any changes are made in #451). The only one I haven't resolved in my mind and we haven't discussed, is location actually useful when resources are all in the same place? 

Sorry this ended up in this confusion – I had started down the path of establishing the top-level naming conventions in https://github.com/18F/piipan/issues/252, but then realized it wasn't really required to get our partner environment set up and I fell behind your work here.

ryanhofdotgov:
@echappen After bungling the PR I promised above, in the interest of simplicity I just committed directly to this PR's branch. If those 2 commits look good to you, I'm 👍 with merging this PR.

echappen:
Thanks for doing that extra work @ryanhofdotgov

echappen:
Re-ran the script with your changes, tested subscriptions and api endpoint. All looks good to me.

PR #444, opened 2021-03-02
Wrap create-resources in a main function

ryanhofdotgov:
Will be easier to review with ignore white space option, but since I wanted to avoid nested functions in bash, there's a lot of line changes for that required code move.


kingcomma (approved):
👍

ryanhofdotgov:
Partially addresses #101

PR #447, opened 2021-03-03
Externalize agency/subscription/deployment config

ryanhofdotgov:


echappen:
@ryanhofdotgov I might be stepping on your toes a bit here with this WIP PR I put up yesterday where I do some more serious renaming of resources in metrics iac:

https://github.com/18F/piipan/pull/442

Would be up for pairing on how to integrate this with what you've been doing.

kingcomma (approved):
👍 Changes look good to me. Happy to re-review after you and @echappen discuss, if needed.

ryanhofdotgov:
Partially addresses #101

PR #451, opened 2021-03-03
Begin documenting our naming conventions for Azure resources

echappen:
I added only dev and prod environments for now, thinking we'd add columns as we create new ones.
Anything undetermined is listed as TBD or tbd in the charts.


ryanhofdotgov:
@echappen I pushed some edits in 3aca978 to avoid implying we are publishing partner specific config values into `main`. Since both you and I are contributors to the PR, I can't add you as a reviewer for that commit and have added @kingcomma.

kingcomma (approved):
👍

PR #459, opened 2021-03-05
Move resource group creation to separate script

ryanhofdotgov:
This will be more important in environments where we do not have permission to create the resource groups ourselves.


kingcomma (approved):
👍

ryanhofdotgov:
Partially addresses #101

PR #465, opened 2021-03-08
feat: Query tool error handling

heymatthenry:
If the query tool app errors while trying to retrieve data from the Orchestrator API, it now presents an error message rather than crashing:

image


query-tool/src/Piipan.QueryTool/Pages/Index.cshtml
```diff
@@ -47,7 +54,7 @@
             <input class="usa-button" type="submit" value="Submit" />
         </fieldset>
     </form>
-    @if (Model.QueryResult.Any())
+    @if (Model.QueryResult != null && Model.QueryResult.Any())
```
echappen, line 57: Out of curiosity, is it possible for `Model.QueryResult` to ever be null if it's not specified as nullable in the page class?

echappen (approved):

query-tool/src/Piipan.QueryTool/Pages/Index.cshtml
heymatthenry: It can. I think(?) in the error case it's getting called by the view before it's initialized.

PR #486, opened 2021-03-09
Example of automated a11y testing in CI

echappen:
This is a proof-of-concept for integrating automated a11y scanning into our CI pipeline.

It creates a new Circle Job called "accessibility" which will run in parallel to the "test" and "build" jobs.

The job uses a dotnet docker image because it needs to run a local version of a dotnet app for pa11y to test.

In the job, node/npm and supporting packages for puppeteer (which powers the headless browsing for pa11y) are installed. Then it runs the app, waits for the app to spin up, then runs pa11y-ci on localhost:5000.

If any pa11y errors are found, they will be logged in the Circle job and the job will fail.

### For consideration:
- Does this even look like a worthwhile approach? There's quite a bit of setup involved.
- Currently this job only runs pa11y for the Dashboard app. If we decide to go this route, I'm thinking it'd be simpler to config separate apps as separate jobs 
- Would something like [Accesslint](https://github.com/marketplace/accesslint) be more efficient? (if it can decipher dotnet-y html templates)
- As the apps grow in complexity (data seeding, auth, etc) We might need to switch to running a [Pa11y Actions](https://github.com/pa11y/pa11y/blob/5724332e44f30b312261d0eb2a560ce23e6e0697/TROUBLESHOOTING.md#how-can-pa11y-log-in-if-my-sites-behind-basic-auth) script.
- There's probably a better way to wait until the dotnet app is spun up than by using `sleep`
- I didn't add the accessibility job to the deploy pipeline, didn't seem necessary as this job will run on every code push


heymatthenry:
I do wonder if it might be easier to use something like [this](https://marketplace.visualstudio.com/items?itemName=DrewLewis.Accessibility) as part of an Azure DevOps pipeline as described [here](https://medium.com/@drewelewis/accessibility-testing-with-azure-devops-pipelines-3d46291bef15). Even if we didn't want to use that specific tool, making our own Azure DevOps pipeline that used pa11y might be more Microsoft-y.

ryanhofdotgov:
@heymatthenry Unfortunately Azure DevOps is not part of Azure Government (yet).

ryanhofdotgov:
Because of the involved setup time, I think I'd be against running it against every PR. Running it nightly would be my preference considering the effort/benefit right now.

We could also explore:
* Only triggering it if something under the dashboard tree was altered?
* Expend the effort to create a custom docker image and host it up in Docker Hub – that's a whole other can of worms, including https://github.com/18F/tts-tech-portfolio/issues/983, and isn't super portable to the partner environment.

echappen:
@ryanhofdotgov I couldn't find a straightforward way to only trigger builds when a subdirectory changes, so I set both a11y jobs to run against the `main` branch on a nightly schedule. 

The subdirectory trigger may still be worth exploring as our subsystems get larger. I'm experimenting with trying a git diff command like [this](https://medium.com/@reuvenharrison/using-circleci-with-monorepos-c04b03f9f39c) but it's still pretty buggy.

ryanhofdotgov (approved):
Cool! 

One thing that occurs to me – it would be good to try and reduce the repetition across `accessibility_dashboard` and `accessibility_query_tool`. [CircleCI's reusable commands ](https://circleci.com/docs/2.0/reusing-config/#authoring-reusable-commands)might be the ticket here.

I'll leave it to you to decide whether to address that in this PR, or defer it to https://github.com/18F/piipan/issues/509 (which we should pick up next sprint).

echappen:
@ryanhofdotgov Good idea! Done [here](https://github.com/18F/piipan/pull/486/commits/e8997fe832a2123ca27248d5e3b79ed925a3f16d).

PR #491, opened 2021-03-10
Wait a bit before publishing func to new func app

ryanhofdotgov:
Waiting 60 seconds before publishing a Function to a new Function App has worked well so far, as Azure will report a Function App has been deployed when it actually hasn't fully deployed.

macOS sleep does not support units, so drop the "s" in "60s" on existing sleep uses.


ryanhofdotgov:
Partially addresses #101

echappen (approved):

Issue #478, opened 2021-03-09
Query and dashboard website: Debug logging for API interactions


cjhskippy:
As a developer, I would like to be able to deploy the query tool and dashboard to new environments, configure debug/trace level logging, and be able to see a trace of the execution so I can more easily identify where things are breaking.

- query tool
  - [x] API error messages are saved in logs that can be reviewed by Piipan developers
  - [x] Error messages for queries do not record PII
 - dashboard
   - [x] API error messages are saved in logs that can be reviewed by Piipan developers
   - [x] Error messages for queries do not record PII

ryanhofdotgov:
@heymatthenry I updated the title and acceptance criteria checkboxes to match the original issue description which included both the query tool and the dashboard in its scope. We can also split up and schedule for next sprint if you'd prefer that.

heymatthenry:
@ryanhofdotgov Nope, happy to finish this up in this sprint.

PR #494, opened 2021-03-11
Add log statements to query tool

heymatthenry:
Along the way to implementing #478, this PR just ensures something is being logged so I can configure logging in the Azure portal.


ryanhofdotgov (approved):
Sweet! Having the structure in place will also make it easier to add quick log entries while debugging problems.

PR #496, opened 2021-03-11
Add tool for updating deps faster

ryanhofdotgov:


tools/update-packages.bash
```diff
@@ -0,0 +1,51 @@
+#!/bin/bash
+#
+# Update package dependencies for C# projects (.csproj), recursively
+# from the indicated path.
+# Only minor updates are considered, unless --highest-major is specified.
+#
+# usage: update-packages.bash <path> [--highest-major]
+
+#source $(dirname "$0")/../tools/common.bash || exit
+source /Users/ryanmhofschneider/projects/scratch/piipan/tools/common.bash || exit
```
echappen, line 10: @ryanhofdotgov Still have a local path here

echappen (changes requested):

tools/update-packages.bash
ryanhofdotgov: Great catch! Resolved in 34eb1ba

echappen (approved):

PR #500, opened 2021-03-12
Use cloud-specific resource id w/managed identity

ryanhofdotgov:
Commercial and Government Azure use different ossrdbms-aad resource ids; this id needs to be specified when accessing PostgreSQL via a managed identity. Bind the cloud name during infrastructure setup as an environment variable, then use that to choose the right id for the active cloud in the ETL function.


ryanhofdotgov:
Partially addresses #101

kingcomma (approved):
👍

PR #501, opened 2021-03-12
Don't use double quotes around orch api URI

ryanhofdotgov:
Resolves "Invalid URI: The URI scheme is not valid." error in query tool when it attempts to call out to the orchestrator API endpoint.


heymatthenry (approved):
Thanks for doing this!

PR #505, opened 2021-03-12
update names of resource groups in tts dev

echappen:


echappen:
@ryanhofdotgov These were some minor changes I found while trying to perform bulk uploads, didn't know if we wanted to formally change these names.

ryanhofdotgov:
Ah, sorry – I've been making lots of small, self-contained changes. This change has already been merged in with https://github.com/18F/piipan/pull/498

PR #508, opened 2021-03-12
Remove service principal creation from main IaC

ryanhofdotgov:
Don't create service principal in main IaC because:
- not necessary in all deployment environments
- resets the SP secret every time IaC runs, which requires it to be updated in CircleCI


ryanhofdotgov:
Partially addresses #101

kingcomma (approved):
👍

PR #511, opened 2021-03-15
Add logging to dashboard app

heymatthenry:
Adding some logging to the dashboard, just like the query tool got in #494.

### DIFF
--- a/dashboard/src/Piipan.Dashboard/Pages/ParticipantUploads.cshtml.cs
+++ b/dashboard/src/Piipan.Dashboard/Pages/ParticipantUploads.cshtml.cs
@@ -5,6 +5,7 @@ using System.Threading.Tasks;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.AspNetCore.Mvc.RazorPages;
 using Microsoft.AspNetCore.WebUtilities;
+using Microsoft.Extensions.Logging;
 using Piipan.Dashboard.Api;
 
 #nullable enable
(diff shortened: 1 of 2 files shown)


dashboard/src/Piipan.Dashboard/Pages/ParticipantUploads.cshtml.cs
```diff
@@ -40,26 +44,50 @@ public async Task OnGetAsync()
 
         public async Task<IActionResult> OnPostAsync()
         {
+            _logger.LogInformation("Querying uploads via search form");
+
             if (BaseUrl == null)
+            {
+                _logger.LogError("BaseUrl is null");
```
ryanhofdotgov, line 51: I think there might be an inconsistent exception handling approach here? Below you catch exceptions, log them, and continue to return the page. I might just put this check in the same try block?

dashboard/src/Piipan.Dashboard/Pages/ParticipantUploads.cshtml.cs
```diff
@@ -40,26 +44,50 @@ public async Task OnGetAsync()
...
+            }
             return Page();
         }
 
         // adds default pagination to the api url if none is present from request params
         private string FormatUrl()
         {
             if (BaseUrl == null)
+            {
+                _logger.LogError("BaseUrl is null");
```
ryanhofdotgov, line 77: It looks like there may be  another exception handling mismatch here in this method, though in this case I'd be inclined to do the opposite since it isn't rendering a page – just let the exception bubble up the call stack.

heymatthenry: @ryanhofdotgov Yes, good call. The error handling here is now left to the `OnGetAsync` and `OnPostAsync` methods.

ryanhofdotgov (approved):
👍

PR #516, opened 2021-03-16
Update dependencies

kingcomma:
- Bump FluentValidation

### DIFF
--- a/match/src/Piipan.Match.Orchestrator/Piipan.Match.Orchestrator.csproj
+++ b/match/src/Piipan.Match.Orchestrator/Piipan.Match.Orchestrator.csproj
@@ -6,7 +6,7 @@
     <RestoreLockedMode Condition="'$(ContinuousIntegrationBuild)' == 'true'">true</RestoreLockedMode>
   </PropertyGroup>
   <ItemGroup>
-    <PackageReference Include="FluentValidation" Version="9.5.2" />
+    <PackageReference Include="FluentValidation" Version="9.5.3" />
     <PackageReference Include="Microsoft.Azure.Functions.Extensions" Version="1.1.0" />
     <PackageReference Include="Microsoft.Extensions.Http" Version="3.1.13" />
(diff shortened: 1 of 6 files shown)


ryanhofdotgov:
@kingcomma ICYMI, one thing I added last week was [update-packages.bash](https://github.com/18F/piipan/blob/main/docs/update-deps.md#semi-automated-steps). Not sure if you tried it out for this PR, it's more useful for large numbers of updates.

ryanhofdotgov (approved):

kingcomma:
@ryanhofdotgov Saw it and used it, it's a great improvement!

One note: I had to run the tool twice because of a combination of circumstances and the order of the output of `find "$start_path" -name "*.csproj"`. In this case:

- The dependency needing updating (`FluentValidation`) is a "direct" dependency for `match/src/Piipan.Match.State/Piipan.Match.State.csproj`
- The dependency needing updating is a "transitive" dependency for `match/tests/Piipan.Match.State.Tests/Piipan.Match.State.Tests.csproj`
- Because of the creation dates for `src` and `tests` (I think), `match/src/Piipan.Match.State/Piipan.Match.State.csproj` appears before `match/tests/Piipan.Match.State.Tests/Piipan.Match.State.Tests.csproj` in the output of `find [...]`
- Thus, `Piipan.Match.State.Tests.csproj` is updated and restored *before* `Piipan.Match.State.csproj`

Because transitive dependencies will only update *after* their associated direct dependencies are updated, the result is an up-to-date `Piipan.Match.State.csproj` but a still out of date `Piipan.Match.State.Tests.csproj`. Running the script a second time catches those out of date dependencies.

If that all makes sense, do you have any qualms with me updating the script to do a `sort` on the output of `find`? Since we're following a consistent directory structure, that should mean `src` projects are always updated before `tests` projects.

ryanhofdotgov:
@kingcomma Sounds good to me. Thanks!

PR #517, opened 2021-03-16
Script to delete resources from resource groups

ryanhofdotgov:
This allows us to test the IaC from a clean slate more easily. It does not remove dangling app registrations in Active Directory.


kingcomma (approved):
👍

Issue #481, opened 2021-03-09
Extend the spec for the Orchestrator API to include match ID


cjhskippy:
Precursor to #452 

- Update documented spec for the Orchestrator API to return a match ID.

echappen:
@cjhskippy Was this for the query endpoint to return a Match ID _instead_ of the Pii Records, or in addition to a Pii Record?

In other words, returning something like this for a match query:

<img width="324" alt="Screen Shot 2021-03-16 at 11 25 43 AM" src="https://user-images.githubusercontent.com/4267629/111344609-8317a980-864a-11eb-8a10-b9fefe8d08fc.png">

Instead of this:

<img width="337" alt="Screen Shot 2021-03-16 at 11 23 16 AM" src="https://user-images.githubusercontent.com/4267629/111344651-8dd23e80-864a-11eb-9e56-fcddd125968e.png">

ryanhofdotgov:
@echappen the match id would be in addition to the existing Pii record data in the response.

echappen:
@ryanhofdotgov Thanks, so each Pii Record would have 1 Match ID like this?

<img width="297" alt="Screen Shot 2021-03-16 at 1 24 14 PM" src="https://user-images.githubusercontent.com/4267629/111360625-f9240c80-865a-11eb-897a-9a63df2f98fe.png">

ryanhofdotgov:
Almost – the match id encapsulates the input of the query, so it is a top-level property, rather than associated with each and every PII record. If there's a match there's most likely only one other state, but there could be several and we wouldn't repeat the match id on each.

kingcomma:
Is there a scenario where the match ID would still be part of the PII record, even if it's shared across multiple? I'm probably getting ahead of myself, but thinking of a scenario where the orchestrator API is extended to support batch/bulk querying (#368).

ryanhofdotgov:
Without looking at the schema, I would say that the match ID is 1:1 to search tuples (fn,mn,ls,dob,ssn) that have matches. In a bulk scenario, I submit a search for 10 tuples, and 3 have matches, I'm going to get 3 match ids across my response.

PR #521, opened 2021-03-16
add match id as spec to open api docs for orchestrator

echappen:
Addresses #481

### DIFF
--- a/match/docs/openapi/schemas/pii-record.yaml
+++ b/match/docs/openapi/schemas/pii-record.yaml
@@ -32,3 +32,6 @@ PiiRecord:
     exception:
       type: string
       description: "Placeholder for value indicating special processing instructions"
+    match_id:
+      type: string
+      description: "the identifier of the match"
--- a/match/docs/openapi/schemas/pii-record.yaml
+++ b/match/docs/openapi/schemas/pii-record.yaml
@@ -32,6 +32,3 @@ PiiRecord:
     exception:
       type: string
       description: "Placeholder for value indicating special processing instructions"
-    match_id:
-      type: string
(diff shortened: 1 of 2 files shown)


match/docs/openapi/schemas/pii-record.yaml
```diff
@@ -32,3 +32,6 @@ PiiRecord:
     exception:
       type: string
       description: "Placeholder for value indicating special processing instructions"
+    match_id:
+      type: string
+      description: "the identifier of the match"
```
kingcomma, line 37: My past work is creating a bit of an issue here 😳 .

When the OpenAPI specs for the orchestrator and state APIs were created, they shared identical `PiiRecord` models and it made sense to me to pull them out into a single YAML resource (this file). As a result, this update will include a `match_id` property in the `PiiRecord` models for both the state and orchestrator API specs. This is mitigated a bit since the state specs are for internal usage, but it should probably be remedied for the sake of accuracy.

Since the specs are only diverging by a single property, maybe the solution is to use the `allOf` keyword [described here](https://swagger.io/docs/specification/data-models/inheritance-and-polymorphism/) in the orchestrator spec?

kingcomma: Noting for posterity that this discussion is also happening on [the issue](https://github.com/18F/piipan/issues/481) itself.

echappen: I think this may be moot anyway — I was mistaken on what Match ID actually represents.

echappen:
Per our discussion I've changed `match_id` to `lookup_id`

kingcomma (approved):
👍

PR #524, opened 2021-03-17
Add SysTag resource tag to key system components

ryanhofdotgov:
This tag is used to ease enumeration of resource instances in IaC and to make a resource's system-level purpose more obvious in the Azure Portal, especially when naming restrictions cause names to be somewhat inscrutable.


kingcomma (approved):
🏆 This is a huge win for Easy Auth config (and beyond).

Issue #483, opened 2021-03-09
Level 1 APIs, file formats, specs


cjhskippy:
Provide outline of the information that a State agency tech team would need to implement a basic integration.

Document:
- [ ] API specs for state agencies to implement bulk uploads, Orchestrator API, Lookup ID API

PR #525, opened 2021-03-17
rough start to state quick start guide API section

echappen:
Addresses #483


echappen:
@ryanhofdotgov I figured this would remain WIP until we finalize more elements of the API, but wanted to externalize the points we talked about the other day:
- the NAC API is a single entity (even though they touch different subsystems)
- This top-level guide would reference the more detailed documentation for each subsystem as the Single Source of Truth
- Authentication to all endpoints is token-based, any dev would need to contact us for access

Let me know if this is going in the right direction.

ryanhofdotgov:
@echappen Definitely the right direction 👍 

I'm anticipating there will be some of those workflow diagrams from @juliaklindpaintner's deck on this page, or perhaps put them on the main README and just reference them here.

juliaklindpaintner:
@echappen I edited the diagrams to correspond more directly to the process described here — they're the first few slides in [this deck](https://docs.google.com/presentation/d/1MwRuw6EcLkcpHuEwGXgZ9_1RRAFO7O9SeCdCsRbZA6k/edit#slide=id.gc9a83cae3a_0_101). I didn't have one that represented the lookup workflow, so I added that. We can work together to refine them and then save them out for this or any other page in the repo. Feel free to leave comments if you see things that should be different!

echappen:
@juliaklindpaintner Thanks! One thing to note:

When starting to write API specs for for Match ID Lookup, the engineering team realized the name "Match ID" was confusing us in a number of ways. So we switched the name to "Lookup ID" in the API spec to differentiate it more from the active matching process. If this is an issue with product/design, let's chat!

juliaklindpaintner:
@echappen I don't see why that would be an issue! Would it make sense to rename the associated API endpoint accordingly?

docs/quick-start-guide-states.md
```diff
@@ -0,0 +1,50 @@
...
+| production | actual data used in the production system | [coming soon] |
+
+### Endpoints Overview
+
+| endpoint | description | Type | Parameters | Response | Authentication |
+|---|---|---|---|---|---|
+| `/BulkUpload` | uploads bulk participant data to the system | POST | a CSV file [in this format](../etl/docs/bulk-import.md) | [coming soon] | Key-based, contact a developer for access |
+| [coming soon] | query the system for status on data processing from a bulk upload | POST | [coming soon] | [coming soon] | Key-based, contact a developer for access |
+| `/query` | query for matches | POST | refer to the [OpenApi Schema](../match/docs/openapi/orchestrator/index.yaml) | refer to the [OpenApi Schema](../match/docs/openapi/orchestrator/index.yaml) | Key-based, contact a developer for access |
+| `matchi_ids/:id` | Match ID Lookup | GET | match ID | [coming soon] | Key-based, contact a developer for access |
```
juliaklindpaintner: Maybe this should be renamed Lookup ID here? It looks like this endpoint also may have a typo — an extra i after match?

echappen: You may be looking at an outdated version of the file. I updated them [here](https://github.com/18F/piipan/blob/7b7c147c26c1a589edbc592b1300155a085f9a37/docs/quick-start-guide-states.md).

juliaklindpaintner:
Whoops, sure was! Sorry about that!

__


Julia Lindpaintner (she/her)

18F <https://18f.gsa.gov/> | Technology Transformation Service
<http://www.gsa.gov/portal/category/25729>

General Services Administration (GSA)
<http://www.gsa.gov/portal/category/100000>

312.813.7727 | ***@***.***

Follow 18F on Twitter <https://twitter.com/18F>

docs/quick-start-guide-states.md
juliaklindpaintner: Whoops, sure was! I'm sorry about that!

echappen:
@ryanhofdotgov I marked this as ready for review, I propose we merge and add docs as they're created

ryanhofdotgov (approved):
OK - sounds good. Before we send out to states, I think in a follow up issue(s) we'd want to:
* ditch the pre-prod/prod rows – we need to check with our partner before publishing here
* Consider how we present the match, lookup id, status APIs a single unified API. The fact they are 3 separate Function Apps is an implementation detail that shouldn't drive the documentation I think.
* Have a nice OpenAPI generated set of HTML documentation to link to.

ryanhofdotgov:
Any other things we should specifically address in a follow up?

echappen:
That list looks good to me. Added them to #623.

PR #526, opened 2021-03-17
Separate Easy Auth config from main IaC

ryanhofdotgov:
This will allow us to continue to run the main IaC in environments where we only have the Contributor role and have a subscription administrator separately run the Easy Auth configuration.

Partially addresses #101. Most of the code changes displayed in the diff are just function code blocks moving from one file to another.


kingcomma (approved):
👍 `get_resources` is such a useful tool in this case!

One tiny nit, looks like the `orch_identity` variable is no longer used in `create-resources` ( ([line 403](https://github.com/18F/piipan/blob/ad0fe1ac87639c5fdce96314b980ab4e9ae78eee/iac/create-resources.bash#L403)) and could be yanked out.

ryanhofdotgov:
> One tiny nit, looks like the `orch_identity` variable is no longer used in `create-resources` 

Good catch! Removed in c09e92a.

PR #527, opened 2021-03-17
Don't let local silently absorb non-zero exit code

ryanhofdotgov:
Per bash docs, `local myvar=$(subcommand)` will always return 0 so don't use it inline with subcommands. Also: explicitly send error messages to `stderr`.


kingcomma (approved):
Sorry, overlooked this one. LGTM.

Issue #28, opened 2020-10-27
Orchestrator API is only accessible by authorized computer systems


cjhskippy:
As a system administrator, I want the Orchestrator API to only be accessible by authorized computer systems so that enrollee's PII is not compromised.

Implementation notes:
- Use [Azure API Management](https://azure.microsoft.com/en-us/services/api-management/)

AC:
- [x] Attempts to access the Orchestrator API without valid credentials return a 401 Access Denied error message
- [x] Limit access with credentials that are unique to each state. Azure administrator can:
  - [x] Authorize a new system to integrate with the Orchestrator API
  - [x] Revoke access of a system that integrates with the Orchestrator API
  - [x] Assign new credentials to a system that already has access to use the Orchestrator API
- [x] Each NAC internal application (ex. query tool) has its own credential (note: not using the same mechanism as the external-facing API)
- [x] Credential rotation mechanism is compatible with NIST-compliant process
- [x] Documented (see #556)
- [ ] ~~Limit access by network~~ (may not be desirable)
- [ ] ~~Accessed through Web Application Firewall~~ (not relevant)

_Note for context: in EDRS, access is limited to IP address (each state specifies which IP ranges to allow).  Also has a password that is updated every 60 days_

PR #530, opened 2021-03-18
Create APIM stub for public API

kingcomma:
First step in creating the API Management (APIM) instance to sit in front of the orchestrator API. Specifically:
- Creates APIM instance from ARM template
- Cleans unused default resources created in the process

This does not create a function API or any publicly accessible resources (other than a "portal" error page at `https://<apim-name>.developer.azure-api.net/`) that seems to be unavoidable at this juncture). This a first step in a series of self-contained PRs for this issue.

I'd especially appreciate feedback on:
- Creating a new APIM instance can take upwards of ~45-50 minutes. Consequently, I've created a `create-apim` script that is intended to be run separately from `create-resources` so developers aren't subjected to that delay on every IaC build.
- APIM instances require a "published email" property. I've externalized this as an argument passed to `create-apim` so as not to include any email addresses in the repo. An alternative would be to use a junk email.
- The name of the APIM instance becomes part of the public-facing URL. Since APIM names are globally unique, I'm appending a standard unique string to the end of the name. The result is a URL like `https://<apim-name>-<unique-string>.azure-api.net/`. Calling this out since (1) the URL is now dependent on the resource group name and (2) we'll need this URL to be stable if/when first round states use the sandbox for testing.

### DIFF
--- a/iac/arm-templates/apim.json
+++ b/iac/arm-templates/apim.json
@@ -1,0 +1,90 @@
+{
+    "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#",
+    "contentVersion": "1.0.0.0",
+    "parameters": {
+        "apiName": {
+            "type": "String"
+        },
+        "resourceTags": {
+            "type": "object"
+        },
+        "location": {
+            "type": "String"
+        },
+        "publisherEmail": {
+            "type": "String"
+        },
+        "publisherName": {
+            "type": "String"
+        },
+        "functionAppName": {
+            "type": "String"
+        }
+    },
+    "variables": {
+        "uniqueApiName": "[concat(toLower(parameters('apiName')), '-', uniqueString(resourceGroup().id))]",
+        "systemTypeTag": {
+            "SysType": "NacApi"
+        }
+    },
+    "resources": [
+        {
+            "type": "Microsoft.ApiManagement/service",
+            "apiVersion": "2020-06-01-preview",
+            "name": "[variables('uniqueApiName')]",
+            "location": "[parameters('location')]",
+            "tags": "[union(parameters('resourceTags'), variables('systemTypeTag'))]",
+            "sku": {
+                "name": "Developer",
+                "capacity": 1
+            },
+            "identity": {
+                "type": "SystemAssigned"
+            },
+            "properties": {
+                "publisherEmail": "[parameters('publisherEmail')]",
+                "publisherName": "[parameters('publisherName')]",
+                "hostnameConfigurations": [
+                    {
+                        "type": "Proxy",
+                        "hostName": "[concat(variables('uniqueApiName'), '.azure-api.net')]",
+                        "defaultSslBinding": true
+                    }
+                ],
+                /* Explicity asserting security settings rather than relying on defaults. */
+                "customProperties": {
+                    "Microsoft.WindowsAzure.ApiManagement.Gateway.Security.Ciphers.TripleDes168": "false",
+                    "Microsoft.WindowsAzure.ApiManagement.Gateway.Security.Protocols.Tls11": "false",
+                    "Microsoft.WindowsAzure.ApiManagement.Gateway.Security.Protocols.Tls10": "false",
+                    "Microsoft.WindowsAzure.ApiManagement.Gateway.Security.Protocols.Ssl30": "false",
+                    "Microsoft.WindowsAzure.ApiManagement.Gateway.Security.Backend.Protocols.Tls11": "false",
+                    "Microsoft.WindowsAzure.ApiManagement.Gateway.Security.Backend.Protocols.Tls10": "false",
+                    "Microsoft.WindowsAzure.ApiManagement.Gateway.Security.Backend.Protocols.Ssl30": "false",
+                    "Microsoft.WindowsAzure.ApiManagement.Gateway.Protocols.Server.Http2": "false"
+                }
+            }
+        },
+        {
+            "type": "Microsoft.ApiManagement/service/apis",
+            "apiVersion": "2020-06-01-preview",
+            "name": "[concat(variables('uniqueApiName'), '/', parameters('functionAppName'))]",
+            "dependsOn": [
+                "[resourceId('Microsoft.ApiManagement/service', variables('uniqueApiName'))]"
+            ],
+            "properties": {
+                "displayName": "Orchestrator API",
+                "subscriptionRequired": true,
+                "protocols": [
+                    "https"
+                ],
+                "path": "/api"
+            }
+        }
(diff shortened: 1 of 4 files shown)


iac/arm-templates/apim.json
```diff
@@ -0,0 +1,90 @@
...
+        },
+        "publisherName": {
+            "type": "String"
+        },
+        "functionAppName": {
+            "type": "String"
+        }
+    },
+    "variables": {
+        "uniqueApiName": "[concat(toLower(parameters('apiName')), '-', uniqueString(resourceGroup().id))]",
```
echappen, line 25: You might have missed this from last week, but Ryan and I started incorporating a final [naming convention](https://github.com/18F/piipan/blob/dfa37755de66f4b06460ca2a62644a34e0125231/docs/azure-naming-conventions.md) for resources into our iac scripts, which is based on Azure's [best practices](https://docs.microsoft.com/en-us/azure/cloud-adoption-framework/ready/azure-best-practices/resource-naming#example-names-storage).

Using this convention, an example name for this resource in a tts/dev environment may be `apim-nacapi-dev` and `fns-apim-nacapi-prod` for fns/prod. 

I don't think this is blocking for now but wanted to make you aware of it.

iac/arm-templates/apim.json
```diff
@@ -0,0 +1,90 @@
...
+        },
+        {
+            "type": "Microsoft.ApiManagement/service/apis",
+            "apiVersion": "2020-06-01-preview",
+            "name": "[concat(variables('uniqueApiName'), '/', parameters('functionAppName'))]",
+            "dependsOn": [
+                "[resourceId('Microsoft.ApiManagement/service', variables('uniqueApiName'))]"
+            ],
+            "properties": {
+                "displayName": "Orchestrator API",
```
echappen, line 75: [question] Do you want this displayName hard-coded?

iac/arm-templates/apim.json
```diff
@@ -0,0 +1,90 @@
...
+                "subscriptionRequired": true,
+                "protocols": [
+                    "https"
+                ],
+                "path": "/api"
```
echappen, line 80: [question] Would api versioning come into play for the path here? (Or anywhere else in the template?)

iac/create-apim.bash
```diff
@@ -0,0 +1,81 @@
...
+  # Load agency/subscription/deployment-specific settings
+  azure_env=$1
+  source $(dirname "$0")/env/${azure_env}.bash
+  verify_cloud
+
+  APIM_NAME='nac-api'
+  PUBLISHER_NAME='NAC Administrator'
+  publisher_email=$2
+  #orch_name=$(get_resources $ORCHESTRATOR_API_TAG $MATCH_RESOURCE_GROUP)
+  orch_name='ofunchaaz7alkqwuvo'
```
echappen, line 59: [question] Did you want this hard-coded?

echappen (commented):
I like the idea of having independent iac scripts that deploy more targeted resources (I always seem to create smaller scripts anyway for testing and triage).

I had some clarifying questions and didn't know if you wanted some stuff in here hard-coded. If I'm mistaken then 👍 from me!

iac/arm-templates/apim.json
kingcomma: This is a helpful call out, thanks! Have we established the approach for determining the `-[environment]` suffix in an IaC script? Something along the lines of `PREFIX` in `env/tts/dev.bash`?

iac/arm-templates/apim.json
kingcomma: Yeah, I spent a minute going back and forth on this. At this stage I'm considering it the path of least resistance. It may make sense to pull this out to a CLI parameter or, perhaps better, just move it up to a hardcoded variable/parameter like `systemTypeTag`. I don't foresee it being used programmatically, but this is also the part of the template that will be most prone to iteration over the next few PRs.

iac/arm-templates/apim.json
kingcomma: Yep — API Management has a [versioning feature](https://docs.microsoft.com/en-us/azure/api-management/api-management-versions#versions-and-revisions) that will be implemented via an additional [apiVersionSets](https://docs.microsoft.com/en-us/azure/templates/microsoft.apimanagement/service/apiversionsets?tabs=json) resource in this template. I'm deferring implementation for now to keep this PR paired down.

iac/create-apim.bash
kingcomma: Good catch! This was a temporary placeholder until I could rebase off of #526 and make use of `get_resources`. Fixed with 915a905.

iac/arm-templates/apim.json
echappen: Yep it'd be something like [this](https://github.com/18F/piipan/blob/main/iac/env/tts/dev.bash#L17): `ENV=$(basename "${BASH_SOURCE%.*}")`

echappen (approved):

iac/iac-common.bash
```diff
@@ -8,6 +8,7 @@ PER_STATE_MATCH_API_TAG="SysType=PerStateMatchApi"
 ORCHESTRATOR_API_TAG="SysType=OrchestratorApi"
 DASHBOARD_APP_TAG="SysType=DashboardApp"
 QUERY_APP_TAG="SysType=QueryApp"
+NAC_API_TAG="SysType=NacApi"
```
ryanhofdotgov, line 11: Wondering if we can continue to describe things in a NAC/Piipan agnostic way? e.g., when I deploy it to partner, it doesn't say Piipan, and when I work on it as a reference model, it doesn't reference NAC?

kingcomma: Makes sense. A few candidates that come to mind:
- "match" (e.g. `SysType=MatchApi`, `apim-matchapi-dev`)
- "public" (e.g. `SysType=PublicApi`, `apim-publicapi-dev`)
- "system" (e.g. `SysType=SystemApi`, `apim-systemapi-dev`)

"Match" is probably the right fit for the current story, but public/system are a bit more flexible if the APIM instance is extended to include other APIs (upload, collaboration, etc).

kingcomma: @ryanhofdotgov I generalized the naming in 276b880 using "public". Happy to discuss this further — just wanted to get something into the PR.

ryanhofdotgov: For sure! And doesn't hold up the PR at all.

ryanhofdotgov: (but see that NAC_API_TAG is now dangling)

ryanhofdotgov (approved):

kingcomma:
Partially addresses #28.

PR #531, opened 2021-03-18
Adapt Easy Auth IaC for Azure Government

ryanhofdotgov:
- Use AzureUSGovernment-specific graph API URL
- Add pauses between resource creation and use in a few more places
- Generalize to reduce code duplication

Partially addresses #101


kingcomma (approved):
👍

Issue #428, opened 2021-02-25
Deploy the walking skeleton in FNS


cjhskippy:
Deploy to FNS Azure and verify that everything is in there.

- All services built to date are running in FNS environment
- Run sprint 9 demo off of code running in FNS Azure

PR #536, opened 2021-03-19
Adapt PerStateMatchApi for AzureUSGovernment

ryanhofdotgov:
Use same approach as in ETL BulkUpload. Code is essentially duplicated from PR #500; defer refactor to #192.


ryanhofdotgov:
Partially addresses #428

kingcomma (approved):
👍

Issue #409, opened 2021-02-24
Convert Metrics Database authentication to managed identity


echappen:
In order to keep our database resources more consistent, we should convert the metrics database to use a managed identity instead of the current admin username and password.

kingcomma:
Steps involved:

- [ ] Metrics Collector
  - [ ] Activate system-assigned identity in IaC
  - [ ] Add database role for identity in IaC
  - [ ] Update application to use identity when connecting to database
- [ ] Metrics API
  - [ ] Activate system-assigned identity in IaC
  - [ ] Add database role for identity in IaC
  - [ ] Update application to use identity when connecting to database
- [ ] Remove Key Vault permissions from IaC for both Functions

kingcomma:
@ryanhofdotgov This work consists of creating two roles in the the `core` database: one for the Metrics Collector's system-assigned identity and one for the Metrics API system-assigned identity. That's about where I hit the limits of my knowedge on roles and privileges in PostgreSQL. Do we want to follow a similar approach to privileges as we do in [`create-database.bash`](https://github.com/18F/piipan/blob/dev/iac/create-databases.bash#L40)?

ryanhofdotgov:
Similar in the broad strokes, but perhaps a little bit more granular? Perhaps:
 - a database `metrics` owned by `metrics` role (or whatever it is called already)
 - grant Metrics API role just read access (https://stackoverflow.com/a/762649, `GRANT...ALL TABLES IN SCHEMA...`)
 - grant collector role read/write, without admin access (e.g., SELECT, INSERT, UPDATE, DELETE, and TRUNCATE)

There may be some differences in how the metrics uses schemas vs the participants records (e.g., `public` vs custom one)

PR #537, opened 2021-03-19
Adapt metrics subsystem for AzureUSGovernment

ryanhofdotgov:
Adopt pattern implemented in ETL, per-state match API for Key Vault. Key Vault usage will be eliminated in #409, so not concerned about the duplicated code across `Api` and `Collect`.


ryanhofdotgov:
Partially addresses #428

echappen (approved):

Issue #157, opened 2020-12-15
Solution'ize our .NET projects


ryanhofdotgov:
As we start to incorporate CI/CD processes, configure specific build settings (e.g., package.lock.json), and grow the number of subsystems in our monorepo, we should start organizing our .NET projects under a per-subsystem solution (i.e., .sln file).

I believe this will allow us to reference/configure the .sln file for a subsytem rather than having to do it for every separately namespaced project/test suite.

Reference: https://docs.microsoft.com/en-us/visualstudio/ide/solutions-and-projects-in-visual-studio?view=vs-2019

PR #539, opened 2021-03-19
Add solution files for each subsystem

echappen:
Addresses #157 

- 1 sln per subsystem
- subsystem sln includes test projects

### DIFF
--- a/dashboard/dashboard.sln
+++ b/dashboard/dashboard.sln
@@ -1,0 +1,56 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio 15
+VisualStudioVersion = 15.0.26124.0
+MinimumVisualStudioVersion = 15.0.26124.0
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "src", "src", "{C7B8CB0F-6271-42ED-B4B6-6A9FAE30D5F9}"
+EndProject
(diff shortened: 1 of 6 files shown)


ryanhofdotgov (approved):
Would it be possible to add a quick cheat sheet to creating them from an existing .csproj, how it affects builds, or testing, dependency updates, etc.? 

Michael and I believed solution files would make our lives better, but admittedly now what we have these, I don't know what to do with them 😂  

In particular, I see that the docs are Visual Studio heavy vs Visual Studio Code.

ryanhofdotgov:
Also, doesn't need to hold up this PR, but didn't see a .sln for the ETL code.

echappen:
I... don't really know what to do with them either 🤣 but I'll keep researching!

echappen:
After some more research and [discussion with other TTS folks](https://gsa-tts.slack.com/archives/C02CD5VUQ/p1616509321008100), it seems the main benefit has to do with having more strategic and faster builds as your projects grow. 

I still haven't found much detail on how exactly .NET achieves this, or how exactly solutions are useful outside of Visual Studio. All the official Microsoft docs are Visual Studio-specific. The closest I could find was [this](https://garywoodfine.com/creating-editing-solution-files-dotnet-core-ubuntu/):

> Although not strictly necessary from a pure development perspective, we don’t necessarily need solution files when developing with VS Code, but solution files do come in handy during the Build and Release phases of projects, so it is generally good practice to start thinking about using them in the early stages of your project development.

For now, even though we're working in VSCode, it's still sort of nice to only run `dotnet build` once at the `.sln` level, instead of having to build every project at the `.csproj` level.

ryanhofdotgov:
@echappen Cool – let's incorporate as a part of our workflow and see how it goes.

Issue #528, opened 2021-03-18
Require first name for NAC queries


cjhskippy:
As a state agency worker, I want the NAC to accept any valid legal names so that I can check the eligibility of every applicant.

The NAC currently leaves the first name field optional, but according to guidance from https://www.ssa.gov/employer/ssnvshandbk/usingSSNVS.htm , we can expect all names to include a first name and a last name.

Subtask of #411

AC:
New Behaviors:
- [x] Query API requires first name
- [x] Query website requires first name

Existing behaviors to keep:
- [x] API + Query website still require last name
- [x] API + Query website still accept optional name fields (middle name)

PR #540, opened 2021-03-19
Require first name in API/Query tool; Don't require middle name

heymatthenry:
Addresses #528 

* Add validations for first name in Match & Orchestrator APIs
* Remove validation for middle name in Query Tool

### DIFF
--- a/match/src/Piipan.Match.Orchestrator/MatchRequest.cs
+++ b/match/src/Piipan.Match.Orchestrator/MatchRequest.cs
@@ -20,7 +20,6 @@ namespace Piipan.Match.Orchestrator
         public string Last { get; set; }
 
         [JsonProperty("first", Required = Required.Always)]
-        [JsonConverter(typeof(NullConverter))]
         public string First { get; set; }
 
         [JsonProperty("middle", NullValueHandling = NullValueHandling.Ignore)]
--- a/match/src/Piipan.Match.Orchestrator/MatchRequest.cs
+++ b/match/src/Piipan.Match.Orchestrator/MatchRequest.cs
@@ -19,7 +19,7 @@ namespace Piipan.Match.Orchestrator
         [JsonProperty("last", Required = Required.Always)]
         public string Last { get; set; }
 
-        [JsonProperty("first", NullValueHandling = NullValueHandling.Ignore)]
+        [JsonProperty("first", Required = Required.Always)]
         [JsonConverter(typeof(NullConverter))]
         public string First { get; set; }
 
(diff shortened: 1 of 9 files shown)


match/src/Piipan.Match.Orchestrator/MatchRequest.cs
```diff
@@ -19,7 +19,7 @@ public class MatchQuery
         [JsonProperty("last", Required = Required.Always)]
         public string Last { get; set; }
 
-        [JsonProperty("first", NullValueHandling = NullValueHandling.Ignore)]
+        [JsonProperty("first", Required = Required.Always)]
         [JsonConverter(typeof(NullConverter))]
```
kingcomma, line 23: *Nitpick:* Should be able to drop this `NullConverter` now that the property is required in all incoming requests.

match/src/Piipan.Match.State/MatchRequest.cs
```diff
@@ -19,7 +19,7 @@ public class MatchQuery
         [JsonProperty("last", Required = Required.Always)]
         public string Last { get; set; }
 
-        [JsonProperty("first", NullValueHandling = NullValueHandling.Ignore)]
+        [JsonProperty("first", Required = Required.Always)]
         [JsonConverter(typeof(NullConverter))]
```
kingcomma, line 23: *Nitpick:* Should be able to drop this converter as well.

kingcomma (commented):
Aside from my nitpicks, I think the SQL that retrieves participant data (see [`Piipan.Match.State.Api.Prepare`](https://github.com/18F/piipan/blob/mh-update-validations/match/src/Piipan.Match.State/Api.cs#L91)) needs a couple adjustments:
- Add an explicit equality condition for first name (eg., `first=@first`) and remove the conditional handling for the now non-existent case when it is `null`
- Consider if the `middle` field should be part of the query at all

I'll let @cjhskippy weigh in on whether or not that second item is out of scope for this story. But my understanding of the desired functionality is that middle names should be accepted but not factored into match determination. E.g.:
- "Foo Bar Doe" should match all of: "Foo Bar Doe", "Foo Doe", and "Foo Baz Doe"
- "Foo Bar" should match all of: "Foo Bar", "Foo Baz Bar", and "Foo * Bar"

As it currently stands, `middle` is an optional field but the exact match algorithm still matches against it (checking equality with `null` in cases when it is not included).

match/src/Piipan.Match.Orchestrator/MatchRequest.cs
heymatthenry: Good nitpick! Don't think I even noticed the `NullConverter`s

match/src/Piipan.Match.State/MatchRequest.cs
heymatthenry: ✅

heymatthenry:
@kingcomma Per @cjhskippy's comment in Slack, I just removed the null case for `first` and will handle the middle name in another ticket.

kingcomma (approved):
👍 Looks good to me!

PR #547, opened 2021-03-22
Fix: move GSA IP restrictions from web app level to Front Door/WAF policy level

echappen:
Initially, restricting access to only GSA IP's was set on the web apps themselves.

When Azure Front Doors were applied to the apps, this blocked traffic from the Front Door Backends.

Instead, the web apps are now restricted to `AzureFrontDoor.Backend` traffic, and GSA IP restrictions are now set as rules on the related Front Door WAF policies. 

This restricts traffic to the app from everywhere except through the front-door URL when on the GSA network.

## Considerations
This creates a co-dependency when creating web app and front door resources: web apps need the front door ID in order to set the backend restrictions, and front doors need the app host name in order to connect the app to the front door resource.

Consequently, the ARM templates become more reliant on naming conventions. It also becomes more difficult to name resources dynamically through the templates, as names are often needed for other resources before the main resource is deployed.

### DIFF
--- a/iac/arm-templates/dashboard-app.json
+++ b/iac/arm-templates/dashboard-app.json
@@ -5,6 +5,9 @@
         "appName": {
             "type": "string"
         },
+        "frontDoorId": {
+            "type": "string"
+        },
         "location": {
             "type": "string"
         },
@@ -19,16 +22,24 @@
         }
     },
     "variables": {
-        "appName": "[concat(parameters('appName'), '-', uniqueString(resourceGroup().id))]",
+        "appName": "[concat(parameters('appName'))]",
         "sku": "S1",
         "baseSiteConfig": {
             "linuxFxVersion": "DOTNETCORE|3.1",
             "ipSecurityRestrictions": [
+                /* Restricts access to Front Door resource only */
+                /* https://docs.microsoft.com/en-us/azure/app-service/app-service-ip-restrictions#manage-access-restriction-rules-programmatically */
                 {
-                    "ipAddress": "159.142.0.0/16",
+                    "ipAddress": "AzureFrontDoor.Backend",
+                    "tag": "ServiceTag",
                     "action": "Allow",
-                    "name": "GSA IP range",
-                    "priority": 100
+                    "priority": 100,
+                    "name": "Allow Azure Front Door access",
+                    "headers": {
+                        "x-azure-fdid": [
+                            "[concat(parameters('frontDoorId'))]"
+                        ]
+                    }
                 }
             ],
             "ftpsState": "Disabled",
--- a/iac/arm-templates/front-door-app-service.json
+++ b/iac/arm-templates/front-door-app-service.json
@@ -66,6 +66,52 @@
                     "mode": "[parameters('wafMode')]",
                     "enabledState": "Enabled"
                 },
+                "customRules": {
+                    "rules": [
+                        /* Allow GSA IP's */
+                        {
+                            "name": "IPAllowListRule",
+                            "enabledState": "Enabled",
+                            "priority": 300, /* Lower number denotes higher priority. This Allow Rule must be a highger priority than the Deny rule below. */
+                            "ruleType": "MatchRule",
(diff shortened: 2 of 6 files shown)


iac/arm-templates/front-door-app-service.json
```diff
@@ -66,6 +66,52 @@
...
+                                    "negateCondition": false,
+                                    "matchValue": [
+                                        "159.142.0.0/16" /* GSA IP range */
+                                    ],
+                                    "transforms": []
+                                }
+                            ],
+                            "action": "Allow"
+                        },
+                        /* Deny any non-GSA IP's */
```
ryanhofdotgov, line 92: Is there possibly a way to simplify this config? Typically for a firewall you'd set a lower priority rule that blocks everything, then a higher priority rule to allow what you need. Would eliminate the double negative rule.

iac/arm-templates/dashboard-app.json
```diff
@@ -19,16 +22,24 @@
...
-                    "ipAddress": "159.142.0.0/16",
+                    "ipAddress": "AzureFrontDoor.Backend",
+                    "tag": "ServiceTag",
                     "action": "Allow",
-                    "name": "GSA IP range",
-                    "priority": 100
+                    "priority": 100,
+                    "name": "Allow Azure Front Door access",
+                    "headers": {
+                        "x-azure-fdid": [
```
ryanhofdotgov, line 39: You mentioned earlier we needed to only allow the Front Door IP range. Where does that config live?

ryanhofdotgov:
Regarding co-dependencies... if it makes it easier/more maintainable, we can probably split up front door config from ARM template. Do bulk of app in template, then assign front door using AZ CLI.

We should plan on naming resources through templates (ditching the all those resource group hashes) next sprint, now that we've successfully deployed everything to Azure Government.

iac/arm-templates/dashboard-app.json
echappen: So the instructions I [linked](https://github.com/18F/piipan/blob/7b7c147c26c1a589edbc592b1300155a085f9a37/docs/quick-start-guide-states.md) to earlier were unclear. You can either list all of those IP's _or_ you can just use a service type (in our case `AzureFrontDoor.Backend`) which accounts for all those IP's.

echappen: In either case, you have to limit to only your Front Door instance, which is what the `x-azure-fdid` does.

iac/arm-templates/front-door-app-service.json
echappen: I can keep playing around with it

iac/arm-templates/dashboard-app.json
ryanhofdotgov: The link in your comment above is pointing to the state quick start –  Is that intentional?

Do you have a reference for the service type accounting for all the IPs?

echappen: Whoops, correct link to docs I was referencing earlier: https://docs.microsoft.com/en-us/azure/frontdoor/front-door-faq#how-do-i-lock-down-the-access-to-my-backend-to-only-azure-front-door

ryanhofdotgov: Thanks for the quick sync to explain that `AzureFrontDoor.Backend` maps to that IP range. Here's the other link that shows step-by-step on using `AzureFrontDoor.Backend` in the portal: https://techcommunity.microsoft.com/t5/azure-architecture-blog/permit-access-only-from-azure-front-door-to-azure-app-service-as/ba-p/2000173

ryanhofdotgov:
thought: it will be interesting at some point to see how ARM template dependencies work when the resources aren't established in the same ARM template file. For example, the WAF template would have an explicit dependency on the App Service.

ryanhofdotgov (approved):
👍 pending a successful IaC run

echappen:
True, I could see as the WAF rules get more complex, it may warrant its own template, even on a per-app basis.

Issue #548, opened 2021-03-22
Explore importing OpenAPI specs when creating APIM instances


kingcomma:
OpenAPI specs can be imported as a `value` property of an APIM instance's [API resource](https://docs.microsoft.com/en-us/azure/templates/microsoft.apimanagement/2019-01-01/service/apis?tabs=json#apicreateorupdateproperties-object). This automates the creation of operations, request/response representations, and other resources. Without importing the spec, these resources must be hardcoded in an ARM template.

Achieving this functionality requires a compiled OpenAPI spec at the time the IaC is run. This may mean reevaluating the decision to decompose the API specs into multiple files and the implied `swagger-cli` dependency.

PR #549, opened 2021-03-22
Add POST operation for matching API query

kingcomma:
Adds initial operation to APIM instance and creates a `/query` endpoint.

Follow up PR will connect this operation to the backend server (orchestrator Function App). Until then, the endpoint is effectively a stub.

The operation is hard-coded in `apim.json` for now, but could be scraped in favor of dynamic creation by importing the OpenAPI spec via the [`value` property on the API resource](https://docs.microsoft.com/en-us/azure/templates/microsoft.apimanagement/2019-01-01/service/apis?tabs=json#apicreateorupdateproperties-object). This work is described in #548.


echappen (approved):

kingcomma:
Partially addresses #28

PR #553, opened 2021-03-22
Point APIM to backend service

kingcomma:
- Reference orchestrator URL from APIM API resource
- Configure Easy Auth between APIM and orchestrator

Follow on work to #549. Puts the pieces in place to connect the APIM instance to the orchestrator API. Will be followed by a PR that connects the APIM frontend to the orchestrator backend using a managed identity authentication policy.

### DIFF
--- a/iac/arm-templates/apim.json
+++ b/iac/arm-templates/apim.json
@@ -16,6 +16,9 @@
         },
         "publisherName": {
             "type": "String"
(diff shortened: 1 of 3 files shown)


ryanhofdotgov (approved):
👍

kingcomma:
Partially addresses #28

PR #555, opened 2021-03-22
Add APIM policy for Public API

kingcomma:
- Add base XML policy, extended to include managed identity auth
- Update policy with application ID URI when running IaC
- Drop unnecessary subscription key header when forwarding request to
  orchestrator API

This is the final piece in creating a working APIM instance for the public API. After the IaC is run, any request with a valid [subscription key](https://docs.microsoft.com/en-us/azure/api-management/api-management-subscriptions) will initiate an API call to the orchestrator API. Specifically:
1. A user/system submits a valid request to the APIM endpoint
2. APIM authenticates with orchestrator API using its system-assigned identity
3. APIM forwards the request to the orchestrator API
4. APIM forwards response from orchestrator API back to user/system

The XML policy is barebones, and the current ["templating" solution using `sed`](https://github.com/18F/piipan/blob/a021622f1bee701a40e3d47bb2ef7ff7e385f838/iac/create-apim.bash#L70-L73) may not scale if it becomes more complex.

Addresses #28

### DIFF
--- a/iac/apim-policy.xml
+++ b/iac/apim-policy.xml
@@ -1,0 +1,18 @@
+<!-- Generated via IaC -->
+
+<policies>
+    <inbound>
+        <base />
+        <authentication-managed-identity resource="{applicationUri}" />
+        <set-header name="Ocp-Apim-Subscription-Key" exists-action="delete" />
+    </inbound>
+    <backend>
+        <base />
+    </backend>
+    <outbound>
+        <base />
+    </outbound>
+    <on-error>
+        <base />
+    </on-error>
+</policies>
--- a/iac/apim-policy.xml
+++ b/iac/apim-policy.xml
@@ -4,6 +4,7 @@
     <inbound>
         <base />
         <authentication-managed-identity resource="{applicationUri}" />
+        <!-- APIM has used this header to authenticate the request, drop it as we forward the request to our internal endpoint, as it has no use for the client's private API key -->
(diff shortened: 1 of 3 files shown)


iac/apim-policy.xml
```diff
@@ -0,0 +1,18 @@
+<!-- Generated via IaC -->
+
+<policies>
+    <inbound>
+        <base />
+        <authentication-managed-identity resource="{applicationUri}" />
+        <set-header name="Ocp-Apim-Subscription-Key" exists-action="delete" />
```
ryanhofdotgov, line 7:
```suggestion
        <!-- APIM has used this header to authenticate the request, drop it as we forward the request to our internal endpoint, as it has no use for the client's private API key -->
        <set-header name="Ocp-Apim-Subscription-Key" exists-action="delete" />
```
Just to have the purpose of the policy more explicit.

ryanhofdotgov (approved):
Nice 👍 

As follow on work, I think we need to:
* explicitly document `sed` as a IaC dependency. It's part of Git for Windows and also included in Cloud Shell. I'll open a PR for this.
* open up an issue to describe the concept of operations for subscriptions (e.g., pointer on how to set them up, how to call the API, etc.). Can you write up the issue/description for this?

thought: I think we are definitely at past the point were we need integration tests that we can at least manually run (if not automate) for our internal and external APIs. And make them part of our definition of done in the future. Something as simple as a wrapper for the curl command you shared with me. That's what I've been using for this PR and as I've tested the components up in Azure Government.

PR #570, opened 2021-03-23
switch title of dashboard from QueryTool to Dashboard

echappen:
Just noticed this title looks sort of confusing when you're presenting:

Screen Shot 2021-03-23 at 11 14 30 AM


heymatthenry (approved):
D'oh! Copy paste error on my part. Good catch!

PR #574, opened 2021-03-23
include lookup api when generating api specs

echappen:
Just noticed I didn't include the new lookup api here


kingcomma (approved):
Tiny nit (non-blocking) that I'm realizing now when viewing the generated spec: the "Match" tag is still present in the spec but not used anywhere. When viewing the spec in the Swagger UI, this results in an empty "Match" section and a "default" section with the `lookup_ids` path:

Screen Shot 2021-03-23 at 8 55 06 PM

echappen:
Good catch, I removed the tag from the lookup spec

PR #579, opened 2021-03-23
Deny all IP addresses as lower-priority firewall rule

echappen:
For simplicity purposes -- this effectively does the same thing as the previous rule to deny all non-GSA IP's


echappen:
I re-ran the iac script in tts/dev with this new config and tested it—all good 👍

ryanhofdotgov (approved):
Cool – thanks!

Issue #576, opened 2021-03-23
Research spike: existing Azure storage acct tools


cjhskippy:
As a first-round state, I want to know what existing tools will make it faster to upload fake PII to the NAC so that I can integrate quickly.

- [ ] Research existence of native tools that you just download + run to use Azure storage account (may be wanted to support command-line execution)
- [ ] Research existing implementations of Azure storage account APIs in common programming languages (Java, .NET)
- [ ] Document findings

PR #583, opened 2021-03-24
Add documentation for how to upload csv files to Azure Blob storage

echappen:
Addresses #576

### DIFF
--- a/etl/docs/uploads.md
+++ b/etl/docs/uploads.md
@@ -10,7 +10,13 @@ AzCopy is a Microsoft-supported command line utility that you can use to upload 
 
 [Download AzCopy for your environment](https://docs.microsoft.com/en-us/azure/storage/common/storage-use-azcopy-v10#run-azcopy)
 
-[How to upload files using AzCopy](https://docs.microsoft.com/en-us/azure/storage/common/storage-use-azcopy-blobs-upload)
+Example command to upload a csv file to a storage account:
+
+```
+azcopy copy '<yourfile.csv>' '<provided-storage-url>'
+```
+
+For more details on using AzCopy, visit the [Microsoft documentation](https://docs.microsoft.com/en-us/azure/storage/common/storage-use-azcopy-blobs-upload).
 
 ### Programmatically
 
@@ -29,4 +35,4 @@ You can also upload files in a Put Blob request through Azure Blob Storage's [RE
 
 ## Access
 
-For all of these uploads methods, you'll need an Azure Access Key and a Connection string to the storage resource. Contact a project representative to gain these credentials.
+For all of these uploads methods, you'll need an Azure Access Key and a Connection string to the storage resource. Contact a project representative to gain these credentials, specifying your intended upload method.
--- a/etl/docs/uploads.md
+++ b/etl/docs/uploads.md
@@ -7,7 +7,7 @@ Once you have [validated the format](./bulk-import.md) of your participant data 
 
 AzCopy is a Microsoft-supported command line utility that you can use to upload files to Azure blob storage resources.
 
-[Download AzCopy for your environment](https://docs.microsoft.com/en-us/azure/storage/common/storage-use-azcopy-v10#run-azcopy)
+[Download AzCopy for your environment](https://docs.microsoft.com/en-us/azure/storage/common/storage-use-azcopy-v10#download-azcopy)
 
 #### Authorization
 
--- a/etl/docs/uploads.md
+++ b/etl/docs/uploads.md
@@ -1,6 +1,6 @@
 ## Uploading Participant Data
 
-> ⚠️ This documentation is under active development and subject to change.
+> ⚠️ The service API for the bulk upload of PII data is not stable enough for client development. This documentation describes a _temporary_ approach that should allow states to quickly provide test data to our system with out expending development effort unnecessarily.
 
 Once you have [validated the format](./bulk-import.md) of your participant data CSV, our recommended approach for uploading the file is through AzCopy.
 #### AzCopy
--- a/etl/docs/uploads.md
+++ b/etl/docs/uploads.md
@@ -22,7 +22,7 @@ All of these items will vary between test and production environments.
 
 Upon receiving these credentials, you'll first authorize AzCopy through a [service principal approach](https://docs.microsoft.com/en-us/azure/storage/common/storage-use-azcopy-authorize-azure-active-directory?toc=/azure/storage/blobs/toc.json#authorize-a-service-principal).
 
-First save the password:
+First, save the password to the `AZCOPY_SPA_CLIENT_SECRET` environment variable, taking care to avoid saving it in your command-line history:
 
 PowerShell: `$env:AZCOPY_SPA_CLIENT_SECRET="$(Read-Host -prompt "Enter key")"`
 
--- a/etl/docs/uploads.md
+++ b/etl/docs/uploads.md
@@ -9,27 +9,40 @@ AzCopy is a Microsoft-supported command line utility that you can use to upload 
 
 [Download AzCopy for your environment](https://docs.microsoft.com/en-us/azure/storage/common/storage-use-azcopy-v10#run-azcopy)
 
-Upon receiving credentials, you'll first authorize AzCopy through a [service principal approach](https://docs.microsoft.com/en-us/azure/storage/common/storage-use-azcopy-authorize-azure-active-directory?toc=/azure/storage/blobs/toc.json#authorize-a-service-principal).
+#### Authorization
 
-Example (Unix/Linux):
+Contact a project representative to gain the necessary credentials for uploading through AzCopy. What you will need:
+
+- application id
+- tenant id
+- password (this is used as the client secret)
+- storage account url
+
+All of these items will vary between test and production environments.
+
+Upon receiving these credentials, you'll first authorize AzCopy through a [service principal approach](https://docs.microsoft.com/en-us/azure/storage/common/storage-use-azcopy-authorize-azure-active-directory?toc=/azure/storage/blobs/toc.json#authorize-a-service-principal).
+
+First save the password:
+
+PowerShell: `$env:AZCOPY_SPA_CLIENT_SECRET="$(Read-Host -prompt "Enter key")"`
+
+Bash: `export AZCOPY_SPA_CLIENT_SECRET=<password>`
+
+Then login using the application id and tenant id:
 
 ```
-$ export AZCOPY_SPA_CLIENT_SECRET="$(Read-Host -prompt "Enter key")"
-$ azcopy login --service-principal  --application-id application-id --tenant-id=tenant-id
+azcopy login --service-principal  --application-id application-id --tenant-id=tenant-id
 ```
 
-Command will return a success or failure message.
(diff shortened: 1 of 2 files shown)


ryanhofdotgov:
thought: a new section in etl.md might be a better fit for this content as this doc was originally just scoped to the file format itself and doesn't really talk about uploading or processing

ryanhofdotgov:
Not sure where your research landed:
* do all of these support storage access keys, including AzCopy?
* do we need to provide Connection Strings, or is the access key and URL for the storage account enough? It seems like users of those libraries would need to know how to build a Connection String from the access key and URL?

echappen:
Here’s what I’ve found. I’ll start with naming the three types of auth because the terminology’s a bit confusing. 
 
Generally, you can [authorize with Azure blob storage](https://docs.microsoft.com/en-us/rest/api/storageservices/authorize-requests-to-azure-storage) in 3 ways:
- **Azure Active Directory** (AD). I think we’re ruling this out because it requires managing user identities through our own Azure account. 
- **Shared Keys** (called “Access Keys” in the portal). These are the two main keys provided by Azure for a storage account. Azure also provides a connection string for each key. Keys can be manually refreshed. 
- **Shared Access Signature** (SAS). These are manually created for a storage account. When you create a SAS “key” it generates a Connection String, a SAS Token, and a Blob Service SAS Url. Unlike Shared Keys, SAS provides finer grained access control, like settings for allowed permissions (read-only, write-only, etc) and expiration dates. 
 
Different tools for connecting to blob storage use one, all, or a subset of these types depending on the tool. 
 
About Shared Keys:
- AzCopy uses the Blob Service SAS Url. It doesn’t use connection strings for neither Shared Key nor SAS. AzCopy supported Shared Key connection strings at one point, but now using them is discouraged in favor of SAS. Current Azure documentation doesn’t list instructions for using Shared Keys, but there are [older docs](https://microsoft.github.io/AzureTipsAndTricks/blog/tip81.html) out there that demonstrate how to use them. AzCopy’s man pages doesn’t list their use either. This [response](https://github.com/Azure/azure-storage-azcopy/issues/186#issuecomment-460114558) from a maintainer of AzCopy indicates that you could still use them, but it’s discouraged. I’ve been unable to send a successful request through AzCopy using the old Shared Key instructions. 
- This isn’t said explicitly in the docs, but it appears Shared Keys are to be used for internal authorization between Azure resources as an alternative to Azure AD. (Shared Keys seem to pre-date AD which is now encouraged for this purpose). 
- Azure [docs](https://docs.microsoft.com/en-us/azure/storage/common/storage-account-keys-manage?tabs=azure-portal) on managing Shared Keys says “Avoid distributing access keys to other users.”
- Because they’re used for internal purposes, Shared Keys provide all permissions to a resource. For example, you can create and remove containers using only these keys. From what I could find, this is only inferred in Azure docs, but is stated by a Microsoft employee in this [response](https://stackoverflow.com/questions/19515733/account-key-vs-shared-access-signature/19548252#19548252). 
- The programmatic tools ([python](https://docs.microsoft.com/en-us/azure/storage/blobs/storage-quickstart-blobs-python#copy-your-credentials-from-the-azure-portal) for example) for connecting to blob storage use the Shared Key connection string for authentication. Because the connection string would provide admin-level access to the storage resource, I’d like to remove these as options for state use. 
- Azure docs show that the [REST API](https://docs.microsoft.com/en-us/rest/api/storageservices/authorize-with-shared-key) supports Shared Keys
 
My understanding is we’d like to use a connection string because it exposes less about our storage internals than a url. 
 
If AzCopy used a SAS connection string, that would be ideal, but it doesn’t seem to support it. 
 
If programmatic tools could use a SAS connection string, that would be more secure than exposing the Shared Key connection string to them and could make these tools viable for state use.

ryanhofdotgov:
@echappen OK – I think there's a few items to clarify; I'll get a Meet on our calendars and also reach out to our Microsoft friends:
* The SAS Connection string shouldn't expose any more internals to states than anything else. It's basically just compose of the URL (which they need to have) and the SAS token.
* The [comment on not supporting shared keys suggests OAuth as equivalent replacement, not SAS](https://github.com/Azure/azure-storage-azcopy/issues/186#issuecomment-460114558). OAuth would be a 3rd auth mechanism that we'd like to understand off-the-shelf support in tools/libraries.
* I think [this doc](https://docs.microsoft.com/en-us/azure/storage/common/storage-auth-aad) is critical to understand what Microsoft is recommending these days – basically no shared keys, use AAD, which can support OAuth or SAS URLs from AAD.
* We were planning to use SAS URLs for AAD managed identities for internal fetching from the storage accounts. 
* To complicate things even, more the SAS URLs that you can generate from the storage account settings (under Shared access signature) seem to not involve AAD? Those shared access signatures seem to be generated from either key1 or key2 share keys.

The good thing is that each state gets its own storage account, so "root" here doesn't mean as much as it could – it's all their own data.

ryanhofdotgov:
3 kinds of storage account SAS: User delegation SAS, Service SAS, Account SAS
https://github.com/MicrosoftDocs/azure-docs/blob/master/articles/storage/common/storage-sas-overview.md

etl/docs/uploads.md
```diff
@@ -0,0 +1,16 @@
...
+> ⚠️ This documentation is under active development and subject to change.
+
+Once you have [validated the format](./bulk-import.md) of your participant data CSV, our recommended approach for uploading the file is through AzCopy.
+#### AzCopy
+
+AzCopy is a Microsoft-supported command line utility that you can use to upload files to Azure blob storage resources.
+
+[Download AzCopy for your environment](https://docs.microsoft.com/en-us/azure/storage/common/storage-use-azcopy-v10#run-azcopy)
+
+For more details on using AzCopy, visit the [Microsoft documentation](https://docs.microsoft.com/en-us/azure/storage/common/storage-use-azcopy-blobs-upload).
```
ryanhofdotgov, line 12: I think we need a little more piipan-specific info here. e.g., given a storage account named XYZ, and credentials you get from us, set environment variable FOO, specify flag BAR, etc. On success, you will see: `ABC...`.

If the instructions need to be platform specific, then also need to specify you are using macOS, Linux, or Windows in the examples.

echappen: Updated with more details and cli examples. Is the gov't domain different from `[storage-account-name].blob.core.windows.net` and if so, should I include it here?

etl/docs/uploads.md
```diff
@@ -0,0 +1,35 @@
...
+> ⚠️ The service API for the bulk upload of PII data is not stable enough for client development. This documentation describes a _temporary_ approach that should allow states to quickly provide test data to our system with out expending development effort unnecessarily.
+
+Once you have [validated the format](./bulk-import.md) of your participant data CSV, you can upload the file to the system through AzCopy.
+### AzCopy
+
+AzCopy is a Microsoft-supported command line utility that you can use to upload files to Azure blob storage resources.
+
+[Download AzCopy for your environment](https://docs.microsoft.com/en-us/azure/storage/common/storage-use-azcopy-v10#run-azcopy)
+
+Upon receiving credentials, you'll first authorize AzCopy through a [service principal approach](https://docs.microsoft.com/en-us/azure/storage/common/storage-use-azcopy-authorize-azure-active-directory?toc=/azure/storage/blobs/toc.json#authorize-a-service-principal).
```
ryanhofdotgov, line 12: I think the state devs will benefit from a more gentle introduction as there is a good chance they don't use any cloud environments, don't use Azure, etc.... 

Making some assumptions since I haven't used the AzCopy yet, I'm guessing it would be something like:
The 18F team will provide these 5 pieces of information:
* Application id - this is like a username
* Tenant Id – this is the Azure identifier for the test environment
* service principal secret – this is the password for the user name
* storage account URL (we want to prep them for a time when they will be testing against an Azure government instance)

etl/docs/uploads.md
```diff
@@ -0,0 +1,35 @@
...
+
+Example (Unix/Linux):
+
+```
+$ export AZCOPY_SPA_CLIENT_SECRET="$(Read-Host -prompt "Enter key")"
```
ryanhofdotgov, line 17: Read-Host is a PowerShell command, so need something equivalent (or a different approach) for sh/bash. The best practice I'm guessing its usage conveys is that you are trying to avoid putting the secret into the shell history.

etl/docs/uploads.md
echappen: Yep, I'll list those here

etl/docs/uploads.md
```diff
@@ -0,0 +1,48 @@
+## Uploading Participant Data
+
+> ⚠️ The service API for the bulk upload of PII data is not stable enough for client development. This documentation describes a _temporary_ approach that should allow states to quickly provide test data to our system with out expending development effort unnecessarily.
+
+Once you have [validated the format](./bulk-import.md) of your participant data CSV, you can upload the file to the system through AzCopy.
+### AzCopy
+
+AzCopy is a Microsoft-supported command line utility that you can use to upload files to Azure blob storage resources.
+
+[Download AzCopy for your environment](https://docs.microsoft.com/en-us/azure/storage/common/storage-use-azcopy-v10#run-azcopy)
```
ryanhofdotgov, line 10:
```suggestion
[Download AzCopy for your environment](https://docs.microsoft.com/en-us/azure/storage/common/storage-use-azcopy-v10#download-azcopy)
```

etl/docs/uploads.md
```diff
@@ -0,0 +1,48 @@
...
+
+All of these items will vary between test and production environments.
+
+Upon receiving these credentials, you'll first authorize AzCopy through a [service principal approach](https://docs.microsoft.com/en-us/azure/storage/common/storage-use-azcopy-authorize-azure-active-directory?toc=/azure/storage/blobs/toc.json#authorize-a-service-principal).
+
+First save the password:
+
+PowerShell: `$env:AZCOPY_SPA_CLIENT_SECRET="$(Read-Host -prompt "Enter key")"`
+
+Bash: `export AZCOPY_SPA_CLIENT_SECRET=<password>`
```
ryanhofdotgov, line 29:
```suggestion
Bash: `read -s -p "Enter key " AZCOPY_SPA_CLIENT_SECRET && export AZCOPY_SPA_CLIENT_SECRET`
```
This will make the PowerShell and Bash commands equivalent. It's not relevant after this change, but as the password can have bash special characters like `$`, you'd have to enclose it in quotes before assigning to the var.

etl/docs/uploads.md
```diff
@@ -0,0 +1,48 @@
...
+- application id
+- tenant id
+- password (this is used as the client secret)
+- storage account url
+
+All of these items will vary between test and production environments.
+
+Upon receiving these credentials, you'll first authorize AzCopy through a [service principal approach](https://docs.microsoft.com/en-us/azure/storage/common/storage-use-azcopy-authorize-azure-active-directory?toc=/azure/storage/blobs/toc.json#authorize-a-service-principal).
+
+First save the password:
```
ryanhofdotgov, line 25:
```suggestion
First, save the password to the `AZCOPY_SPA_CLIENT_SECRET` environment variable, taking care to avoid saving it in your command-line history:
```
As the azcopy itself shouts every time you run it regarding passwords in command-line history, might be good to explicitly address it here.

ryanhofdotgov (approved):
Some suggestions, but I think this is in good shape – thanks!

PR #584, opened 2021-03-24
Initial documentation for APIM

### LINKED ISSUE #556: Document use of APIM
Documentation should include:
- Configuration details for our specific use case
- Process for making API requests
- Process for issuing and managing subscription keys
- Explanation for giving a subscription access to an API operation

kingcomma:
Closes #556 and final step of #28.

### DIFF
--- a/match/docs/public-api.md
+++ b/match/docs/public-api.md
@@ -1,0 +1,45 @@
+# Public API
+
+## Summary
+
+The public API is intended as a collection of external-facing endpoints for consumption by state systems. It is managed as an Azure API Management (APIM) instance and deployed by the [IaC](../../docs/iac.md).
+
+Currently the API includes a single endpoint which maps to the [orchestrator](orchestrator-match.md) API's query endpoint.
+
+For a general overview of APIM, refer to [Microsoft's documentation](https://docs.microsoft.com/en-us/azure/api-management/). Piipan makes use of the following concepts:
+
+- [APIs](https://docs.microsoft.com/en-us/azure/api-management/api-management-key-concepts#-apis-and-operations)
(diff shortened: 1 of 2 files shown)


match/docs/public-api.md
```diff
@@ -0,0 +1,45 @@
+# Public API
```
ryanhofdotgov, line 1: thought: we probably need to think about this nomenclature as this API isn't really public. *External* is more accurate, but not very precise.

match/docs/public-api.md
```diff
@@ -0,0 +1,45 @@
...
+- [Operations](https://docs.microsoft.com/en-us/azure/api-management/api-management-key-concepts#-apis-and-operations)
+- [Policies](https://docs.microsoft.com/en-us/azure/api-management/api-management-howto-policies)
+- [Versions](https://docs.microsoft.com/en-us/azure/api-management/api-management-versions)
+- [Subscriptions](https://docs.microsoft.com/en-us/azure/api-management/api-management-subscriptions)
+
+## Calling the API
+
+To call an endpoint:
+
+1. [Procure an API key](#managing-api-keys)
```
ryanhofdotgov, line 21:
```suggestion
1. [Obtain an API key](#managing-api-keys)
```
"Procure" in government usually implies some type of purchasing process.

ryanhofdotgov (approved):
LGTM – we can come back to the nomenclature. Perhaps a topic for our sync on Friday.

match/docs/public-api.md
kingcomma: Agree :100: I think the weekly sync is a great place to discuss.

Issue #452, opened 2021-03-04
Generate unique lookup ID in Orchestrator API


cjhskippy:
As a State Agency worker, I want to be able to share PII with another state agency securely, with as little data entry and memorization as possible so that I can process cases faster.

- As we currently conceive it, State agency workers will share PII by providing a unique Lookup ID to each other
- Only authorized personnel will have access to get PII from Lookup IDs - there isn't a strong need to make it hard to guess valid Lookup IDs
- The Lookup ID will be manually typed into emails and state agency systems.  In order to make this easy and avoid problems caused by typos, the unique identifier should:
  - be as short as we can reasonably make it
  - Use easily-typed characters (recommend case-insensitive alphabet + numbers)
  - Be displayed in the most easily-recognized way (recommend saving letters IN CAPS so the default is to display them that way by state agency systems too)
  - avoid the chances of Lookup IDs that are different by only one character - these increase the chances of a typo causing an agency to look up the wrong case information.
  - avoid characters that look identical at a glance (0/O and 1/I)
  - avoid including offensive words in our random strings
- Assuming we save up to 3 years of matches, 1% of queries result in a match, 40m participants overall, and queries aren't repeated for the same case, a name space of 12 million possibilities should be good enough. 6 character strings using 29 possible choices for each character (~600m possibilities) should be sufficient.

Closely related to #195 

AC:
- [x] Lookup IDs must be unique
- [x] Lookup IDs should not be sequential
- [x] Lookup IDs can include any capital letters except for vowels and any numbers except for 0 and 1 (exclude A,E,I,O,U,0,1)
- [x] Lookup IDs are six digits in length
- [x] When a match is returned by the Orchestrator API, the NAC should generate a new Lookup IDs and include the Lookup ID in its response.

juliaklindpaintner:
@cjhskippy I haven't solved for this problem before but after a brief amount of research on user-friendly alphanumeric codes here are my two cents (and I imagine none of this will be news to our engineers): 
- A common rule of thumb is that people can keep 7 characters in their mind +/-2, so 5 characters seems good, but I think up to 7 would be okay
- Definitely agree on case-INsensitive and no special characters (so base 36, which I think is how you got the ~60million permutations)
- Consider removing characters that can be ambiguous depending on the font, especially since we don't control that in state systems (0/O, 1/I) --> That would be base 32 for ~33million combinations in a 5 character string
[...]

PR #585, opened 2021-03-25
Stub lookup ID into orchestrator response

kingcomma:
- Update model
- Reflect null potential in spec

Small first step towards #452. Lookup ID generation has not been implemented. All responses will contain a `lookup_id` property with a `null` value. E.g.,
```
{
    "lookup_id": null,
    "matches": [
        {
            "last": "Lynn",
            "first": "Wesley",
            "middle": "Eura",
            "ssn": "000-12-3457",
            "dob": "1940-08-01",
            "exception": null,
            "state_name": "Echo Alpha",
            "state_abbr": "ea"
        }
    ]
}
```

### DIFF
--- a/match/docs/openapi/schemas/match-query.yaml
+++ b/match/docs/openapi/schemas/match-query.yaml
@@ -40,6 +40,7 @@ MatchQueryResponse:
   properties:
     lookup_id:
       type: string
+      nullable: true
       description: "the identifier of the match request"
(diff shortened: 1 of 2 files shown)


echappen (approved):
[non-blocking question] Do we want to make snake case an explicit convention for our api json? (eg. lookup_id vs lookupId)

PR #594, opened 2021-03-30
Update dependabot to reflect `metrics` directory structure

kingcomma:
`dependabot.yml` has references to `metrics` subsystem paths as they were before #442. Updated paths to reflect current directory structure.


echappen (approved):
Good catch, thanks!

Issue #195, opened 2021-01-08
Save match results for future reference


cjhskippy:
Blocked by #452

When the NAC finds a likely case of duplicate participation, there will always be a need to share PII between states, including participant name, date of birth, and SSN.  States do not have an easy, reliable, and secure way to do this.

In order to avoid sharing PII, we envision generating a unique identifier for every positive match that can be shared in plaintext.  Eligibility workers would then log in to the NAC, enter the identifier, and then find the PII.

As a state eligibility worker, I want a secure place where I can look up a NAC match that was found by another state so that I can determine if the person is still enrolled in SNAP benefits in my state.

Look to #452 for considerations on generating the lookup ID

AC:
When matches are found by API or manual query:
- [ ] The match query is saved:
  - [ ] Name, DOB, SSN of participant at initiating state
  - [ ] Whether person is a vulnerable individual at initiating state
- [ ] The unique lookup ID (see #452) is saved
  

Related to #196 + #259 - Users have a way to look up previous NAC matches through FNS website + API, respectively

juliaklindpaintner:
@cjhskippy Just to make it explicit/check my assumptions: 
- What is saved for the query and result is not necessarily what is displayed to the state agency user (research will inform what info is useful to them at query and ID lookup)
- I assume we are saving initiating state in order to track metrics. Have you thought of a reason why that data would be needed by the state agency workers? My only concern is that there is a way to distinguish a match result from a query by a state (which does not suggest existing duplicate participation) and a match result in the bulk match (which does suggest existing duplicate participation). I see you included "if any" so I'm assuming that in the bulk match this field would be empty or something? Τhat makes me wonder if there's any use in saving some data element that indicates which process a match was the result of? 
- I assume you are suggesting not saving DOB and SSN associated with the match result because those have to be exact matches? Any reason we would want to build for a future scenario where there might be some fuzziness allowed?
- I assume the above describes the bare minimum that should be saved. As additional optional elements are identified, those could be added (since we have heard that information about dates of eligibility, case status, etc may be useful).

cjhskippy:
1. (query result != what the SA worker sees) - agree!
2. Initiating state would probably not just be for metrics.  I would imagine that playing in to access controls - only workers at the states involved in a match would have permission to look up PII.  Also yes to the metrics use case.  Good call re: differentiating between reasons for initiating the match.  I do think we'll need to do that (issue #462 now) but I think we can defer it for now.
[...]

PR #598, opened 2021-03-31
Add Table Storage resource for lookup API

kingcomma:
[Azure Table Storage](https://docs.microsoft.com/en-us/azure/storage/tables/table-storage-overview#:~:text=Azure%20Table%20storage%20is%20a,needs%20of%20your%20application%20evolve.) seems like the right initial fit for storing queries that will later be retrieved via the lookup API. This PR modifies the IaC to create the necessary storage account and table resource, and bind the storage account to the orchestrator. Binding is done via a connection string — future work should switch to using shared access signatures (see #213).

Partially addresses #195 and #452.

### DIFF
--- a/iac/arm-templates/function-orch-match.json
+++ b/iac/arm-templates/function-orch-match.json
@@ -8,6 +8,9 @@
         "location": {
             "type": "string"
         },
+        "LookupStorageName": {
+            "type": "string"
+        },
         "StateApiUriStrings": {
             "type": "string"
         }
@@ -16,6 +19,7 @@
         "functionAppName": "[concat('ofunc', uniqueString(resourceGroup().id))]",
         "hostingPlanName": "matchapi",
         "storageAccountName": "[concat('ofstor', uniquestring(resourceGroup().id))]",
+        "lookupTableName": "lookup",
         "functionWorkerRuntime": "dotnet",
         "storageAccountType": "Standard_LRS",
         "systemTypeTag": {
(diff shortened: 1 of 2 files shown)


iac/arm-templates/function-orch-match.json
```diff
@@ -26,6 +30,42 @@
...
+            "name": "[parameters('LookupStorageName')]",
+            "tags": "[parameters('resourceTags')]",
+            "location": "[parameters('location')]",
+            "sku": {
+                "name": "[variables('storageAccountType')]"
+            },
+            "kind": "Storage"
+        },
+        {
+            "type": "Microsoft.Storage/storageAccounts/tableServices",
```
echappen, line 46: Would this be [NoSQL](https://docs.microsoft.com/en-us/azure/storage/tables/table-storage-overview)? I can't remember if we discussed the pros and cons of NoSQL over using Postgres with jsonb.

echappen: Thanks for clarifying. That all seems beneficial at this juncture.

ryanhofdotgov (approved):
LGTM – at some point it would be interesting to get better understanding between this approach and consumption/serverless Cosmos DB that just became available. They use the same Table API.

kingcomma:
@ryanhofdotgov Agreed. Table Storage seems like the lower barrier to entry and has a [small cost savings](https://docs.microsoft.com/en-us/azure/cosmos-db/table-api-faq#where-is-table-api-not-identical-with-azure-table-storage-behavior) (trading cost for performance). And since both use the same API, as you noted, hopefully it is straightforward to switch to a serverless Cosmos DB instance if needed.

PR #601, opened 2021-03-31
Add VNet resources and connect them to participant-records db

echappen:
Addresses #11 

See Comments on #11. Adding vnet rules required upgrading the database from Basic to GeneralPurpose tier. In order to upgrade, I had to drop and recreate the database server resource. 

I'm still not sure how the current firewall rules (Azure and GSA networks only) relate to the vnet/subnet. 

Todo:
- [x] incorporate vnet and subnet info into top-level iac

### DIFF
--- a/iac/arm-templates/participant-records.json
+++ b/iac/arm-templates/participant-records.json
@@ -29,6 +29,12 @@
         },
         "resourceTags": {
             "type": "object"
+        },
+        "vnetName": {
+            "type": "string"
+        },
+        "subnetName": {
+            "type": "string"
         }
     },
     "variables": {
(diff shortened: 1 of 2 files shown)


echappen:
@ryanhofdotgov @kingcomma Before I run the full iac with this config, does anything stand out to you?

kingcomma:
I don't feel like I have enough VNet knowledge for anything to stand out to me 😅 but I can at least say it shouldn't block me if running the full IaC has a temporary effect on other resource's ability to connect to the db.

ryanhofdotgov:
@echappen At a quick glance it LGTM, but don't have a lot of depth in the Azure specifics here right now. Any other firewall rules we've set up, are intended to just be temporary measures so internal things can talk to each other (or us devs can conveniently talk to the internal components). Once appropriate service endpoints, VNets are set up, those should mostly go away with the exception of the web apps. (The web apps still need an OIDC server in front of them before we remove those blocks).

echappen:
Running full iac was successful, opening for review

ryanhofdotgov:
Just looking at this very briefly, I think there's a second piece still to dive into; i.e., make sure the ETL BulkUpload Function as well as the per-state match API Function is connecting to the database via this internal network. 

If you pull up the VNet it has an info box that says "Selected subnet does not have service endpoint enabled for Microsoft.Sql. Enabling access may take up to 15 minutes to complete." I think that's part of the missing piece.

When everything is in place, we should be able to turn off the "Allow access to Azure service" option on this db, and then run through an end-to-end scenario (e.g., upload new CSV with a new PII record, see it get stored, query for it via web app for the first time).

echappen:
Thanks, this is helpful! I'll look into it.

PR #604, opened 2021-03-31
Save match query with unique lookup ID

### LINKED ISSUE #195: Save match results for future reference

kingcomma:
- When a match is found, save the query to Table Storage and return the lookup ID.
- If the lookup ID is already in use, verify that it is paired with a matching query and return the ID.
- If the lookup ID is already in use and paired with a *non-matching* query, throw an exception that results in a `500` response.

**Questions / considerations**:
- Is this the right approach to handling the possibility of lookup ID collisions?
- I've added a new `Lookup` class where I'm thinking the save/retrieve methods for the Lookup API can live.
- I've added a new `LookupStorage` class as a helper that wraps the basic Table Storage API methods we'll be using. My intent is for it to be usable by the Lookup API (i.e., use `LookupStorage.PointQueryAsync` to retrieve a specific query). It should also ease dependency injection / unit testing (though see below).
- I haven't found a good way to unit test `LookupStorage`. The `CloudTable` dependency isn't easily mocked. Integration tests should be possible but aren't yet implemented. Until then, it tanks code coverage.

Closes #195

### DIFF
--- a/docs/iac.md
+++ b/docs/iac.md
@@ -60,6 +60,8 @@ The following environment variables are pre-configured by the Infrastructure-as-
 | `MetricsApiUri` | URI for the Metrics API endpoint. | Piipan.Dashboard |
 | `KeyVaultName` | Name of key vault resource needed to acquire a secret | Piipan.Metrics.Api, Piipan.Metrics.Collect |
 | `CloudName` | Name of the active Azure cloud environment, either `AzureCloud` or `AzureUSGovernment` | Piipan.Etl, Piipan.Match.State, Piipan.Metrics.Api, Piipan.Metrics.Collect |
+| `LookupConnectionString` | Azure Storage Account connection string for accessing Table Storage service used for storing lookup IDs. | Piipan.Match.Orchestrator |
+| `LookupTableName` | Name of the Table Storage table where lookup IDs are stored.  | Piipan.Match.Orchestrator |
 
 
 ## `SysType` resource tag
--- a/match/docs/orchestrator-match.md
+++ b/match/docs/orchestrator-match.md
@@ -22,6 +22,8 @@ The following environment variables are required by `Query` and are set by the [
 | Name | |
 |---|---|
 | `StateApiUriStrings` | [details](../../docs/iac.md#\:\~\:text=StateApiUriStrings) |
+| `LookupConnectionString` | [details](../../docs/iac.md#\:\~\:text=LookupConnectionString) |
+| `LookupTableName` | [details](../../docs/iac.md#\:\~\:text=LookupTableName) |
 
 ## Binding to state APIs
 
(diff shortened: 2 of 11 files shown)


ryanhofdotgov:
I'm feeling very slow this week – I'm realizing all our back and forth about hashing input, and keeping it stable, etc. was really done outside this larger context and I gave you bad advice – sorry.

1. It's this layer's responsibility to handle situations where chosen look up id is already used. Bubbling up the exception on an initial collision would be bad.

1. We've talked about cryptographic hash to keep the resulting id immutable on the same input. That's a nice property, but to handle very rare collisions of input --> hash, and to always only return the single result (not the set of collisions), we are forced to take a different approach. We need to ditch the hash, just generate a 32 bits randomly (or 2^35, etc. to get the full randomness across full 7 base 29 characters without padding).

1. If there's a collision, get a new random number. If there's a 2nd collision, do it again. Put it in a loop, max out a 10? times and then bubble up a fatal error. That should never happen based on rates of matching, size of id. Log when there's a 2nd collision and hypothetically if it ever started happening often, someone would just have to decide to increase the size of the look up id.

kingcomma:
@ryanhofdotgov Not at a problem at all. This all makes sense and maps to the initial mental model I had before I tripped myself up on the idea of a 1:1 relationship between a lookup ID and a specific query. It won't be difficult to switch back. I'll mark this as draft while I implement.

One question that's still outstanding in my mind: Do we still want to achieve that 1:1 relationship between an ID and query? With the random generated ID, this would shake out to:
- Add a row to the table with `lookup_id` as the `RowKey` and the serialized PII tuple as the `Body` field (as happens now)
- Add a second row to the table with the serialized PII tuple as as the `RowKey` and the `lookup_id` as the `Body`

That second row would facilitate running a [point query](https://docs.microsoft.com/en-us/azure/cosmos-db/table-storage-design-guide#how-your-choice-of-partitionkey-and-rowkey-affects-query-performance) to check if the PII tuple has already been stored before generating the lookup ID and storing the new set of rows.

I'll push ahead without this functionality since it's not critical for the user story, but can adjust course if we deem it necessary.

kingcomma:
Superseded by #622.

PR #609, opened 2021-04-01
Update Azure.Core dependency and add query-tool to dependabot

kingcomma:
When updating the Azure.Core dependency the projects in `query-tool` were updated despite not being in the list of projects flagged by dependabot's PRs. They were missed because the paths were not in the dependabot config.

This adds the paths to the config and updates Azure.Core in all projects.


heymatthenry (approved):
Ty for catching this!

PR #618, opened 2021-04-01
Bump Microsoft.NET.Test.Sdk to 16.9.4 for all projects

echappen:


kingcomma (approved):
👍 though I think maybe `/shared/` will still need to be updated after merging?

PR #620, opened 2021-04-01
Query state APIs concurrently

### LINKED ISSUE #452: Generate unique lookup ID in Orchestrator API

kingcomma:
Make better use of async calls to per-state APIs by handling them concurrently rather than sequentially.

Closes #452

### DIFF
--- a/match/src/Piipan.Match.Orchestrator/Api.cs
+++ b/match/src/Piipan.Match.Orchestrator/Api.cs
@@ -136,12 +136,16 @@ namespace Piipan.Match.Orchestrator
             var stateRequests = new List<Task<MatchQueryResponse>>();
             var stateApiUris = StateApiUris();
 
-            // Loop through each state, compile results
-            // XXX Refactor to leverage async operations
             foreach (var uri in stateApiUris)
             {
-                var stateMatches = await MatchState(uri, request, log);
-                matches.AddRange(stateMatches.Matches);
+                stateRequests.Add(MatchState(uri, request, log));
+            }
+
+            await Task.WhenAll(stateRequests.ToArray());
+
+            foreach (var stateRequest in stateRequests)
+            {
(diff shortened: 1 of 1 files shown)


heymatthenry (approved):
Nice!

PR #621, opened 2021-04-01
Switch to randomly generated lookup ID

kingcomma:
Per conversation in #604, switches lookup ID to be a randomly generated 7-character string that conforms to our custom alphabet.

Uses the C# implementation of Nanoid (a popular [JavaScript tool](https://www.npmjs.com/package/nanoid)).

### DIFF
--- a/match/src/Piipan.Match.Orchestrator/Lookup.cs
+++ b/match/src/Piipan.Match.Orchestrator/Lookup.cs
@@ -1,7 +1,3 @@
-using System;
-using System.Security.Cryptography;
-using System.Text;
-
 namespace Piipan.Match.Orchestrator
 {
     static class LookupId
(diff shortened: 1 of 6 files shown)


match/src/Piipan.Match.Orchestrator/Lookup.cs
```diff
@@ -10,29 +6,13 @@ static class LookupId
...
-            // Encode using custom alphabet
-            do
-            {
-                idString.Insert(0, Chars[(int)(hashInt % Chars.Length)]);
-                hashInt = (long)(hashInt / Chars.Length);
-            } while (hashInt > 0);
-
-            // Returned IDs should always be `Length` length
-            return idString.ToString().PadLeft(Length, Chars[0]);
+            return Nanoid.Nanoid.Generate(Chars, Length);
```
ryanhofdotgov, line 15: Hurrah! This is exactly what I hoped you would find.

ryanhofdotgov (approved):
Nice! Was just waiting for builds to succeed again before approving.

PR #622, opened 2021-04-01
Save match query with unique lookup ID

### LINKED ISSUE #195: Save match results for future reference

kingcomma:
Refactored approach based on conversation in #604.

- When a match is found, save the query to Table Storage and return the lookup ID.
- If there is a lookup ID collision, generate a new ID and retry insert. Repeat up to 10 times before throwing exception.

Same considerations as the previous PR:
- I've added a new Lookup class where I'm thinking the save/retrieve methods for the Lookup API can live.
- I've added a new LookupStorage class as a helper that can wrap the basic Table Storage API methods we'll be using. My intent is for it to be usable by the Lookup API (i.e., use LookupStorage.PointQueryAsync to retrieve a specific query). It should also ease dependency injection / unit testing (though see below).
- I haven't found a good way to unit test LookupStorage. The CloudTable dependency isn't easily mocked. Integration tests should be possible but aren't yet implemented. Until then, it tanks code coverage.

Closes #195 and #452 (by ensuring IDs are unique in storage)

### DIFF
--- a/docs/iac.md
+++ b/docs/iac.md
@@ -60,6 +60,8 @@ The following environment variables are pre-configured by the Infrastructure-as-
 | `MetricsApiUri` | URI for the Metrics API endpoint. | Piipan.Dashboard |
 | `KeyVaultName` | Name of key vault resource needed to acquire a secret | Piipan.Metrics.Api, Piipan.Metrics.Collect |
 | `CloudName` | Name of the active Azure cloud environment, either `AzureCloud` or `AzureUSGovernment` | Piipan.Etl, Piipan.Match.State, Piipan.Metrics.Api, Piipan.Metrics.Collect |
+| `LookupConnectionString` | Azure Storage Account connection string for accessing Table Storage service used for storing lookup IDs. | Piipan.Match.Orchestrator |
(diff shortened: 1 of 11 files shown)


ryanhofdotgov (approved):
LGTM 👍

PR #624, opened 2021-04-02
Update per-state APIs to ignore middle name when performing match

### LINKED ISSUE #550: Match logic ignores middle name
As a FNS Business Owner, I want to avoid false negative matches so that we prevent simple ways to avoid duplicate enrollment.

We think there is likely to be a higher incidence of false positives if we require the middle name to be an exact match - it may be optional on some state forms or may just be an initial.

- [ ] Matches are found when the first name, last name in both states match, but only one state has a middle name.
- [ ] Matches are found when the first name, last name in both states match, but the middle name is completely different.
- [ ] Matches are found when the first name, last name in both states match, and neither state has a middle name.

Related to #541

kingcomma:
Updates per-state APIs to ignore middle name when performing match.

To test, follow the [directions to use the duplicate participation API](https://github.com/18F/piipan/blob/main/match/docs/duplicate-participation-api.md#calling-the-api) or submit a request directly to the [orchestrator API](https://github.com/18F/piipan/blob/main/match/docs/orchestrator-match.md#remote-testing). Example.csv has been uploaded to each state with a varying middle name for the `Lynn,Wesley,Eura,8/01/1940,000-12-3457,` row. Submitting a request for, for example:
```
{
    "query": {
        "last": "Lynn",
        "dob": "1940-08-01",
        "ssn": "000-12-3457",
        "first": "Wesley",
        "middle": "E"
    }
}
```

Should return the following results (note the differing `middle` property):

```
{
    "lookup_id": "...",
    "matches": [
        {
            "last": "Lynn",
            "first": "Wesley",
            "middle": null,
            "ssn": "000-12-3457",
            "dob": "1940-08-01",
            "exception": null,
            "state_name": "Echo Alpha",
            "state_abbr": "ea"
        },
        {
            "last": "Lynn",
            "first": "Wesley",
            "middle": "Eura",
            "ssn": "000-12-3457",
            "dob": "1940-08-01",
            "exception": null,
            "state_name": "Echo Bravo",
            "state_abbr": "eb"
        },
        {
            "last": "Lynn",
            "first": "Wesley",
            "middle": "E",
            "ssn": "000-12-3457",
            "dob": "1940-08-01",
            "exception": null,
            "state_name": "Echo Charlie",
            "state_abbr": "ec"
        }
    ]
}
```

Closes #550

### DIFF
--- a/match/src/Piipan.Match.State/Api.cs
+++ b/match/src/Piipan.Match.State/Api.cs
@@ -96,11 +96,9 @@ namespace Piipan.Match.State
                 dob = request.Query.Dob,
                 last = request.Query.Last,
                 first = request.Query.First,
-                middle = request.Query.Middle
             };
             var sql = "SELECT upload_id, first, last, middle, dob, ssn, exception FROM participants " +
                         "WHERE ssn=@ssn AND dob=@dob AND last=@last AND first=@first " +
-                        "AND " + (p.middle == null ? "middle IS NULL" : "middle=@middle") + " " +
                         "AND upload_id=(SELECT id FROM uploads WHERE created_at = (SELECT MAX(created_at) FROM uploads))";
 
             return (sql, p);
--- a/match/tests/Piipan.Match.State.Tests/ApiTests.cs
+++ b/match/tests/Piipan.Match.State.Tests/ApiTests.cs
@@ -225,22 +225,6 @@ namespace Piipan.Match.State.Tests
             Assert.Equal(bodyFormatted, request.ToJson());
         }
 
-        // SQL contains null
-        [Fact]
-        public void SqlHasIsNullCondition()
-        {
-            // Arrange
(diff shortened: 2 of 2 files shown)


echappen (approved):
👍 Might want to take Middle Name out of the api spec too

kingcomma:
I got tripped up thinking about that too. My understanding of the user story is that we still want to accept middle name even if it's not currently factored into the matching algorithm. And since the incoming query is stored in the lookup table on match, the EW will be able to see the exact info they entered on a previous search. Though I'm not sure how much value that provides, if any...

ryanhofdotgov:
Product owner was specifically concerned with having changes to the matching algorithm be easy to do, and not require API changes at the state level, so we left it in. Hypothetically, you could do fuzzy (or exact) matching on all or some of the data elements.

Issue #196, opened 2021-01-08
Look up NAC match via website


cjhskippy:
As a state eligibility worker, I want a secure place where I can look up a NAC match that was found by another state so that I can determine if the person is still enrolled in SNAP benefits in my state.

AC:
- [ ] Available in same website where eligibility workers can query the NAC
- [ ] Users can enter the unique lookup ID and get:
   - [ ] Name, DOB, SSN of the participant or applicant at the initiating state
   - [ ] Other states involved in the match, if any
     - [ ] The name used at the other states
- [ ] If lookup ID is not valid, provide error message (something like "no records exist for match code X").

juliaklindpaintner:
@cjhskippy Upon reflection during wireframing, I'm not sure it makes sense to display 'initiating state' in the results of a Match ID. Theoretically, a search could be conducted by multiple states over the course of time. If a match file is specific to an individual, rather than to an inquiry, the "initiating state" could change. Based on this user story, the information a user is looking for in searching by Match ID is simply the PII that allows them to look them up in their eligibility system. I agree it should show any state(s) in which the individual is currently enrolled, but perhaps this is not framed as "matching state." Happy to discuss/elaborate!

cjhskippy:
Ooh, I think this will make for good discussion.  In my head, I was envisioning one match file per person per case action (application, recertification, addition of a household member).  The match would be resolved alongside that case action and would not persist to other times when the same individual is involved in another potential match.

Issue #455, opened 2021-03-04
Display lookup ID in query website


cjhskippy:
Depends on #195 or a related story to generate match IDs when queries produce results.

As a State agency worker, I want to be able to share PII with another state agency securely when I encounter a NAC match so that I can prevent duplicate SNAP participation.

We are currently assuming that every match will create a unique lookup ID that can be shared with the other state.  The website will need to provide this ID alongside the other match details.

AC:
When a match is encountered in the query website:
- [ ] A lookup ID is generated just like in #452 
- [ ] The lookup ID is displayed to the user alongside the other match results.

PR #625, opened 2021-04-02
Add lookup ID display and search to Query Tool

heymatthenry:
This PR does a few things to get the Query Tool ready to do queries based on Lookup IDs:

* Create a new page with a one-field form
* Extract the results table into a partial to share between both forms
* Adds the lookup ID field to the `PiiRecord` class
* Makes a new `Lookup` class
* Relates `Lookup` and `PiiRecord` through a `IQueryable` interface

To that last point, I opted to use an interface rather than inheritance because while both classes needed a field for the Lookup ID, it needs to be required on one but not the other, and the only ways I could think of doing that via inheritance would've required method overrides that defeated the purpose of inheritance.

Finally, while I did do a `OnPostAsync` method that returned a dummy record when you submitted the Lookup ID form, I did not include it in this PR because I didn't think it should actually ship.

### DIFF
--- a/query-tool/src/Piipan.QueryTool/OrchestratorApiRequest.cs
+++ b/query-tool/src/Piipan.QueryTool/OrchestratorApiRequest.cs
@@ -20,10 +20,10 @@ namespace Piipan.QueryTool
         }
 
         public string RequestUrl;
-        public Dictionary<string, PiiRecord> Query = new Dictionary<string, PiiRecord>();
+        public Dictionary<string, IQueryable> Query = new Dictionary<string, IQueryable>();
         private readonly IAuthorizedApiClient _apiClient;
 
-        public async Task<List<PiiRecord>> SendQuery(string url, PiiRecord query)
+        public async Task<List<PiiRecord>> SendQuery(string url, IQueryable query)
         {
             RequestUrl = url;
             Query.Add("query", query);
--- a/query-tool/src/Piipan.QueryTool/OrchestratorApiRequest.cs
+++ b/query-tool/src/Piipan.QueryTool/OrchestratorApiRequest.cs
@@ -20,13 +20,18 @@ namespace Piipan.QueryTool
         }
 
         public string RequestUrl;
-        public Dictionary<string, IQueryable> Query = new Dictionary<string, IQueryable>();
+        public Dictionary<string, object> Query = new Dictionary<string, object>();
         private readonly IAuthorizedApiClient _apiClient;
 
         public async Task<List<PiiRecord>> SendQuery(string url, IQueryable query)
         {
             RequestUrl = url;
-            Query.Add("query", query);
+
+            if (query is PiiRecord) {
+                Query.Add("query", query as PiiRecord);
+            } else if (query is Lookup) {
+                Query.Add("query", query as Lookup);
+            }
             return await QueryOrchestrator();
         }
 
--- a/query-tool/src/Piipan.QueryTool/OrchestratorApiRequest.cs
+++ b/query-tool/src/Piipan.QueryTool/OrchestratorApiRequest.cs
@@ -20,16 +20,19 @@ namespace Piipan.QueryTool
         }
 
         public string RequestUrl;
-        public Dictionary<string, object> Query = new Dictionary<string, object>();
+        public Dictionary<string, IQueryable> Query = new Dictionary<string, IQueryable>();
         private readonly IAuthorizedApiClient _apiClient;
 
         public async Task<List<PiiRecord>> SendQuery(string url, IQueryable query)
         {
             RequestUrl = url;
 
-            if (query is PiiRecord) {
+            if (query is PiiRecord)
+            {
                 Query.Add("query", query as PiiRecord);
-            } else if (query is Lookup) {
+            }
+            else if (query is Lookup)
+            {
                 Query.Add("query", query as Lookup);
             }
             return await QueryOrchestrator();
--- a/query-tool/src/Piipan.QueryTool/OrchestratorApiRequest.cs
+++ b/query-tool/src/Piipan.QueryTool/OrchestratorApiRequest.cs
@@ -23,50 +23,111 @@ namespace Piipan.QueryTool
         public Dictionary<string, IQueryable> Query = new Dictionary<string, IQueryable>();
         private readonly IAuthorizedApiClient _apiClient;
 
-        public async Task<List<PiiRecord>> SendQuery(string url, IQueryable query)
+        public async Task<Dictionary<string, object>> SendQuery(string url, Lookup query)
         {
             RequestUrl = url;
             Query.Add("query", query);
             return await QueryOrchestrator();
         }
 
-        public List<PiiRecord> Matches { get; private set; }
+        public async Task<Dictionary<string, object>> SendQuery(string url, PiiRecord query)
+        {
+            RequestUrl = url;
+            Query.Add("query", query);
+            return await QueryOrchestrator();
+        }
 
-        private async Task<List<PiiRecord>> QueryOrchestrator()
+        public Dictionary<string, object> ResponseDict { get; private set; }
+
+        private async Task<Dictionary<string, object>> QueryOrchestrator()
         {
             try
             {
                 _logger.LogInformation("Querying Orchestrator API");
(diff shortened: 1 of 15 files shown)


query-tool/src/Piipan.QueryTool/OrchestratorApiRequest.cs
```diff
@@ -20,10 +20,10 @@ public OrchestratorApiRequest(IAuthorizedApiClient apiClient, ILogger logger)
         }
 
         public string RequestUrl;
-        public Dictionary<string, PiiRecord> Query = new Dictionary<string, PiiRecord>();
+        public Dictionary<string, IQueryable> Query = new Dictionary<string, IQueryable>();
         private readonly IAuthorizedApiClient _apiClient;
 
-        public async Task<List<PiiRecord>> SendQuery(string url, PiiRecord query)
+        public async Task<List<PiiRecord>> SendQuery(string url, IQueryable query)
```
kingcomma, line 26: I'm hitting an error resulting in the "There was an error running your search" message when I attempt to run a match query either locally or remotely (after I deployed this branch's code).

I think this change is the culprit. Now, when `Query` is serialized in `QueryOrchestrator` ([line 41](https://github.com/18F/piipan/blob/7c0b2798fa024041e0ed93ed9afef2f9b5727992/query-tool/src/Piipan.QueryTool/OrchestratorApiRequest.cs#L41)) the result is a string like `{"query":{"LookupId":null}}`.

I'm guessing this is because `Query` is now being serialized as `IQueryable` where previously it was being serialized as `PiiRecord`.

If you add a simple `Console.WriteLine(jsonString);` after line 41, do you see the same output?

heymatthenry: Yep, you were right. I had to cast before serializing which then necessitated some fiddling with the type annotations for `Query`.

heymatthenry:
@kingcomma Ok, back in business! And somewhere in there I did a rebase from `main` instead of a merge. And sorry for the extraneous commits. I didn't notice until I had pushed 😬

ryanhofdotgov:
@heymatthenry Any chance we can clean this PR up so it's not as noisy? I think perhaps just create a new local branch from main, then cherry pick the 8 or so commits that are legit? I can pair with you later today if that would help.

kingcomma:
@ryanhofdotgov @heymatthenry I was able to clear up a lot of the noise by [temporarily switching the target branch](https://stackoverflow.com/a/46782679) to something other than `main` and then switching it back to `main`. This brought the list of commits down to 7 and limits the diff to only the new incoming changes. I think this resolves the concern?

ryanhofdotgov:
@kingcomma Thanks - that solves it.

query-tool/src/Piipan.QueryTool/OrchestratorApiRequest.cs
```diff
@@ -20,13 +20,18 @@ public OrchestratorApiRequest(IAuthorizedApiClient apiClient, ILogger logger)
...
         {
             RequestUrl = url;
-            Query.Add("query", query);
+
+            if (query is PiiRecord) {
+                Query.Add("query", query as PiiRecord);
+            } else if (query is Lookup) {
+                Query.Add("query", query as Lookup);
```
kingcomma, line 33: *thought*: Now that we have a working API, it turns out the lookup call will be a `GET` not a `POST`. Since `QueryOrchestrator` uses `_apiClient.PostAsync`, there may need to be some additional logic or perhaps even a different method to handle lookups.

Also, `AuthorizedJsonApiClient` in the shared library doesn't have its `GetAsync` method wired up yet 😅 . I'll open an issue and get that settled today.

query-tool/src/Piipan.QueryTool/PiiRecord.cs
```diff
@@ -41,5 +41,9 @@ public class PiiRecord
         [Display(Name = "StateAbbr")]
         [JsonPropertyName("state_abbr")]
         public string StateAbbr { get; set; }
+
+        [Display(Name = "Lookup ID")]
+        [JsonPropertyName("lookupId")]
```
kingcomma, line 46: *thought*: The [API spec](https://github.com/18F/piipan/blob/main/match/docs/openapi/schemas/match-query.yaml#L41) currently defines this property as `lookup_id` in responses from the orchestrator. Not sure if you'll be serializing the response JSON directly to `PiiRecord`, but flagging just in case.

kingcomma (commented):
I think this is OK to merge, but now that the lookup API is functional I think there's also a little bit of refactoring that will need to be done. For example, using a `GET` request for lookup ID queries.

Curious how you want to handle, @heymatthenry? Merge and iterate on a new branch, or keep this open and make changes on this PR?

heymatthenry:
Getting there but not quite done. In e10550a, I:

* Overhauled `QueryOrchestrator` `OrchestratorApiRequest` to return a dictionary instead of a list of `PiiRecord`s
* Redid the views to account for this (So the old PII-based queries are up & running again)
* Got `QueryOrchestrator` sending (I think?) well-formed LookupID queries, but there's still a little error in there keeping it from actually rendering

So still TBD:

* Fix the `null`-related error in `QueryOrchestrator` referenced in the last bullet ☝🏻 
* Add the LookupID to the `_ResultsTable` partial
* Update the tests to reflect the changes to `OrchestratorApiRequest`

kingcomma:
This is ready for review. This PR:

- Updates the API binding to use the base URI (rather than the `/query` endpoint)
- Adds the necessary request methods to `OrchestratorApiRequest` for calling either match or lookup endpoints
- Adds the lookup ID to the match results table
- Adds a new `/LookupById` page for searching by lookup ID
- Updates unit tests

Testing can be done on the remote app:
1. Connect to the GSA network and load the query tool's Front Door URL
2. Submit a valid match query that generates a lookup ID
3. Load `/LookupById` to use the lookup ID to retrieve the stored query

Testing can also be done locally:
1. Ensure your CLI account has been [added to the Orchestrator's app role](https://github.com/18F/piipan/blob/main/match/docs/orchestrator-match.md#remote-testing)
2. Run the app locally with `dotnet run`
3. Submit a valid match query that generates a lookup ID
4. Load `/LookupById` to use the lookup ID to retrieve the stored query

Closes #196 and #455.

echappen:
[question] In my remote testing, it seems the lookup ID query is case sensitive. I think we decided in [this discussion](https://github.com/18F/piipan/issues/452) that it should be case insensitive. Was this unintentional, or did we decide differently?

Also, the query seems to strip trailing whitespace, but not leading whitespace. For example, "XRLYKNW  " will return results but "  XRLYKNW" will not. Not a huge deal, but wanted to flag it. 

Screen Shot 2021-04-20 at 9 31 29 AM
Screen Shot 2021-04-20 at 9 31 19 AM

kingcomma:
Great catch @echappen. Since the case-insensitivity fix is on the API-side, I've created a new issue for handling it (#713).

Whitespace handling will require some looking into — whitespace (leading or trailing) is not explicitly trimmed, but trailing whitespace seems to be getting dropped from the lookup ID at some point between submission and the API request being received. I'll follow up when I have some more info.

echappen:
I should have specified that I don't think the whitespace thing is a blocker. Since the case sensitivity is handled elsewhere, this PR looks good to me 👍

echappen (approved):

kingcomma:
It turns out the trailing whitespace is being stripped when the lookup API endpoint is converted to a `Uri` instance. E.g.:
```
var urlString = "https://localhost/string     ";
var uri = new Uri(urlString);
Console.WriteLine(uri.ToString());
```
Outputs
```
"https://localhost/string"
```

Since the lookup ID is passed in the URL as a `GET` request, whitespace is effectively trimmed. But that is not the case for the query requests where form data is passed in the body of a `POST` request.

ryanhofdotgov:
@kingcomma Haven't looked at the code, but the input field string should not just be tacked on to the query URL and being made. If it isn't happening already, it should validate that it consists only of valid input (no spaces, no special chars, only the allowed characters in a look up id).

Do we need to make a separate ticket for that follow-on?

kingcomma:
Yep, just created the ticket for that work #724

Issue #578, opened 2021-03-23
Integration tests for an API


cjhskippy:
Set up first automated integration tests for APIs in the NAC.

- [ ] at least oneAPI has at least one integration test that runs automatically on code commits, authenticates properly, shares results

kingcomma:
Reference: https://github.com/18F/piipan/pull/555#pullrequestreview-618028972

PR #629, opened 2021-04-02
Add lookup API endpoint

### LINKED ISSUE #259: Look up previous NAC match via API
Depends on #195 

As a state agency eligibility worker responding to a NAC match, I want to be able to look up the people involved in the match within my eligibility system.

Another motivation for this feature is to avoid the need for identity management of individual SA workers - because this look up is provided via API, access to Piipan can be granted to an entire system.

It should be possible to use the same credentials for this API function and the NAC query API.  We anticipate the same systems needing access to both of these capabilities.

The lookup does NOT need to check to see if the match is still current - it just needs to show the PII that was related to a match that happened in the past.

AC:
- [ ] Orchestrator Match API updated to return a lookup ID on match
  - [ ] Specification updated 
- [ ] State agency systems can look up an existing NAC match via API call
    - [ ] Specification updated 
- [ ] API calls can provide the unique lookup ID and get:
  - [ ] Name, DOB, SSN of the participant
  - [ ] ~~If the individual is flagged as vulnerable (true / false)~~
  - [ ] ~~The state that initiated the match~~ (moved to #628)
  - [ ] ~~Other states involved in the match~~ (moved to #627)
- [ ] If lookup ID is not valid, provide error message (something like "no records exist for match code X").
- [ ] Only states that are involved in the match should be able to access the match information.  An attempt by an authorized API to access an unrelated match should return a 403 Access Denied response.

kingcomma:
- Updates orchestrator Function App with `LookupIds` entry point
  - Queries table storage for a row with `RowKey` matching the incoming `lookupId`
  - Returns a response containing the matching query or `null` if none
- Adds a simple bash script for manual integration testing (as a precursor to #578)

**Question**: Should the incoming lookup ID be validated according to our schema (and presumably trigger `400` responses if invalid)?

Partially addresses #259 — API will be added to APIM instance as follow on work.

---

This work can be tested locally (with a connection to the remote lookup table) using the following steps:
1. Fetch app settings for the orchestrator app (to get down binding details for the lookup table):
```
cd match/src/Piipan.Match.Orchestrator
func azure functionapp fetch-app-settings <orchestrator-name>
```
2. Start up the orchestrator app:
```
func start
```
3. Send a `GET` request to the local endpoint (e.g., `http://localhost:7071/api/v1/lookups_ids/{lookupId}`)

Remote tests can be run using the simple integration test script at `match/tools/test-lookup-api.bash`:
1. (If not already completed) Add your CLI user to `OrchestratorApi.Query` application role for the orchestrator app:
```
# project root
./tools/assign-app-role.bash tts/dev <orchestrator-name> OrchestratorApi.Query
```
2. Run the integration test:
```
./match/tools/test-lookup-api.bash tts/dev <lookup-id>
```

You can find existing lookup IDs at Portal > Storage Accounts > stlookupapidev > Storage Explorer > Tables > lookup.

### DIFF
--- a/match/src/Piipan.Match.Orchestrator/Api.cs
+++ b/match/src/Piipan.Match.Orchestrator/Api.cs
@@ -79,6 +79,24 @@ namespace Piipan.Match.Orchestrator
             return (ActionResult)new JsonResult(response);
         }
 
+        /// <summary>
+        /// API endpoint for retrieving a MatchQuery using a lookup ID
+        /// </summary>
+        /// <param name="req">incoming HTTP request</param>
+        /// <param name="lookupId">lookup ID string (pulled from route)</param>
+        /// <param name="log">handle to the function log</param>
+        [FunctionName("lookup_ids")]
+        public async Task<IActionResult> LookupIds(
+            [HttpTrigger(AuthorizationLevel.Anonymous, "get", Route = "lookups_ids/{lookupId}")] HttpRequest req,
+            string lookupId,
+            ILogger log)
+        {
+            LookupResponse response = new LookupResponse { Data = null };
+            response.Data = await Lookup.Retrieve(lookupId, _lookupStorage, log);
+
+            return (ActionResult)new JsonResult(response);
+        }
+
         private MatchQueryRequest Parse(string requestBody, ILogger log)
         {
             // Assume failure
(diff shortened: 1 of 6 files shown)


echappen (approved):
> Should the incoming lookup ID be validated according to our schema (and presumably trigger 400 responses if invalid)?

I could see this being helpful to end users by leading them to correct a malformed ID, instead of thinking it just wasn't found. But it doesn't seem necessary at this time imo. 

LGTM. Thanks for writing that integration testing tool!

PR #630, opened 2021-04-05
Case insensitive matching

### LINKED ISSUE #551: Case-insensitive matching
If states don't save their data with upper and lower case characters, don't let that screw up matching against states that allow both.

kingcomma:
- Normalize first and last to uppercase in query
- Optimize upload ID subquery
- Add manual integration test script

CSV with varying capitalization for the `Lynn,Wesley,Eura,8/01/1940,000-12-3457,` row has been uploaded to each state's participants db. The updated query can be tested using `match/tools/test-match-api.bash`.

@ryanhofdotgov I had forgotten that we already had an index on `(ssn, upload_id)`.  That is sufficient to keep the match query to an index scan, and adding an additional index on `ssn` alone seemed unnecessary.

Because of that index, normalizing with `upper()` did not have an effect on the cost:

```
-- Query without upper
 Index Scan using participants_ssn_idx on participants  (cost=0.33..2.55 rows=1 width=48)
   Index Cond: ((ssn = '000-12-3461'::text) AND (upload_id = $0))
   Filter: ((dob = '1913-04-04'::date) AND (last = 'Wallen'::text) AND (first = 'Homer'::text))
   InitPlan 1 (returns $0)
     ->  Limit  (cost=0.15..0.18 rows=1 width=4)
           ->  Index Only Scan Backward using uploads_pkey on uploads  (cost=0.15..30.30 rows=1130 width=4)
```

```
-- Query with upper
 Index Scan using participants_ssn_idx on participants  (cost=0.33..2.56 rows=1 width=48)
   Index Cond: ((ssn = '000-12-3461'::text) AND (upload_id = $0))
   Filter: ((dob = '1913-04-04'::date) AND (upper(last) = 'WALLEN'::text) AND (upper(first) = 'HOMER'::text))
   InitPlan 1 (returns $0)
     ->  Limit  (cost=0.15..0.18 rows=1 width=4)
           ->  Index Only Scan Backward using uploads_pkey on uploads  (cost=0.15..30.30 rows=1130 width=4)
```

Closes #551

### DIFF
--- a/match/src/Piipan.Match.State/Api.cs
+++ b/match/src/Piipan.Match.State/Api.cs
@@ -98,8 +98,8 @@ namespace Piipan.Match.State
                 first = request.Query.First,
             };
             var sql = "SELECT upload_id, first, last, middle, dob, ssn, exception FROM participants " +
-                        "WHERE ssn=@ssn AND dob=@dob AND last=@last AND first=@first " +
-                        "AND upload_id=(SELECT id FROM uploads WHERE created_at = (SELECT MAX(created_at) FROM uploads))";
+                        "WHERE ssn=@ssn AND dob=@dob AND upper(last)=upper(@last) AND upper(first)=upper(@first) " +
+                        "AND upload_id=(SELECT id FROM uploads ORDER BY id DESC LIMIT 1)";
 
             return (sql, p);
         }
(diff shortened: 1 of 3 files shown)


ryanhofdotgov (approved):
👍

Issue #639, opened 2021-04-05
Update documentation for match APIs


kingcomma:
Now that most of the implementation details for the match APIs has come into focus, the docs could use a cleaning up. Edits should:

- Clarify the boundaries for each API, Function App, and the APIM instance
- Shore up deployment and testing instructions

PR #640, opened 2021-04-05
Add lookup API to APIM, update documentation

kingcomma:
Adds a new operation for the lookup API to the existing APIM instance. To test, send a conforming request to the API endpoint and include an active [API/subscription key](https://github.com/18F/piipan/blob/main/match/docs/duplicate-participation-api.md#calling-the-api) in the request's header. E.g.,
```
curl --location --request GET 'https://apim-duppartapi-dev.azure-api.net/v1/lookup_ids/<lookup-id>' \
  --header 'Content-Type: application/json' \
  --header 'Ocp-Apim-Subscription-Key: <api-key>'
```

This PR also includes some barebones documentation for the lookup API. I created #639 to take a comprehensive pass over the documentation now that most of the details have come into focus.

### DIFF
--- a/iac/arm-templates/apim.json
+++ b/iac/arm-templates/apim.json
@@ -139,6 +139,57 @@
                 "value": "[parameters('policyXml')]",
                 "format": "xml"
             }
+        },
+        {
+            "type": "Microsoft.ApiManagement/service/apis/operations",
+            "apiVersion": "2020-06-01-preview",
+            "name": "[concat(parameters('apiName'), '/', variables('matchSetName'), '/get-lookup')]",
+            "dependsOn": [
+                "[resourceId('Microsoft.ApiManagement/service', parameters('apiName'))]",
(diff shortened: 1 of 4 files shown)


kingcomma:
Closes #259

echappen:
Would `test-lookup-api.bash` use `apim-duppartapi-dev.azure-api.net` now?

kingcomma:
That's a good question. My quick scripts were intended to facilitate testing the internal APIs (i.e., calling the Function Apps directly) where the process of requesting a bearer token from the CLI and then submitting the request via curl or Postman was a bit unwieldy. That process was also easily automated (obtaining an API key, less so).

I'm of the mind that since integration testing needs to be improved across the board (#578 and #154), it's better to take the time to determine the right approach and handle APIM tests in separately. But open to being convinced otherwise.

echappen (approved):

Issue #659, opened 2021-04-07
Codify Database Private Link setup into iac script


echappen:
Based on what we learned in #11, we will need to change the iac script in order to establish a private connection from participant-records database to the Azure resources that connect to it:

- [ ] Shift state Match resources to same resource group as participant-records DB (rg-core)
- [ ] Upgrade all function apps that need to talk to participant-records DB to P1V2 sku from Standard
- [ ] Create VNet and Private Endpoint resources in `participant-records.json` ARM template
- [ ] Disable public access and access to all azure resources on DB programmatically (not sure if this can be done in ARM template or will have to be done in CLI or portal)
- [ ] In state function app template, integrate with same VNet created above programmatically (ARM template would be ideal)
- [ ] In match function app template, integrate with same VNet above programmatically (ARM template would be ideal)

echappen:
@kingcomma @ryanhofdotgov Moving state match function apps into rg-core will cause a name conflict with the etl function apps.

Based on our [naming conventions](https://github.com/18F/piipan/blob/dfa37755de66f4b06460ca2a62644a34e0125231/docs/azure-naming-conventions.md) I propose changing the names of these resources to something like this (using `ea` state as an example:

| resource | name |
|---|---|
| Etl function app | func-ea-etl-dev |
| Etl function app storage account | stor-ea-etl-dev |
| Match function app | func-ea-match-dev |
| Match function app storage account | stor-ea-match-dev |

Does this look good to you?

_Context on why we're moving these resources into the same resource group is that private endpoints are only private when talking to resources within the same resource group._

ryanhofdotgov:
That looks good, modulo storage accounts not allowing hyphens in its name and we will need the apps and the storage accounts to support PREFIX in its name.

echappen:
Thanks! Updated:

| resource | name |
|---|---|
| Etl function app | fns-func-ea-etl-dev |
| Etl function app storage account | fnsstoreaetldev |
| Match function app | fns-func-ea-match-dev |
| Match function app storage account | fnsstoreamatchdev |

kingcomma:
Looks good to me too, including @ryanhofdotgov's comments.

ryanhofdotgov:
@echappen Just double-checking, but all the check list items on this ticket have been done and we can close this issue?

echappen:
Yep!

PR #678, opened 2021-04-09
Set up private network for database and function apps in iac

echappen:
Reworking of #601, closes issues #11 and #659

Does the following:

- Shifts match api apps to core resource group (because VNets are resource-group-specific)
- Creates a single premium app service plan; puts both ETL and Match apps onto it (VNet integration is only available on premium plans or higher)
- Creates virtual network resource in separate ARM template with 2 subnets: one for the database private link, the other for the function apps
- Configures private endpoint on VNet when creating participant-records database
- Integrates ETL and Match apps with VNet
- Locks down public access to database at the end of iac by removing firewalls and changing DenyPublicAccess setting on it

Todo:
- [X]  Run on tts/dev `-dev` resources

### DIFF
--- a/iac/arm-templates/function-state-match.json
+++ b/iac/arm-templates/function-state-match.json
@@ -5,6 +5,9 @@
         "resourceTags": {
             "type": "object"
         },
+        "stateAbbr": {
+            "type": "string"
+        },
         "stateName": {
             "type": "string"
         },
@@ -146,6 +149,10 @@
                         {
                             "name": "CloudName",
                             "value": "[parameters('cloudName')]"
+                        },
+                        {
+                            "name": "StateAbbr",
+                            "value": "[parameters('stateAbbr')]"
                         }
                     ]
                 }
--- a/iac/arm-templates/function-state-match.json
+++ b/iac/arm-templates/function-state-match.json
@@ -8,9 +8,6 @@
         "stateName": {
             "type": "string"
         },
-        "stateAbbr": {
-            "type": "string"
-        },
         "location": {
             "type": "string"
         },
@@ -25,12 +22,24 @@
         },
         "cloudName": {
             "type": "string"
+        },
+        "functionAppName": {
+            "type": "string"
+        },
+        "storageAccountName": {
+            "type": "string"
+        },
+        "managedIdentityName": {
+            "type": "string"
+        },
+        "appServicePlanName": {
+            "type": "string"
         }
     },
     "variables": {
-        "functionAppName": "[concat(parameters('stateAbbr'), 'func', uniqueString(resourceGroup().id))]",
-        "hostingPlanName": "matchapi",
-        "storageAccountName": "[concat(parameters('stateAbbr'), 'fstor', uniquestring(resourceGroup().id))]",
+        "functionAppName": "[parameters('functionAppName')]",
+        "storageAccountName": "[parameters('storageAccountName')]",
+        "managedIdentityName": "[parameters('managedIdentityName')]",
         "functionWorkerRuntime": "dotnet",
         "storageAccountType": "Standard_LRS",
         "systemTypeTag": {
@@ -39,7 +48,7 @@
         "applicationInsightsTag": {
             "[concat('hidden-link:', resourceId('Microsoft.Web/sites', variables('functionAppName')))]": "Resource"
         },
-        "identityName": "[concat(parameters('stateAbbr'), 'admin')]"
+        "identityName": "[variables('managedIdentityName')]"
     },
     "resources": [
         {
@@ -52,21 +61,6 @@
                 "name": "[variables('storageAccountType')]"
             },
             "kind": "Storage"
-        },
-        {
-            "type": "Microsoft.Web/serverfarms",
-            "apiVersion": "2020-06-01",
-            "name": "[variables('hostingPlanName')]",
-            "tags": "[parameters('resourceTags')]",
-            "location": "[parameters('location')]",
-            "sku": {
-                "name": "Y1",
-                "tier": "Dynamic"
-            },
-            "properties": {
-                "name": "[variables('hostingPlanName')]",
-                "computeMode": "Dynamic"
-            }
         },
         {
             "type": "Microsoft.Web/sites",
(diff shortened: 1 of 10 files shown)


echappen:
@ryanhofdotgov @kingcomma Before I run this on `-dev` resources in tts sandbox, any feedback?

iac/arm-templates/function-state-match.json
```diff
@@ -142,10 +135,6 @@
                             "name": "StateName",
                             "value": "[parameters('stateName')]"
                         },
-                        {
-                            "name": "StateAbbr",
-                            "value": "[parameters('stateAbbr')]"
-                        },
```
kingcomma, line 148: This was not well documented, but the `StateAbbr` config setting is used to populate the `PiiRecord.StateAbbr` instances returned by the per-state APIs. We'll need to leave this here (and still pass it in as a parameter) to keep the APIs working for now.

iac/arm-templates/function-state-match.json
```diff
@@ -82,11 +76,10 @@
                 }
             },
             "dependsOn": [
-                "[resourceId('Microsoft.Web/serverfarms', variables('hostingPlanName'))]",
```
kingcomma, line 85: Hopefully not an issue, but now that this resource is created elsewhere and we aren't relying on `dependsOn`, I wonder if there's a chance we'll run into another "resource does not exist" race condition.

echappen: I didn't run into this issue after running the script quite a few times. The App Service Plan is created pretty early on in the script, ~20 minutes (for me) before these apps are deployed.

iac/arm-templates/function-state-match.json
echappen: Thanks for clarifying! Added it back.

iac/create-resources.bash
```diff
@@ -29,6 +29,12 @@ set_constants () {
   # Name of PostgreSQL server
   PG_SERVER_NAME=participant-records
 
+  # Names of participant records database Vnet and Subnet
+  VNET_NAME=$PREFIX-vnet-core-$ENV
```
ryanhofdotgov, line 33: nitpick: as this don't refer to global resources, we can drop the prefix.

iac/secure-resources.bash
```diff
@@ -0,0 +1,48 @@
...
+# denying public access to database.
+# Once done, the only access to database would be through its private endpoint.
+#
+# Arguments:
+#   env (eg: tts/dev)
+#   resource_group (eg: rg-core-dev)
+#   server_name (eg: fns-db-participant-records-dev)
+#
+# Usage:
+# ./iac/secure-resources.bash tts/dev rg-core-dev fns-db-participant-records-dev
```
ryanhofdotgov, line 12: niptick: I'm inclined to call it something other than `secure-resources`, since they are already secure (to a degree) before you run this script. This is just creating another layer of security by removing unnecessary network access, so maybe something like `remove-external-network`?

I am imaging over time we will be modifying this script to remove external network access for all the internal Function Apps, databases, and storage accounts.

iac/secure-resources.bash
```diff
@@ -0,0 +1,48 @@
...
+
+main () {
+  azure_env=$1
+  source $(dirname "$0")/env/${azure_env}.bash
+  verify_cloud
+
+  resource_group=$2
+  server_name=$3
+
+	echo "Collect all firewall rules on database"
```
ryanhofdotgov, line 25: super nitpick:  4 vs 2 space indent for bash scripts

iac/secure-resources.bash
```diff
@@ -0,0 +1,48 @@
...
+	ids=`az postgres server firewall-rule list \
+		--resource-group $resource_group \
+		--server-name $server_name \
+		--query "[].id" \
+		--output tsv`
+
+	len=`echo $ids | awk '{print length}'`
```
ryanhofdotgov, line 32: nitpick: if there is a way to do this in bash without introducing another scripting language syntax (awk), I'd do that instead. I did verify that `awk` was in Cloud Shell, it's probably also in the Windows for Git environment we require for IaC.

ryanhofdotgov (approved):
Nice work! I've left some nitpicky stuff, but feel free to merge in once it's been successfully run against `dev`.

iac/arm-templates/virtual-network.json
```diff
@@ -0,0 +1,154 @@
...
+                            "type": "object"
+                        },
+                        "subnetPrefix": {
+                            "type": "string",
+                            "defaultValue": "10.0.0.0/24",
+                            "metadata": {
+                                "description": "The address space for the subnet."
+                            }
+                        },
+                        "etlSubnetPrefix": {
```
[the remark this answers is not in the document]
echappen: Yes—now that we only need 2 subnets, we can hardcode the prefixes here. The function apps will be delegated to the `funcSubnetName` throughout `create-resources.bash`. (I'll change the name of this prefix to `func` to match the name)

iac/create-resources.bash
echappen: To clarify - we should only use the prefix for global resources?

ryanhofdotgov: Yeah, that's what I ended up writing up here, but we can revisit as necessary: https://github.com/18F/piipan/blob/main/docs/azure-naming-conventions.md#naming-components

Reasoning is the belief that:
* prefix on a resource with a public DNS entry helps security folks at the top-level agency quickly recognize what agency/service owns the corresponding resource
* some type of unique string will be needed to create unique name in global scope, Microsoft examples show that somewhere in the middle, but we have the prefix so just use that
* if not needed, prefix is line noise that doesn't help anyone

iac/arm-templates/virtual-network.json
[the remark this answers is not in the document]
ryanhofdotgov: Is `matchSubnetPrefix` then unused and should be deleted?

iac/create-resources.bash
echappen: That makes sense, thanks! In that case I'll remove the prefix from other non-global things I added it onto.

iac/arm-templates/virtual-network.json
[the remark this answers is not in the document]
echappen: Ah yep, I'll delete that too.

echappen:
This has been run on `-dev` resource groups and manually tested, so I will merge now.

Issue #660, opened 2021-04-08
Apply naming conventions to IaC


ryanhofdotgov:
When https://github.com/18F/piipan/pull/451 is closed, sweep through the IaC, updating resource naming conventions to match [azure-naming-conventions.md](https://github.com/18F/piipan/blob/main/docs/azure-naming-conventions.md).

Purge and recreate all resource groups for `dev` from scratch when IaC changes complete.

ryanhofdotgov:
At the same time, swap in `tts` for `PREFIX` in `iac/env/tts/dev.bash`.

echappen:
Convo here on how to determine naming scope: https://github.com/18F/piipan/pull/679#issuecomment-818764365

ryanhofdotgov:
@echappen Took a look at https://drive.google.com/file/d/10rdGCeh3gG8F_34GOJ6CG7-nYkaFjqYt/view:
- what does the asterisk on some of the Globals denote?
- shorten `PREFIX-psql-participant-records-ENV` to `PREFIX-psql-participants-ENV`?
- rename `PREFIX-kv-secret-keeper-ENV` to `PREFIX-kv-core-ENV`?
- as we use dash for a name delimeter, I'd change `PREFIX-app-query-tool-ENV` to `PREFIX-app-querytool-ENV`
- likewise for `PREFIX-fd-query-tool-ENV` (and we've already sent off the name to get a cert) 

It doesn't have to happen this round, but I am imagining that the orchestrator Function App will get renamed to something more general as we pile more Functions into it. Likewise, the metrics database will probably cease to just be metrics.

Not sure if this was intending to be complete, but we have also have the APIM instances as well as the managed identities.

echappen:
Forgot to mention: asterisk refers to [this note about web app scope](https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/resource-name-rules#microsoftweb)

I wanted to ask about the participant-records database: given that we may be keeping lookup ID in there too, should we change the name to something more general than `participants`?

The rest of your items look good to me. I'll add APIM and Managed Identities to the doc as well.

echappen:
A markdown file was a bad way to share this on my part. Here's a better [csv of all the names](https://docs.google.com/spreadsheets/d/1Ry4vzX4SfHjOvrCkfy97WnUN9glWWq5wXG0v0PsF8CQ/edit?usp=sharing).

Updated:
- PSQL Database server
- Key Vault
- Query Tool Web App
[...]

PR #679, opened 2021-04-12
update query tool name in deployment

echappen:
Changed the name of query tool app in iac so had to change it in deployment too

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -146,7 +146,7 @@ jobs:
           command: az webapp deployment source config-zip -g rg-core-dev -n fns-app-dashboard-dev --src /build/workspace/dashboard.zip
       - run:
           name: Upload Query Tool
-          command: az webapp deployment source config-zip -g rg-core-dev -n fns-app-query-tool-dev --src /build/workspace/query-tool.zip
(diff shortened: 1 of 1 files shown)


ryanhofdotgov (approved):
👍 but query tool app name is actually a globally scoped resource. No worries – we will fix that up again when we pick up: https://github.com/18F/piipan/issues/660

There's quite a few things to add the prefix and/or environment to, including keyvault, databases, etc. Plus resource types to databases and identities.

echappen:
This may be a good time for me to confirm how to check whether or not a resource is global. I was referencing the Location of the resource—if it's in a region (like West US) vs "Global." Is this the right way to check?

Screen Shot 2021-04-13 at 8 46 19 AM

kingcomma:
I've been relying on the "scope" column in this doc from MS: [Naming rules and restrictions for Azure resources](https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/resource-name-rules)

The naming scope is separate from a resource's location. E.g., two APIM resources (which have a globally scoped naming requirement) in different locations can't share the same name.

echappen:
Perfect, thanks!

PR #680, opened 2021-04-13
Add context to match API README

cjhskippy:
Clarify which APIs are internal-only vs. exposed externally.  Added a little context about the APIs.  Attempted to standardize on current naming conventions.  One possible further improvement would be to fold the links to the orchestrator and lookup ID APIs into /docs/duplicate-participation-api.md

### DIFF
--- a/match/README.md
+++ b/match/README.md
@@ -1,11 +1,18 @@
-# Active Match API
+# Duplicate Participation APIs
 
-*Piipan subsystem for active matching during program participant (re)certification*
+A collection of externally-available and internal APIs that:
+* Allow Piipan to find duplicate participation during participant certification or recertification
+* Allow agencies to find PII of individuals involved in a match using a Lookup ID
 
-## Documentation
+All APIs adhere to [OpenAPI specifications](./docs/openapi.md).
 
+## External APIs
+APIs that are made available to state systems integrating with Piipan make up the [Duplicate Participation API](./docs/duplicate-participation-api.md).
+
+These include:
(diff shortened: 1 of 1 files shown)


ryanhofdotgov (commented):
I think we probably need to discuss a revised, cohesive documentation strategy.

The original idea was internal team focused:
* we have product-wide documentation in https://github.com/18F/piipan/tree/main/docs
* subsytem-specific documentation in every subtree's doc folder
* Every doc folder would have a top-level README with an index into the Markdown (the `Documentation` section). This way you aren't just dumped into a folder of Markdown with terse filenames, but get a more curated Table of Contents.

Also, some of the APIs are essentially internal and external APIs as we call them from the dashboard/query tool, so there isn't always a clean cut division.

There's also a challenge of deciding who the audience of a given doc is – this one's purpose is for the Piipan dev team. The quick start guide that @echappen is working on in #690 definitely for the states. But it is difficult to write a single document that is appropriate for both Piipan devs and states. Even having them in the same repo/site can be confusing.

ryanhofdotgov:
@cjhskippy Does the previous comment make sense? Does the current quick start guide meet the goals you were aiming for?

cjhskippy:
@ryanhofdotgov Yes, thank you.  I think you're right to lean toward axing this change and considering a more comprehensive approach to documentation in the future.  As long as states start with a link to the quick start guide, or generally have no trouble navigating to it, keeping the rest of the docs developer-focused for now sounds right.

Issue #623, opened 2021-04-01
Finalize Quickstart Development Guide for State Distribution


echappen:
We've added a rough draft of the Quickstart guide to the repo and are adding links to it as documentation for each endpoint is developed. Before we share it with states, we'll need to:

- check with our partner about what level of detail we can present about development/production environments
- Consider how we present the match, lookup id, status APIs a single unified API. The fact they are 3 separate Function Apps is an implementation detail that shouldn't drive the documentation I think.
- Have a nice OpenAPI generated set of HTML documentation to link to.
  - Only needs to include query + lookup ID APIs at this time
  - At sprint review, note current status of bulk upload docs in case follow-up work is needed.

echappen:
> Have a nice OpenAPI generated set of HTML documentation to link to

@kingcomma @ryanhofdotgov I'm trying to remember our Eng Sync convo about this:
1. What would ideal access to this site look like? Can it be publicly available as long as it's not interactive?
1. Did we rule out using APIM's own [developer portal](https://docs.microsoft.com/en-us/azure/api-management/api-management-howto-developer-portal) for API documentation?

ryanhofdotgov:
There might be a short term and long term solution here. Short term constraints:
* nothing in our TTS Azure subscription can be exposed without authentication
* Federalist or Markdown in Github (not GitHub Pages) are really the only ATO'd solutions at TTS at our disposal for publishing web pages to the public
* nothing in our partner's subscription can be exposed outside it's agency until it has an ATO

echappen:
Given these constraints, the short-term solution I'm envisioning would be plopping our generated OpenApi yaml files into a [federalist-uswds-jekyll](https://github.com/18F/federalist-uswds-jekyll) repo, styling it kind of [like this](https://robertlove.github.io/jekyll-openapi/) (also from a Jekyll [repo](https://github.com/robertlove/jekyll-openapi)) and hosting it on Federalist. Does that sound like a good path forward?

kingcomma:
I wonder if there is an even shorter-term solution where the specs are rendered to markdown rather than HTML? Something like [swagger-markdown](https://www.npmjs.com/package/swagger-markdown) (haven't looked into this too much).

ryanhofdotgov:
Either option sounds good – goal is just to spend a minimum amount of styling work to make it usable to send out to states sooner rather than later and look like something you'd expect when learning an API for the 1st time (i.e., not just raw YAML).

We can iterate on it as we expand stable API.

[...]

PR #690, opened 2021-04-14
create single openapi spec for apim api; generate markdown docs

echappen:
Addresses 2nd and 3rd items of #623 

- Creates a top-level OpenAPI spec to represent the Duplicate Participation API managed by APIM
- Adds a script to generate markdown documentation for this spec
- Links to this markdown documentation in State Quickstart Guide

Considerations:
- The swagger-markdown tool I originally wanted to use is incompatible with OpenApi 3.0—the generated markdown wasn't showing any info about request parameters. So I found another cli tool called [widdershins](https://github.com/Mermade/widdershins). It's normally used for generating markdown for [Slate](https://github.com/slatedocs/slate) Docs, but produces sufficient markdown on its own. Some refinement's still needed (like around the generated authentication instructions), but this was the prettiest and most flexible MD generator I could find that was compatible with 3.0. 
- Open to thoughts on how to generate these docs automatically for the repo (in CI, in a pre-commit hook, etc).

### DIFF
--- a/docs/openapi/duplicate-participation-api.yaml
+++ b/docs/openapi/duplicate-participation-api.yaml
@@ -12,3 +12,11 @@ paths:
     $ref: '../../match/docs/openapi/orchestrator/query.yaml'
   /lookup_ids/{id}:
     $ref: '../../match/docs/openapi/lookup/id.yaml'
+security:
+  - ApiKeyAuth: []
+components:
+  securitySchemes:
+    ApiKeyAuth:
+      type: apiKey
+      in: header
+      name: Ocp-Apim-Subscription-Key
--- a/docs/openapi/duplicate-participation-api.yaml
+++ b/docs/openapi/duplicate-participation-api.yaml
@@ -1,10 +1,12 @@
 openapi: 3.0.0
 info:
   title: "Duplicate Participation API"
-  version: 0.1.0
+  version: 1.0.0
   description: "The API for the Duplicate Participation system where bulk upload, matching, and lookups will occur"
 tags:
   - name: "Duplicate Participation API"
+servers:
+  - url: "/v1"
 paths:
   /query:
     $ref: '../../match/docs/openapi/orchestrator/query.yaml'
--- a/docs/openapi/duplicate-participation-api.yaml
+++ b/docs/openapi/duplicate-participation-api.yaml
@@ -1,0 +1,12 @@
+openapi: 3.0.0
+info:
+  title: "Duplicate Participation API"
+  version: 0.1.0
+  description: "The API for the Duplicate Participation system where bulk upload, matching, and lookups will occur"
+tags:
+  - name: "Duplicate Participation API"
+paths:
+  /query:
+    $ref: '../../match/docs/openapi/orchestrator/query.yaml'
+  /lookup_ids/{id}:
+    $ref: '../../match/docs/openapi/lookup/id.yaml'
--- a/docs/openapi/duplicate-participation-api.yaml
+++ b/docs/openapi/duplicate-participation-api.yaml
@@ -3,8 +3,6 @@ info:
   title: "Duplicate Participation API"
   version: 1.0.0
   description: "The API for the Duplicate Participation system where bulk upload, matching, and lookups will occur"
-tags:
-  - name: "Duplicate Participation API"
 servers:
   - url: "/v1"
 paths:
--- a/docs/openapi/generated/duplicate-participation-api/openapi.md
+++ b/docs/openapi/generated/duplicate-participation-api/openapi.md
@@ -1,10 +1,14 @@
 <!-- Generator: Widdershins v4.0.1 -->
 
-<h1 id="duplicate-participation-api">Duplicate Participation API v0.1.0</h1>
+<h1 id="duplicate-participation-api">Duplicate Participation API v1.0.0</h1>
 
 > Scroll down for code samples, example requests and responses. Select a language for code samples from the tabs above or the mobile navigation menu.
 
 The API for the Duplicate Participation system where bulk upload, matching, and lookups will occur
+
+Base URLs:
+
+* <a href="/v1">/v1</a>
 
 <h1 id="duplicate-participation-api-match">Match</h1>
 
@@ -14,7 +18,7 @@ The API for the Duplicate Participation system where bulk upload, matching, and 
 
 ```shell
 # You can also use wget
-curl -X POST /query \
+curl -X POST /v1/query \
   -H 'Content-Type: application/json' \
   -H 'Accept: application/json'
 
@@ -104,7 +108,7 @@ This operation does not require authentication
 
 ```shell
 # You can also use wget
-curl -X GET /lookup_ids/{id} \
+curl -X GET /v1/lookup_ids/{id} \
   -H 'Accept: application/json'
 
 ```
--- a/docs/openapi/generated/duplicate-participation-api/openapi.md
+++ b/docs/openapi/generated/duplicate-participation-api/openapi.md
@@ -10,9 +10,16 @@ Base URLs:
 
 * <a href="/v1">/v1</a>
 
+# Authentication
+
+* API Key (ApiKeyAuth)
+    - Parameter Name: **Ocp-Apim-Subscription-Key**, in: header. 
+
 <h1 id="duplicate-participation-api-match">Match</h1>
 
-## post__query
+## Query for Matches
+
+<a id="opIdQuery for Matches"></a>
 
 > Code samples
 
@@ -20,7 +27,8 @@ Base URLs:
 # You can also use wget
 curl -X POST /v1/query \
   -H 'Content-Type: application/json' \
-  -H 'Accept: application/json'
+  -H 'Accept: application/json' \
+  -H 'Ocp-Apim-Subscription-Key: API_KEY'
 
 ```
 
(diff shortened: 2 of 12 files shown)


docs/quick-start-guide-states.md
```diff
@@ -34,16 +34,26 @@ Seperate endpoints and credentials will be provoded for each environment.
...
+| [coming soon] | query for status on data processing from a bulk upload | POST |
+| `/query` | query for active matches | POST |
+| `lookup_ids/:id` | Returns PII for a Lookup ID | GET |
+
+Detailed documentation for each endpoint can be found [here](./openapi/generated/duplicate-participation-api/openapi.md).
 
 ## Authentication
 
-To use all endpoints, states will need two sets of credentials: one for bulk uploading and another for the rest of the endpoints. Contact us for access.
+Users will obtain two API subscription keys that will be rotated on a rolling basis. Keys will be placed into request headers.
```
ryanhofdotgov, line 48:
```suggestion
States will be issued API keys that are placed into request headers to authenticate a web service call. The bulk upload API requires a separate API key from the query and lookup APIs.
```

docs/quick-start-guide-states.md
```diff
@@ -34,16 +34,26 @@ Seperate endpoints and credentials will be provoded for each environment.
...
+
+Example using cURL:
+
+```
+curl --request PUT '<uri>' --header 'Ocp-Apim-Subscription-Key: <api-key>'
+```
+
+Different endpoints may require different sets of keys. Contact us for access to these keys.
```
ryanhofdotgov, line 56:
```suggestion

```

ryanhofdotgov (commented):
Nice find re: Widdershins. 

It might be out of the scope for this specific PR, but I think the `/v1/...` portion of the endpoint needs to be in the generated docs somehow. The generated docs also report that the endpoints don't require authentication.

docs/quick-start-guide-states.md
```diff
@@ -34,16 +34,26 @@ Seperate endpoints and credentials will be provoded for each environment.
...
 ### Endpoints Overview
 
-| endpoint | description | Type | Parameters | Response | Authentication | Instructions |
+| Endpoint | Description | Type |
 |---|---|---|---|---|---|---|
-| `/BulkUpload` | uploads bulk participant data to the system | POST | a CSV file [in this format](../etl/docs/bulk-import.md) | [coming soon] | contact us | [link](./etl/docs/upload.md) |
-| [coming soon] | query for status on data processing from a bulk upload | POST | [coming soon] | [coming soon] | contact us | [coming soon] |
-| `/query` | query for active matches | POST | refer to the [OpenApi Schema](../match/docs/openapi/orchestrator/index.yaml) | refer to the [OpenApi Schema](../match/docs/openapi/orchestrator/index.yaml) | contact us | [coming soon] |
-| `lookup_ids/:id` | Returns PII for a Lookup ID | GET | Lookup ID | [coming soon] | contact us | [coming soon] |
+| `/BulkUpload` | uploads bulk participant data to the system | POST |
```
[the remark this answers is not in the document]
kingcomma: Should the suggested change be `/<state-abbr>/upload/:filename`, to match the format of the `/lookup_ids/:id` endpoint?

Also, thought: I'd love to pick folks brain on a way to make `<state-abbr>` actually the state's two-letter postal abbreviation rather than the name of the storage account resource (as it is now).

docs/openapi/generated/duplicate-participation-api/openapi.md
```diff
@@ -0,0 +1,147 @@
...
+
+<h1 id="duplicate-participation-api">Duplicate Participation API v0.1.0</h1>
+
+> Scroll down for code samples, example requests and responses. Select a language for code samples from the tabs above or the mobile navigation menu.
+
+The API for the Duplicate Participation system where bulk upload, matching, and lookups will occur
+
+<h1 id="duplicate-participation-api-match">Match</h1>
+
+## post__query
```
kingcomma, line 11: *Thought*: looks like we can add [`operationId`](https://swagger.io/docs/specification/paths-and-operations/#operationId)  to the specs to clean these headers up. E.g., `operationId: postQuery`.

docs/openapi/generated/duplicate-participation-api/openapi.md
```diff
@@ -0,0 +1,147 @@
...
+    "dob": "2019-08-24"
+  }
+}
+```
+
+<h3 id="post__query-parameters">Parameters</h3>
+
+|Name|In|Type|Required|Description|
+|---|---|---|---|---|
+|body|body|object|true|none|
```
kingcomma, line 47: *Thought:* Does listing the body seem unnecessary and potentially confusing for developers? I lean that way. Looks like there's an [`--omit-body` option](https://github.com/Mermade/widdershins#options) that would remove this if others agree.

docs/openapi/generated/duplicate-participation-api/openapi.md
```diff
@@ -0,0 +1,147 @@
...
+|»» last|string|true|none|Last name|
+|»» ssn|string|true|none|Social Security number|
+|»» dob|string(date)|true|none|Date of birth|
+|»» state_name|string|false|none|Full state/territory name|
+|»» state_abbr|string|false|none|State/territory two-letter postal abbreviation|
+|»» exception|string|false|none|Placeholder for value indicating special processing instructions|
+
+<aside class="success">
+This operation does not require authentication
+</aside>
```
kingcomma, line 97: *Thought:* since all of these endpoints will require the same form of authentication (API key for now), you can probably do something similar to the `security` and `securitySchemes` sections in the [bulk upload schema](https://github.com/18F/piipan/blob/229b4816b76a720717c63594bd7ae65c6398dbfb/etl/docs/openapi/openapi.yaml#L39) in PR #689

docs/openapi/duplicate-participation-api.yaml
```diff
@@ -0,0 +1,12 @@
+openapi: 3.0.0
+info:
+  title: "Duplicate Participation API"
+  version: 0.1.0
+  description: "The API for the Duplicate Participation system where bulk upload, matching, and lookups will occur"
+tags:
+  - name: "Duplicate Participation API"
```
kingcomma, line 7: *Thought*: for me, this adds a duplicative (ha) "Duplicate Participation API" header when viewed using the Swagger UI (screenshot at bottom). My understanding is the [global tags section](https://swagger.io/docs/specification/grouping-operations-with-tags/) is more for optionally adding more context to the tags used throughout the spec.

Which brings me to another thought — I think it would be helpful to have a conversation around nomenclature at some point so we have a clear language to use when referring to the various API groupings. E.g., should lookup and match be separate or should we have a "match" API with "match query" and "match lookup" operations? Not super relevant at this point, but just thinking ahead for when we start communicating with folks who are new to system.

Screen Shot 2021-04-15 at 9 55 44 AM

kingcomma (commented):
I'm impressed this approach of stitching together the other APIs into one doc is working so well!

I added a few comments/thoughts, mostly around formatting and organization.

docs/openapi/generated/duplicate-participation-api/openapi.md
echappen: Thanks! I was trying to find how we could do that.

docs/openapi/generated/duplicate-participation-api/openapi.md
echappen: I agree we should remove it. I think this is a feature of widdershins that's not working as I expect it to. Ideally it would show the nested body parameters, like it does for the response body. But even when I add the `--expandBody` flag it doesn't do that. 

So atm there's not a way to indicate whether a request parameter is required or not. I think it's fine for now, but it could become confusing in the future.

docs/quick-start-guide-states.md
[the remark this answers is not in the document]
ryanhofdotgov: Good catch. I hadn't realized that <state-abbr> here was actually <storage-account-name>. I've updated it here, but might need to clarify elsewhere.

You might be able to have a simple mapping function (URL rewriting) within the APIM policy? Abbr + standard pattern + dev/test/preprod/prod? Or... since every state gets their own API endpoint for uploads, you could hard-code the account name in the dynamically generated policy, right?

docs/openapi/duplicate-participation-api.yaml
echappen: Good catch—removing the global tag makes it look better in the Swagger Editor and it doesn't change anything in the markdown. I agree we should talk about the groupings at our next sync. 

Screen Shot 2021-04-15 at 10 03 39 AM

echappen:
@kingcomma @ryanhofdotgov Your feedback improved this a lot, thank you! I was able to get rid of those misleading auth messages, remove the weird body table, and clean up the naming of the operations.

docs/quick-start-guide-states.md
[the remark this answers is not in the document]
echappen: If the storage account name is derived from a column in the csv, then the storage name is already the postal code? Or is the csv abbreviation supposed to be different than the postal code?

ryanhofdotgov:
Looks like something isn't quite right with the markdown syntax? https://github.com/18F/piipan/blob/623-api-docs/docs/quick-start-guide-states.md

echappen:
Fixed!

ryanhofdotgov (approved):

PR #700, opened 2021-04-16
Change order of operations for updating packages

kingcomma:
- First restore all projects
- Then update all projects
- Finally force-restore all projects

This update fixes the occasional issues we've seen with `update-packages` missing updates or erroring out. Running three operations one-by-one against all packages avoids issues with out-of-sync lock files and transient dependencies being missed depending on the ordering of `find`.

### DIFF
--- a/tools/update-packages.bash
+++ b/tools/update-packages.bash
@@ -8,6 +8,28 @@
 
 source $(dirname "$0")/common.bash || exit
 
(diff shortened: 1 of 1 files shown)


ryanhofdotgov (approved):
Thanks for digging into this! Maybe we can share with the Dependabot team ;)

PR #709, opened 2021-04-19
remove first name from match query in state match api

### LINKED ISSUE #708: Only match against last name
Partner has repeatedly commented that matching algorithm should be exact match on last name, dob, ssn. 

Looks like relevant line that needs to be changed is in [Api.cs](https://github.com/18F/piipan/blob/b4cfcbde31c718628ec3813c21a3f2b929274dcc/match/src/Piipan.Match.State/Api.cs#L100)

echappen:
Closes #708 

[Manual testing](https://github.com/18F/piipan/blob/main/match/docs/state-match.md#remote-testing) with a different first name (i.e. Teddy instead of Theodore) should return results for all states:
```
curl -X POST -i \
  -H "Content-Type: application/json" \
  -H "Authorization: Bearer ${TOKEN}" \
  -d '{"query": { "first": "Teddy", "last": "Farrington", "dob": "1931-10-13", "ssn": "000-12-3456" }}' \
  https://ofunchaaz7alkqwuvo.azurewebsites.net/api/v1/query\?

### DIFF
--- a/match/src/Piipan.Match.State/Api.cs
+++ b/match/src/Piipan.Match.State/Api.cs
@@ -98,7 +98,7 @@ namespace Piipan.Match.State
                 first = request.Query.First,
             };
             var sql = "SELECT upload_id, first, last, middle, dob, ssn, exception FROM participants " +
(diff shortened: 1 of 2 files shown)


kingcomma (approved):
👍

PR #710, opened 2021-04-19
Exclude cshtml files from code coverage

kingcomma:
- Uses [default code climate exclusion list](https://docs.codeclimate.com/docs/excluding-files-and-folders#auto-generated-file-and-folder-exclusions)
- Adds `.cshtml` files to list

CodeClimate does not seem to offer a method of previewing the impact of changes to the config file. Code coverage metrics should be reviewed if/after merged.

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -20,3 +20,18 @@ checks:
     enabled: false
   identical-code:
     enabled: false
+exclude_patterns:
+- "config/"
+- "db/"
+- "dist/"
+- "features/"
+- "**/node_modules/"
+- "script/"
+- "**/spec/"
+- "**/test/"
+- "**/tests/"
+- "Tests/"
+- "**/vendor/"
+- "**/*_test.go"
+- "**/*.d.ts"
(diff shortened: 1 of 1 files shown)


echappen (approved):

ryanhofdotgov:
I've been burned on a prior assisted acquisition where a vendor had excluded so much of their code from test coverage, leaving me actually 40% coverage after they rolled off the project. 

If the goal is to just understand better where we lack coverage, as reporting options in Code Climate are pretty basic, I might just run a tool locally with the lcov output. Or: Just be really strict about flipping this back afterwards; i.e., next sprint.

kingcomma:
@ryanhofdotgov That makes sense. I'd prefer not to open up a ticket and make sure we flip this back on. For now, I'll just use something like the following to filter locally. If this becomes a pain point for other developers, perhaps we can reassess approaches.

```
dotnet test --collect:"XPlat Code Coverage" -- DataCollectionRunSettings.DataCollectors.DataCollector.Configuration.ExcludeByFile="**/*.cshtml
```

PR #716, opened 2021-04-20
State Guide Refinement Part 2

echappen:
- Updates link to bulk upload docs
- Re-generates docs (which I forgot to do in last PR)
- makes some capitalization tweaks


kingcomma (approved):
👍

PR #717, opened 2021-04-20
Update openapi.md

erinzstrenio:
minor grammatical edits


echappen:
@erinzstrenio This is an auto-generated file, so unfortunately these edits would be overridden with the next generation. We're using A tool called [Widdershins](https://github.com/Mermade/widdershins/tree/master) to generate the markdown for us.

erinzstrenio:
Ah, got it. No worries. None of the edits were major.

PR #721, opened 2021-04-21
Set example env var for query tool when a11y testing in CI

echappen:
Having any value for OrchApiUri will allow the home page to load when navigated to by pa11y. 

Example of a successful CI build with this config can be found [here](https://app.circleci.com/pipelines/github/18F/piipan/1020/workflows/015983ca-1972-4579-b95b-9bce825b24a1/jobs/2285).

It's worth considering as another issue whether we should start having pa11y test things like query results, which would require a useable API uri.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -158,6 +158,8 @@ jobs:
   accessibility_query_tool:
     docker:
       - image: mcr.microsoft.com/dotnet/sdk:3.1
+    environment:
+      OrchApiUri: "http://example.com"
     steps:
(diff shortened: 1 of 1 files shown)


kingcomma (approved):
👍 

*Thought:* we'll probably need to dig into configuration management a bit more at some point. How are bindings like these handled in integration tests, locally, in the partner's CI/CD environment, etc?

echappen:
Agreed! Those pain points will reveal themselves as we get further into integration and a11y testing.

PR #722, opened 2021-04-21
Swap in stage for preprod

ryanhofdotgov:
Use staging/stage to refer to the pre-production environment, rather than `preprod`.


echappen:
Thanks! We should probably update it in the Quick-Start guide environment table as well.

ryanhofdotgov:
@echappen Just saw your last comment after I had already created a separate PR (was just doing quick edits using GitHub web editor, rather than a local checkout)

echappen:
No problemo! Approved 👍

echappen (approved):

PR #725, opened 2021-04-21
Validate lookup ID input

### LINKED ISSUE #724: Validate lookup ID in query tool
User input into the query tool's lookup ID field should be validated upon submission. It should match the limited character set from which IDs are generated and be case-insensitive.

See conversation in https://github.com/18F/piipan/pull/625#issuecomment-824282750.

kingcomma:
Closes #724

### DIFF
--- a/query-tool/src/Piipan.QueryTool/Lookup.cs
+++ b/query-tool/src/Piipan.QueryTool/Lookup.cs
@@ -1,4 +1,3 @@
-using System;
 using System.ComponentModel.DataAnnotations;
 using System.Text.Json.Serialization;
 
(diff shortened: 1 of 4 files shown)


ryanhofdotgov:
Does a left/right white space trim occur before this validation check?

kingcomma:
It does not. It appears when using [model binding](https://docs.microsoft.com/en-us/aspnet/core/mvc/models/model-binding?view=aspnetcore-5.0) and data annotations for validation, there is no mechanism for adjusting the form value before it reaches the validation step. Currently, whitespace on either end would not pass validation and the user would see the error message.

If we want to keep validation exclusive to the server-side for now, it's an option to expand the regular expression to allow for leading and/or trailing whitespace and then trim the value before calling the lookup API. I.e., `"   BBB2345    "` would pass validation but `"  AAA0000 "` would not (as is the case now).

Alternatively, if the goal is just to correct user input, we could trim on the client side.

kingcomma:
That blog entry looks like the ticket. All the options I was finding were similar (custom model bindings) but required a lot of boilerplate and custom validation that seemed like more risk than benefit compared to the alternatives. 

I also did not realize that `\s` would not match non-breaking spaces — thanks for that!

ryanhofdotgov (approved):
Cool 👍, though I wish ASP.NET had this built-in.

kingcomma:
Agreed — I've found several references to folks saying the same, but no references to a built-in option.

I think there's likely some refinement here at some point. The solution is close to the blog post, but had to be adapted based on the source for ASP.NET core's own binders. I imagine some of the boilerplate is unnecessary.

PR #726, opened 2021-04-21
Switch from pre-prod to staging in environments table in Quick Start Guide

echappen:


ryanhofdotgov:
Resovled in https://github.com/18F/piipan/pull/723

PR #737, opened 2021-04-23
Finalize naming of Azure resources in iac

### LINKED ISSUE #660: Apply naming conventions to IaC

echappen:
Closes #660

### DIFF
--- a/ddl/apply-ddl.bash
+++ b/ddl/apply-ddl.bash
@@ -1,9 +1,11 @@
 #!/bin/bash
-# 
+#
 # Creates participant records tables and their access controls for
 # each configured state. PGHOST, PGUSER, PGPASSWORD must be set.
 #
 # usage: apply-ddl.bash
+
+source $(dirname "$0")/../iac/iac-common.bash || exit
 
 set -e
 
@@ -13,6 +15,7 @@ set -u
 : "$PGHOST"
 : "$PGUSER"
 : "$PGPASSWORD"
+: "$ENV"
 
 # Azure user connection string will be of the form:
 # administatorLogin@serverName
@@ -37,8 +40,9 @@ main () {
   while IFS=, read -r abbr name ; do
     db=`echo "$abbr" | tr '[:upper:]' '[:lower:]'`
     owner=$db
-    admin=${db}admin
-    
+    admin=`state_managed_id_name $db $ENV`
+    admin=`echo "$admin" | tr '-' '_'`
+
     echo "Applying DDL to database $db..."
     apply_ddl $db $owner $admin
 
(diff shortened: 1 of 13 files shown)


docs/iac.md
```diff
@@ -25,12 +25,24 @@ To (re)create the Azure resources that `piipan` uses:
 ```
     az login
 ```
+
+**NOTE**: When running `create-resources` for the first time in new resource groups, `create-apim` should be run within `create-resources`, after creating the rest of the Azure resources but before configuring Easy Auth.
```
ryanhofdotgov, line 29: Would you write up a separate issue to follow up on this, and make this more straight-forward?

iac/create-databases.bash
```diff
@@ -183,16 +185,14 @@ main () {
...
+    identity=`state_managed_id_name $db $ENV`
     client_id=`az identity show \
       --resource-group $RESOURCE_GROUP \
       --name $identity \
       --query clientId --output tsv`
 
-    # Database role has the same name as its corresponding AD managed identity
-    role=$identity
+    # Database role takes AD managed identity name and formats it for postgres naming rules
+    role=`echo "$identity" | tr '-' '_'`
```
ryanhofdotgov, line 195: super nitpick: this can be done with [string substitution directly in bash](https://tldp.org/LDP/abs/html/string-manipulation.html#:~:text=Replace%20all%20matches%20of)

ryanhofdotgov (approved):
LGTM - nice work tracking down all these details!

docs/iac.md
echappen: https://github.com/18F/piipan/issues/739

ddl/apply-ddl.bash
```diff
@@ -41,7 +41,7 @@ main () {
     db=`echo "$abbr" | tr '[:upper:]' '[:lower:]'`
     owner=$db
     admin=`state_managed_id_name $db $ENV`
-    admin=`echo "$admin" | tr '-' '_'`
+    admin=`echo ${admin//-/_}`
```
ryanhofdotgov, line 44: You can also drop the echo and subshell, something like this should work:
`admin=${admin//-/_}`

echappen:
This was run on tts/dev and is good to go from my end

ryanhofdotgov (approved):

PR #741, opened 2021-04-26
fix name of query tool app in circle

echappen:


ryanhofdotgov (approved):
We will probably should change the prefix used in our TTS sandbox soon, do you know if there any other places where we hard-code the prefix?

echappen:
Other than in iac/env directories, none that I can find.

PR #744, opened 2021-04-26
Integration tests for per-state match API

### LINKED ISSUE #578: Integration tests for an API

kingcomma:
Add an initial set of integration tests for the per-state API Function App.

- Add new `Piipan.Match.State.IntegrationTests` project
- Use `docker-compose` to run tests (for now — will need to be addressed in other environments)
- Document process for running tests
- Add tests and coverage reporting to CircleCI

I split this out into its own project in an attempt to preserve the ability to run `dotnet test` on `Piipan.Match.State.Tests` without having to go through `docker-compose run ...`. 

Closes #578.


ryanhofdotgov (approved):
Nice! 👍 👍

Issue #762, opened 2021-04-29
Rename per-state bulk APIs in APIM


ryanhofdotgov:
Using the storage account names to form the display name of the per-state bulk upload APIs in APIM makes them hard to distinguish from each other. See if we can change it to something like "<state-abbr> Bulk upload API"

<img width="449" alt="Screen Shot 2021-04-29 at 8 40 25 AM" src="https://user-images.githubusercontent.com/37100189/116579312-1dfcd800-a8c7-11eb-923a-d7b49a0e453e.png">

ryanhofdotgov:
There are 2 name settings: Version Set Name and Settings under `v1`.

Version Set Name affects what is shown in the left pane, top level. Version specific setting controls what is listed in Subscription keys table as well as the drop-down when you assign an API to a Subscription.

<img width="400" alt="Screen Shot 2021-04-29 at 8 57 29 AM" src="https://user-images.githubusercontent.com/37100189/116581673-6cab7180-a8c9-11eb-8c6d-b579c0df47d0.png">

<img width="400" alt="Screen Shot 2021-04-29 at 8 58 31 AM" src="https://user-images.githubusercontent.com/37100189/116581681-6f0dcb80-a8c9-11eb-8427-e27670657af5.png">

ryanhofdotgov:
Duplicate of #764

PR #764, opened 2021-04-29
Update bulk upload API naming and endpoints

### LINKED ISSUE #728: Change bulk upload APIM endpoints to use state abbreviations
When #660 is closed, we will have a consistent approach to naming per-state storage accounts across environments. This should allow the per-state storage account names to be constructed dynamically from within an ARM template given only the state abbreviation. Leverage this to switch the upload endpoints to use the state abbreviation rather than storage account name (eg, `/ea/upload`).

The ARM template should:
- Accept `env` and `prefix` parameters
- Add the storage accounts as dependencies to the `Microsoft.ApiManagement/service/apis` resources

kingcomma:
- Use state abbreviations for upload endpoints
- Dynamically generate storage account names in ARM template
- Update documentation to reflect two APIs

This starts implementing some of the API naming conventions we discussed. After this update, per-state upload endpoints will be in the following format:
```
/bulk/ea/v1/upload/filename.csv
```

This update does not change the endpoints for the `/query` or `/lookup_ids` methods.

I'm tentatively referring to the two APIs as "bulk" (or just "bulk upload" for now) and "duplicate participation" in the documentation.

The updated IaC has been run against the duppart APIM resource in `-dev` and can be viewed in the Portal.

Closes #728, related to #762.

### DIFF
--- a/iac/arm-templates/apim.json
+++ b/iac/arm-templates/apim.json
@@ -2,6 +2,12 @@
     "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#",
     "contentVersion": "1.0.0.0",
     "parameters": {
+        "env": {
+            "type": "String"
+        },
+        "prefix": {
+            "type": "String"
+        },
         "apiName": {
             "type": "String"
         },
@@ -23,7 +29,7 @@
         "dupPartPolicyXml": {
             "type": "String"
         },
-        "uploadAccounts": {
+        "uploadStates": {
             "type": "String"
         },
         "uploadBaseDomain": {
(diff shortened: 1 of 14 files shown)


iac/arm-templates/apim.json
```diff
@@ -40,13 +46,21 @@
...
-        }]
+        "uploadSetName": "bulk",
+        "uploadStatesList": "[split(toLower(parameters('uploadStates')), ' ')]",
+        "copy": [
+            {
+                "name": "uploadApiNames",
+                "count": "[length(variables('uploadStatesList'))]",
+                "input": "[concat(variables('uploadSetName'), '-', variables('uploadStatesList')[copyIndex('uploadApiNames')])]"
+            },
+            // Generate storage account names based on naming convention: {prefix}st{state-abbr}upload{env}
```
ryanhofdotgov, line 57: thought: will something fail in the template if the storage account with this name doesn't exist? Or will there be a run time error?

ryanhofdotgov (approved):

iac/arm-templates/apim.json
kingcomma: Unfortunately, no. I was hoping the storage account could be specified in the `dependsOn` property for the API resource, but you can't reference resources that are not in the same template.

I tried a few other tricks, like using the `resourceId()` function to get an ID for each storage account hoping that would cause a failure if the account did not exist. It did not.

Another approach might be to pass both the list of state abbreviations and the list of storage accounts. But then we'd still want/need to do some validation to make sure the lists are the same length, sorted in the same order, etc.

kingcomma:
@echappen Zero rush, but wanted to make sure you had a chance to review this (especially the docs / quickstart changes) before anything gets merged, if you want.

iac/arm-templates/apim.json
ryanhofdotgov: Yeah, at that point you'd want to pass into the template some type of (JSON) structure and then the template code iterate over it, pull out fields, act on those fields, etc.

I've added this particular annoyance to [this list for FastTrack](https://docs.google.com/document/d/1kGb3KaAyQSm6vPgf_UY0SJicRmTbcf0Q_-dVxhUpqQc/edit).

echappen (approved):
Docs look great to me! Do you think it's ok to remove the ⚠️ Construction warning at the top of the quick start guide now?

PR #776, opened 2021-05-04
Clarify API documentation and update with new bulk upload approach

### LINKED ISSUE #639: Update documentation for match APIs

kingcomma:
Closes #639 for now — will readdress if/when source directories are combined into larger groupings (according to our approach to resource groups).

### DIFF
--- a/etl/docs/api.md
+++ b/etl/docs/api.md
@@ -8,13 +8,32 @@ Implemented as one API for each per-state storage account, all managed in a sing
 
 ## APIM implementation
 
-The IaC creates _N_ per-state API version sets within the APIM instance. Each version set contains an API resource with a single `PUT` operation. The API resource is configured to require an active subscription.
+The [IaC](../../iac/arm-templates/apim.json) creates _N_ per-state APIs within the APIM instance. Each API resource contains a single `PUT` operation. The API resource is configured to require an active subscription.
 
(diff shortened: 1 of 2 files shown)


etl/docs/api.md
```diff
@@ -8,13 +8,32 @@ Implemented as one API for each per-state storage account, all managed in a sing
...
+For example, the endpoint for uploading `example.csv` to state EA in the `tts/dev` environment might be:
+
+```https://tts-apim-duppartapi-dev.azure-api.net/bulk/ea/upload/example.csv```
+
+## Calling the API
+
+To call an endpoint:
+
+1. [Obtain an API key](../../match/docs/duplicate-participation-api.md#managing-api-keys)
+1. Send a [valid request](openapi/openapi.yaml), passing the API key in the `Ocp-Apim-Subscription-Key` header
```
echappen, line 39: [suggestion] Linking this to the markdown docs instead of the yaml docs might make it easier to read.

echappen (approved):

etl/docs/api.md
kingcomma: Yeah, I was torn between readability and limiting the number of times these doc branch out into docs in other directories (though I broke that rule with linking out the API key documentation). I also wasn't sure if we were considering the markdown docs to be strictly for the quickstart guide?

PR #784, opened 2021-05-04
Wait a bit so newly created AAD SP is queryable

### LINKED ISSUE #780: Add sleep before enable_easy_auth in configure-easy-auth.bash
Got this error in USGovermentCloud, re-running configure-easy-auth resolved it:
```
ArgumentParseError: argument --id: expected one argument
configure-easy-auth.bash: error on (or around) line 163
```
Error occurred after `Configuring Easy Auth settings for ...` was printed in log for the particular state, but before `Assigning app role` was printed.

Looks like in `enable_easy_auth`, `az ad sp update` is getting called with an empty string for `app_aad_sp`, which is populated from a search higher up. Theory is that the service principal created previously is not available for query.

Adding 60 second sleep before calling enable_easy_auth should paper over this, similar to other places in the script.

ryanhofdotgov:
Closes #780

### DIFF
--- a/iac/configure-easy-auth.bash
+++ b/iac/configure-easy-auth.bash
@@ -233,6 +233,8 @@ configure_easy_auth_pair () {
   func_app_sp=$(create_aad_app_sp $func $func_app_reg_id)
 
   # Activate App Service Authentication for the Function App API
+  # Wait a bit so we can find the SP in AAD when we search for it
+  sleep 60
   enable_easy_auth $func $group
 
   # Give the client component access to the Function App API


kingcomma (approved):
*thought*: Should we poke this a little more at some point and see if 60 seconds is the best compromise between speed and error avoidance? This script alone is up to 21+ minutes (3 60s sleeps x 7 states).

ryanhofdotgov:
@kingcomma I'm predicting that we're going to be compelled to collapse all those per-state Function Apps into a single Function App due to a deeper security tradeoff of per-state separation and private endpoints – good chance it will devolve back to 3 60s sleeps.

PR #786, opened 2021-05-04
Cloud-aware blob storage resource IDs in APIM

### LINKED ISSUE #769: Confirm storage account resource ID URI in Government Cloud
As of #764 `https://{storageAccountName}.blob.core.windows.net` is used as the resource ID URI for requesting Azure AD tokens for accessing storage accounts. With #600 verify this is also the correct ID in Azure Government Cloud.

Relevant doc: https://docs.microsoft.com/en-us/azure/storage/common/storage-auth-aad-app?tabs=dotnet#azure-storage-resource-id

kingcomma:
Storage resource IDs are different in Azure Commercial vs Azure Government.

Updates resource ID in per-state upload API policies so APIM can successfully retrieve managed identity authentication tokens.

Closes #769

### DIFF
--- a/iac/apim-bulkupload-policy.xml
+++ b/iac/apim-bulkupload-policy.xml
@@ -3,7 +3,7 @@
 <policies>
     <inbound>
         <base />
-        <authentication-managed-identity resource="https://{storageAccountName}.blob.core.windows.net" />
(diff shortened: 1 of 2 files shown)


ryanhofdotgov (approved):
Thanks for the fast fix!

Issue #529, opened 2021-03-18
Automated and consistent build/test/deploy CLI interface for subsystems


ryanhofdotgov:
As a developer, I want a consistent and automated way to build/test/deploy each Piipan function app or app service from the command-line.

Because of unique naming schemes, this used to entail going into the Portal, discovering the name of the Function App or App Service, but with the [event of SysType tags](https://github.com/18F/piipan/blob/main/docs/iac.md#systype-resource-tag), this can possibly be all handled by the deploy script.

Goal would be to eventually use these scripts in CI pipeline and possibly in iac as well, but that's not within the scope of this task. 

Related to #509 

Implementation notes:
- [By .NET convention](https://gist.github.com/davidfowl/ed7564297c61fe9ab814) each project root has a `build.sh`. (It also has a build.cmd for Windows, but we can ignore that for now). We should make this `build.bash` and not have to worry about limiting ourselves to POSIX sh, [as we have done for the IaC scripts] (https://github.com/18F/piipan/blob/main/docs/adr/0010-use-bash-for-scripting.md).
- If no argument is provided, script just builds. Use CLI arguments to indicate test or deploy
- Deploy will need to be passed name of env (e.g., `tts/dev`)
- Handle use cases where subsystems have multiple instances and you want to deploy to all, or just one

Issue #509, opened 2021-03-12
CircleCI config needs refactoring


ryanhofdotgov:
The CircleCI config has grown unwieldy: actions need to be externalized into scripts which are environment aware, and reusable in Team Foundation Server.

These portions of the CircleCI config need to be addressed:
- [ ] Code coverage collection
- [ ] Builds
- [ ] Deploys
- [ ] Tests

PR #791, opened 2021-05-05
Create one script to run all unit tests; use in Circle

echappen:
Part of #529 and #509 

This is a sketch for testing all subsystems at once—wanted to see if I'm on the right track. I was planning on 1 script for each task (build/test/deploy) that does the task for all subsystems at once. But I could also see a scenario where one script does all tasks for one subsystem at a time. 

Integrating a script like this into Circle config helped me figure out what would be required for CI usage, although I realize that's a little out of scope here. 

#### Misc
- Wasn't sure how to elegantly pass flags while also using a `main` function. 
- Also not sure what the deal is with my indentation in `unit-test-all.bash` on Github (shows up as correctly as 2 spaces in my VSCode)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -39,8 +39,6 @@ commands:
 
 jobs:
   test:
-    environment:
-      MSBUILD_PROPS: "-p:ContinuousIntegrationBuild=true"
     docker:
       - image: mcr.microsoft.com/dotnet/sdk:3.1
       - image: postgres:11-alpine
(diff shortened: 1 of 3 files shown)


ryanhofdotgov:
Cool! This is generally on track as far as #509, but for #529 I was thinking that each project would have a build.bash. Kind of like this[ standard .NET project structure](https://gist.github.com/davidfowl/ed7564297c61fe9ab814) that has inspired our own directory structures.

So I could do something like this:
```
$ cd etl
$ build.sh
$ build.sh --test
$ build.sh --deploy tts/test
$ cd ../match
$ build.sh
$ build.sh --test
$ build.sh --deploy tts/test
```
Then in the CircleCI script, you would just iterate over the project dirs and call `build.sh --test` for example. Those build.sh scripts could possibly share script code. But the key is I want to manually able to build/test/deploy using this interface (and not have to remember the dotnet commands), and not just rely on CircleCI.

ryanhofdotgov:
We could also (eventually) extend `--test` to support the unit tests as well as any integration tests with an optional parameter.

ryanhofdotgov:
Quickly reviewing what command-line parsing options mesh with our cross-platform requirements, it seems like the bash built-in `getopts` is the best solution. Not to be confused with the GNU `getopt` which isn't on macOS (not sure about Azure cloud shell or Windows for Git):
https://sookocheff.com/post/bash/parsing-bash-script-arguments-with-shopts/

So perhaps an easier to implement command-line interface would be something like:
```
build.bash [test|deploy] [-e envname]
```

echappen:
@ryanhofdotgov Thanks for the feedback. Deployment for apps like dashboard currently use the [ZIP deployment method](https://github.com/18F/piipan/blob/main/dashboard/docs/dashboard.md#deployment-approach). Should using tools like `zip` be avoided in these build scripts?

ryanhofdotgov:
`zip` is 👍  – the guiding philosophy is that we want as possible, to minimize the number of dependencies we have to install to run our IaC scripts and do a basic deploy. `zip` is in Azure Cloud Shell, macOS, Linux, and Windows, Git for Windows environment and can do what we need without any platform/version-specific zip options.

Deviating from this [list of prerequisites](https://github.com/18F/piipan/blob/main/docs/iac.md#prerequisites) may require us to go get approval for new software at the partner agency. We can do that, but it takes time. It's also super nice to just have things work out of the box with Azure Cloud Shell and not have to install anything.

echappen:
@ryanhofdotgov Given your feedback, [here's an example](https://github.com/18F/piipan/pull/792) of a `build.bash` script for the dashboard subsystem. (I put it in another PR since this PR seems less applicable to #529 but wanted to keep it around independently.)

dashboard/build.bash
```diff
@@ -5,16 +5,18 @@
 #
 # Arguments:
 # [none]                Build project binaries
-# test                  Run tests
+# test [-c]             Run tests
```
ryanhofdotgov, line 8: suggestion: as each script has the exact same command-line arguments, you could move these docs to a comment block above `main` in the shared bash file, and then just point the reader to that function/file in each subsystem script.

PR #792, opened 2021-05-05
Example of a build.bash script for dashboard subsystem

echappen:
Partially addresses #529

- options to run in test-only mode or deployment mode
- deployment mode passes an Azure environment argument

### DIFF
--- a/dashboard/build.bash
+++ b/dashboard/build.bash
@@ -4,15 +4,16 @@
 # Relies on a solutions file (sln) in the subsystem root directory
 #
 # Arguments:
-# build                   Build project binaries
-# test                    Run tests
-# deploy [-e <azure_env>] Deploy to specified Azure Environment (e.g. tts/dev)
+# [none]                Build project binaries
+# test                  Run tests
+# deploy -e <azure_env> Deploy to specified Azure Environment (e.g. tts/dev)
 #
-# Description
-# When deploying, an optional environment flag [-e] can be passed. Defaults to tts/dev.
+# Description:
+# When passed no arguments, script runs in build mode.
+# When deploying, an environment flag [-e] must be passed.
 #
 # Usage:
-# ./build.bash build
+# ./build.bash
 # ./build.bash test
 # ./build.bash deploy
 # ./build.bash deploy -e tts/test
--- a/dashboard/build.bash
+++ b/dashboard/build.bash
@@ -10,14 +10,12 @@
 # ./build.bash
 # ./build.bash -t
 # ./build.bash -d tts/dev
-#
-# Note: -t and -d flags are exclusive. When used together, it will only run in test mode.
 
 source $(dirname "$0")/../tools/common.bash || exit
 
 # set modes
 test_mode="false"
-azure_env=""
+azure_env="none"
 
 while getopts ":td:" arg; do
 	case "${arg}" in
@@ -31,35 +29,41 @@ set_constants () {
   DASHBOARD_APP_NAME=$PREFIX-app-dashboard-$ENV # TODO: make this DRY
 }
 
-main () {
-  test_mode=$1
-  azure_env=$2
+run_tests () {
+  echo "Running tests"
+  dotnet test
+}
+
+run_deploy () {
+  azure_env=$1
   source $(dirname "$0")/../iac/env/${azure_env}.bash
   source $(dirname "$0")/../iac/iac-common.bash
   verify_cloud
-
   set_constants
+  echo "Publishing project"
+  dotnet publish -o ./artifacts
+  echo "Deploying to Azure Environment ${azure_env}"
+  pushd ./artifacts
+    zip -r dashboard.zip .
+  popd
+  az webapp deployment \
+    source config-zip \
+    -g $RESOURCE_GROUP \
+    -n $DASHBOARD_APP_NAME \
+    --src ./artifacts/dashboard.zip
+}
 
-  # run build
+main () {
+  test_mode=$1
+  azure_env=$2
+
   dotnet build
 
   if [ "$test_mode" = "true" ]; then
-    echo "Running tests"
-    dotnet test
-  else
-    echo "Publishing project"
-    dotnet publish -o ./artifacts
-    if [ "$azure_env" != "" ]; then
-      echo "Deploying to Azure Environment ${azure_env}"
-      pushd ./artifacts
-        zip -r dashboard.zip .
-      popd
-      az webapp deployment \
-        source config-zip \
-        -g $RESOURCE_GROUP \
-        -n $DASHBOARD_APP_NAME \
-        --src ./artifacts/dashboard.zip
-    fi
+    run_tests
+  fi
+  if [ "$azure_env" != "none" ]; then
+    run_deploy $azure_env
   fi
 
   script_completed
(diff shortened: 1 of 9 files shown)


dashboard/build.bash
```diff
@@ -0,0 +1,72 @@
...
+# Builds dashboard project
+#
+# Optional Flags:
+# -t        Run tests only
+# -d [env]  Build and deploy to specified Azure Environment (e.g. tts/dev)
+#
+# Usage:
+# ./build.bash
+# ./build.bash -t
+# ./build.bash -d tts/dev
```
ryanhofdotgov, line 12: I'd make the command-line interface more explicit, along the lines of the [.NET CLI tools](https://docs.microsoft.com/en-us/dotnet/core/tools/) or [Mercurial](https://www.mercurial-scm.org/doc/hg.1.html).

Basic pattern is `<executable-name> <command> <options>`. `build.bash test` instead of `build.bash -t` makes it more obvious what is happening. If bash `getopts` supported long form options like `--test`, then choosing between these approaches would be more about style than readability.

In this way, `$1` is expected to be the command, either `build`, `test` or `deploy`, followed by zero or more options. Pop off that off the list of args, then process the remaining array of args using getopts. If there is no `$1` or `$1` starts with `-`, then you do the default of `build` and pass the whole array to getopts.

dashboard/build.bash
```diff
@@ -0,0 +1,72 @@
...
+
+source $(dirname "$0")/../tools/common.bash || exit
+
+# set modes
+test_mode="false"
+azure_env="none"
+
+while getopts ":td:" arg; do
```
ryanhofdotgov, line 20: Move this whole block into `main` or a function?

dashboard/build.bash
```diff
@@ -0,0 +1,72 @@
...
+run_tests () {
+  echo "Running tests"
+  dotnet test
+}
+
+run_deploy () {
+  azure_env=$1
+  source $(dirname "$0")/../iac/env/${azure_env}.bash
+  source $(dirname "$0")/../iac/iac-common.bash
+  verify_cloud
```
ryanhofdotgov, line 41: You want an CLI option error to fail fast, ideally before any amount of work (e.g., a build) is done. This comment may be moot if the `dotnet build` gets removed on deploys.

dashboard/build.bash
```diff
@@ -0,0 +1,72 @@
...
+    -g $RESOURCE_GROUP \
+    -n $DASHBOARD_APP_NAME \
+    --src ./artifacts/dashboard.zip
+}
+
+main () {
+  test_mode=$1
+  azure_env=$2
+
+  dotnet build
```
ryanhofdotgov, line 60: Does doing the build unconditionally help us? i.e., will `dotnet test` rebuild everything, or will it look at timestamps, etc and do the smart thing? Same questions for `dotnet publish`.

dashboard/build.bash
echappen: ok thanks for explaining that, I was wondering how to do long form options with `getopts`...

I'm wondering why we want `build`, `test`, and `deploy` to be exclusive operations? What if I want to do all 3 at once?

dashboard/build.bash
echappen: I was having trouble accessing `getopts` inside a function. How would you pass the options from the command to the `main` function at the bottom of the script?

dashboard/build.bash
ryanhofdotgov: I proposed them as options because we're creating a (text-based) user interface here for humans first (CI/CD pipelines second), so balancing tradeoffs simplicity, consistency, intuitiveness, and implementation complexity – defining useful primitives in an interface that can be built upon.

Similar to how git commit and git push are separate options or dotnet build and dotnet test are separate operations (though sometimes you might end up calling them one right after the other).

dashboard/build.bash
ryanhofdotgov: From my brief scan of the docs, I think I saw getopts looks at the value of `$@` – this will get set on every function call so that it is specific to the function if you read from it within the function.

- at global level, `$@` will be list of args from command-line to the executable script.
- in `main()` or `foo()`, `$@` will be what ever gets passed to it when invoked from within the script.

dashboard/build.bash
echappen: Ah no, I had mistakenly thought it was needed, but it appears to be redundant. Both `dotnet test` and `dotnet publish` will build the binaries.

dashboard/build.bash
echappen: Got it, thanks! How does this summary look?
```
# Builds project with optional testing and app deployment
# Relies on a solutions file (sln) in the subsystem root directory
#
# Synopsis:
# build                   Build project binaries
# test                    Run tests
# deploy [-e <azure_env>] Deploy to specified Azure Environment (e.g. tts/dev)
#
# Description
# When deploying, an optional environment flag [-e] can be passed. Defaults to tts/dev
#
# Usage:
# ./build.bash build
# ./build.bash test
# ./build.bash deploy
# ./build.bash deploy -e tts/test
```

dashboard/build.bash
echappen: ah, thank you!

echappen:
@ryanhofdotgov I completed the rest of the build scripts based on your feedback. To revive an older discussion, I'd like to port the names in `set_constants` into `iac-common` to reduce name duplication between these scripts and iac. Are you ok with using `iac-common` for that purpose?

dashboard/build.bash
ryanhofdotgov: Looks good – thanks! Having a default deployment environment is prone to error though; it's kind of like having a default file path for `rm`. Plus, we don't want any TTS-specific behaviors hard-coded. I'd remove that and just do a check that if `deploy` is specified, that `-e` is also specified, otherwise error and exit.

For my own personal keystroke laziness and subjective style, I'd like to also see the no-argument default be a build, as described in the top comment above, but if I feel strongly enough about it next week I'll just open up a PR for that.

ryanhofdotgov:
> Are you ok with using `iac-common` for that purpose?

No objections. We've punted on developing a rule of thumb on the organization of those. Though I do think this code in particular would benefit from:
1. Adding appropriate SysTags for the metric Function Apps, etc.
1. Using the same SysTag query approach as the ETL and match components to get the identifier/handle to the resource to update.

The SysTag approach doesn't really have a ton of benefit over name lookup when there is only instance of a type in the system. But we haven't quite worked out how we might address the need for load balanced app services/function apps, and using the systag approach rather than finding them by name, might eventually allow us to more easily deploy updates when there are multiple instances of a resource (e.g., `func-xyz-001-dev`, `func-xyz-002-dev`).

ryanhofdotgov:
@echappen Clarifying my comments in the 2 threads above, I think the only thing holding up this PR is removing the `tts/dev` default.

echappen:
@ryanhofdotgov Will do, thank you!

dashboard/build.bash
echappen: > I'd like to also see the no-argument default be a build

Got this into the PR!

ryanhofdotgov (approved):
Looking forward to using this!

PR #806, opened 2021-05-07
Correct docs - match on last name only

cjhskippy:
Update documentation to note that the matches only require last name to be an exact match along with DOB + SSN


echappen (approved):

ryanhofdotgov:
Thinking about this some more, we should clarify that when we describe the matching algorithm, it is not part of the fixed API and is subject to change; i.e., state integrators should not build anything relying on the fact that it is only exact match on those tuples. That's really an internal implementation detail that is subject to change without co-ordination at the API level. I'll open a separate ticket.

PR #815, opened 2021-05-11
Add Shellcheck of full repo to CI pipeline

### LINKED ISSUE #156: Add shellcheck to our CI process
We've got lots of bash shell scripts in our repo, let's apply static analysis to them during our CircleCI test job to flag any potential problems. [Shellcheck](https://github.com/koalaman/shellcheck) seems the de facto choice.

echappen:
closes #156

I tried out running [shellcheck](https://github.com/koalaman/shellcheck) in our repo and adding it as a Circle CI job. 

Running shellcheck on the entire repo produces >100 warnings—mostly linting and shellcheck configuration. These are simple fixes but they'll take some time, maybe a day or two. Considering how heavily we on rely bash scripting, I think this is worth doing. But before I start fixing all the files, I wanted to show an example of what shellcheck catches (`dashboard/build.bash`) to see if we're ok with moving forward.

Some things I discovered:
- In order for shellcheck to source files correctly from scripts, you must specify sourced files in your scripts (`# shellcheck source=./tools/common.bash`) along with running shellcheck with the `-x` option.
- By default, shellcheck [does not have a recursive option](https://github.com/koalaman/shellcheck/wiki/Recursiveness). The command added to `.circleci/config.yml` is the recommended recursive approach.
- At first I tried adding CircleCi's built-in orb job, but quickly realized that a more custom job would be needed, since our scripts are in different directory levels throughout the repo.

### DIFF
--- a/ddl/apply-ddl.bash
+++ b/ddl/apply-ddl.bash
@@ -5,7 +5,8 @@
 #
 # usage: apply-ddl.bash
 
-source $(dirname "$0")/../iac/iac-common.bash || exit
+# shellcheck source=./iac/iac-common.bash
+source "$(dirname "$0")"/../iac/iac-common.bash || exit
 
 set -e
 
@@ -29,22 +30,22 @@ apply_ddl () {
   owner=$2
   admin=$3
 
-  psql $PSQL_OPTS -d $db \
-    -v owner=$owner \
-    -v admin=$admin \
-    -v superuser=$SUPERUSER \
+  psql "$PSQL_OPTS" -d "$db" \
+    -v owner="$owner" \
+    -v admin="$admin" \
+    -v superuser="$SUPERUSER" \
     -f ./per-state.sql
 }
 
 main () {
-  while IFS=, read -r abbr name ; do
-    db=`echo "$abbr" | tr '[:upper:]' '[:lower:]'`
+  while IFS=, read -r abbr ; do
+    db=$(echo "$abbr" | tr '[:upper:]' '[:lower:]')
     owner=$db
-    admin=`state_managed_id_name $db $ENV`
+    admin=$(state_managed_id_name "$db" "$ENV")
     admin=${admin//-/_}
 
     echo "Applying DDL to database $db..."
-    apply_ddl $db $owner $admin
+    apply_ddl "$db" "$owner" "$admin"
 
   done < ../iac/states.csv
 }
--- a/ddl/apply-ddl.bash
+++ b/ddl/apply-ddl.bash
@@ -23,14 +23,14 @@ set -u
 SUPERUSER=${PGUSER%@*}
 
 export PGOPTIONS='--client-min-messages=warning'
-PSQL_OPTS='-v ON_ERROR_STOP=1 -X -q'
+PSQL_OPTS=(-v ON_ERROR_STOP=1 -X -q)
 
 apply_ddl () {
   db=$1
   owner=$2
   admin=$3
 
-  psql "$PSQL_OPTS" -d "$db" \
+  psql "${PSQL_OPTS[@]}" -d "$db" \
     -v owner="$owner" \
     -v admin="$admin" \
     -v superuser="$SUPERUSER" \
@@ -38,7 +38,7 @@ apply_ddl () {
 }
 
 main () {
-  while IFS=, read -r abbr ; do
+  while IFS=, read -r abbr _; do
     db=$(echo "$abbr" | tr '[:upper:]' '[:lower:]')
     owner=$db
     admin=$(state_managed_id_name "$db" "$ENV")
--- a/iac/create-apim.bash
+++ b/iac/create-apim.bash
@@ -106,8 +106,9 @@ main () {
 
   upload_policy_path=$(dirname "$0")/apim-bulkupload-policy.xml
   upload_policy_xml=$(< "$upload_policy_path")
-  # shellcheck disable=SC2178
-  state_abbrs="$(get_state_abbrs)"
+
+  local state_abbrs
+  state_abbrs=$(get_state_abbrs)
 
   apim_identity=$(\
     az deployment group create \
@@ -125,7 +126,7 @@ main () {
         publisherName="$PUBLISHER_NAME" \
         orchestratorUrl="$orch_api_url" \
         dupPartPolicyXml="$duppart_policy_xml" \
-        uploadStates="${state_abbrs[*]}" \
+        uploadStates="$state_abbrs" \
         uploadPolicyXml="$upload_policy_xml" \
         location="$LOCATION" \
         resourceTags="$RESOURCE_TAGS")
--- a/iac/create-apim.bash
+++ b/iac/create-apim.bash
@@ -15,7 +15,8 @@
 #
 # usage: create-apim.bash <azure-env> <admin-email>
 
-source $(dirname "$0")/../tools/common.bash || exit
+# shellcheck source=./tools/common.bash
+source "$(dirname "$0")"/../tools/common.bash || exit
 
 clean_defaults () {
   local group=$1
@@ -24,8 +25,8 @@ clean_defaults () {
   # Delete "echo API" example API
   az apim api delete \
     --api-id echo-api \
-    -g ${group} \
-    -n ${apim} \
+    -g "${group}" \
+    -n "${apim}" \
     -y
 
   # Delete default "Starter" and "Unlimited" products and their associated
@@ -33,15 +34,15 @@ clean_defaults () {
   az apim product delete \
     --product-id starter \
     --delete-subscriptions true \
-    -g ${group} \
-    -n ${apim} \
+    -g "${group}" \
+    -n "${apim}" \
     -y
 
   az apim product delete \
     --product-id unlimited \
     --delete-subscriptions true \
-    -g ${group} \
-    -n ${apim} \
+    -g "${group}" \
+    -n "${apim}" \
     -y
 }
 
@@ -51,11 +52,11 @@ generate_policy () {
   local uri=$2
   local APP_URI_PLACEHOLDER="{applicationUri}"
   local xml
-  xml=$(< $path)
+  xml=$(< "$path")
 
   xml=${xml/$APP_URI_PLACEHOLDER/$uri}
 
-  echo $xml
+  echo "$xml"
 }
 
 grant_blob () {
@@ -65,14 +66,14 @@ grant_blob () {
 
   az role assignment create \
     --role "Storage Blob Data Contributor" \
-    --assignee $assignee \
+    --assignee "$assignee" \
     --scope "${DEFAULT_PROVIDERS}/Microsoft.Storage/storageAccounts/${storage_account}"
 }
 
 get_state_abbrs () {
   local state_abbrs=()
 
-  while IFS=, read -r abbr name ; do
+  while IFS=, read -r abbr _; do
     abbr=$(echo "$abbr" | tr '[:upper:]' '[:lower:]')
     state_abbrs+=("${abbr}")
   done < states.csv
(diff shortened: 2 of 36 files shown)


ryanhofdotgov (dismissed):
Approach looks good. I'd probably do it in 2 phases or at least 2 separate commits to make review easier:
* first, address all the double-quoting recommendations and purging all the backtick style for sub-commands. 
* look at what is left after the noise is gone, fix those. This is where we will need to look more closely, make sure nothing breaks.

ryanhofdotgov:
Also: As you are making changes to be compliant with Shellcheck, I'd avoid doing any refactoring or major reformatting – just limit it to the minimal number of changes for easy inspection. Then maybe I'd execute the whole IaC on `-foo` or something to make sure nothing broke in the process.

ddl/apply-ddl.bash
```diff
@@ -29,22 +30,22 @@ apply_ddl () {
   owner=$2
   admin=$3
 
-  psql $PSQL_OPTS -d $db \
-    -v owner=$owner \
-    -v admin=$admin \
-    -v superuser=$SUPERUSER \
+  psql "$PSQL_OPTS" -d "$db" \
```
ryanhofdotgov, line 33: I'm not sure double-quoting PSQL_OPTS is going to result in intended result. In this case, we want the shell to expand it so all the options get passed. Can you confirm?

echappen: Alternatively I could [make the options an array](https://github.com/koalaman/shellcheck/wiki/SC2086#exceptions) and split them:
```
PSQL_OPTS=(-v ON_ERROR_STOP=1 -X -q)
psql "${PSQL_OPTS[@]}" -d "$db" \
 ...
```

ryanhofdotgov: Sure – that seems simple enough.

iac/create-metrics-resources.bash
```diff
@@ -41,73 +42,76 @@ set_constants () {
...
+    --name "$DB_SERVER_NAME" \
     --resource-type "Microsoft.DbForPostgreSQL/servers" \
-    --query properties.fullyQualifiedDomainName -o tsv`
+    --query properties.fullyQualifiedDomainName -o tsv)
+  export PGHOST
   export PGPASSWORD=$PG_SECRET
 
   echo "Insert db table"
-  psql -U $DB_ADMIN_NAME@$DB_SERVER_NAME -p 5432 -d $DB_NAME -w -v ON_ERROR_STOP=1 -X -q - <<EOF
+  psql -U $DB_ADMIN_NAME@"$DB_SERVER_NAME" -p 5432 -d $DB_NAME -w -v ON_ERROR_STOP=1 -X -q - <<EOF
```
ryanhofdotgov, line 114: Not that any of this really needs double-quotes, but if we are adding them, it should be around the complete argument to `-U`:
```
psql -U "$DB_ADMIN_NAME@$DB_SERVER_NAME"
```
Shellcheck will likely complain about `$DB_NAME` too.

echappen: Thanks! There were a couple instances where shellcheck would recommend a fix and I thought, do I just do exactly what shellcheck's telling me, or do what I think is actually accurate? But yeah, this is one of those that can be improved now that I'm more comfortable with the tool.

echappen:
@ryanhofdotgov This is now ready for a final review. I ran iac on `tts/foo`, manually tested etl, match, and metrics apis (using their own supporting dev scripts), and ran all subsystem build script variations.

iac/create-apim.bash
```diff
@@ -83,60 +84,61 @@ get_state_abbrs () {
...
 
-  duppart_policy_xml=$(generate_policy apim-duppart-policy.xml ${orch_base_url})
+  duppart_policy_xml=$(generate_policy apim-duppart-policy.xml "${orch_base_url}")
 
   upload_policy_path=$(dirname "$0")/apim-bulkupload-policy.xml
-  upload_policy_xml=$(< $upload_policy_path)
-  state_abbrs=$(get_state_abbrs)
+  upload_policy_xml=$(< "$upload_policy_path")
+  # shellcheck disable=SC2178
+  state_abbrs="$(get_state_abbrs)"
```
ryanhofdotgov, line 110: This one is a tricky case. What's happening is that we have a variable called `state_abbrs` in the global scope as well as a local one in the `get_abbrs` function. So shell check is getting confused.

I would:
1. remove the shellcheck disable directive
1. Do `local state_abbrs` on line above `state_abbrs="$(get_state_abbrs)"`
1. restore `state_abbrs="$(get_state_abbrs)"` to `state_abbrs=$(get_state_abbrs)` as double-quotes there should be unnecessary and not flagged by shellcheck
1.  restore `uploadStates="$state_abbrs"` where it is used below

iac/create-resources.bash
```diff
@@ -36,17 +37,11 @@ set_constants () {
   # Base name of lookup API storage account
   LOOKUP_STORAGE_NAME=${PREFIX}stlookupapi${ENV}
 
-  # Display name of service principal account responsible for CI/CD tasks
```
ryanhofdotgov, line 39: Just double-checking: this was all redundant?

iac/create-resources.bash
```diff
@@ -85,137 +80,143 @@ az_connection_string () {
 main () {
   # Load agency/subscription/deployment-specific settings
   azure_env=$1
-  source $(dirname "$0")/env/${azure_env}.bash
-  source $(dirname "$0")/iac-common.bash
+  # shellcheck source=./iac/env/tts/dev.bash
```
ryanhofdotgov, line 83: Can you say more about this shellcheck directive? Here it is pointing directly to dev.bash, but code is parameterized to azure_env.

iac/create-service-principal.bash
```diff
@@ -15,13 +15,16 @@
...
-source $(dirname "$0")/../tools/common.bash || exit
+# shellcheck source=./tools/common.bash
+source "$(dirname "$0")"/../tools/common.bash || exit
 
 main () {
   # Load agency/subscription/deployment-specific settings
   azure_env=$1
-  source $(dirname "$0")/env/${azure_env}.bash
-  source $(dirname "$0")/iac-common.bash
+  # shellcheck source=./iac/env/tts/dev.bash
```
ryanhofdotgov, line 24: Same as previous comment.

iac/create-state-storage-service-principal.bash
```diff
@@ -13,14 +13,17 @@
...
+# shellcheck source=./tools/common.bash
+source "$(dirname "$0")"/../tools/common.bash || exit
 
 main () {
   # Load agency/subscription/deployment-specific settings
   azure_env=$1
   storage_acct_name=$2
-  source $(dirname "$0")/env/${azure_env}.bash
-  source $(dirname "$0")/iac-common.bash
+  # shellcheck source=./iac/env/tts/dev.bash
```
ryanhofdotgov, line 23: Same as previous comment. One thing to keep in mind is that `tts/dev.bash` may not exist in all repos.

iac/delete-resources.bash
```diff
@@ -5,35 +5,38 @@
...
     --mode Complete \
     --query doesNotExist
 }
 
 main () {
   # Load agency/subscription/deployment-specific settings
   azure_env=$1
-  source $(dirname "$0")/env/${azure_env}.bash
-  source $(dirname "$0")/iac-common.bash
+  # shellcheck source=./iac/env/tts/dev.bash
```
ryanhofdotgov, line 26: Ditto

iac/remove-external-network.bash
```diff
@@ -11,36 +11,37 @@
...
 
-source $(dirname "$0")/../tools/common.bash || exit
+# shellcheck source=./tools/common.bash
+source "$(dirname "$0")"/../tools/common.bash || exit
 
 main () {
   azure_env=$1
-  source $(dirname "$0")/env/${azure_env}.bash
-  source $(dirname "$0")/iac-common.bash
+  source "$(dirname "$0")"/env/"${azure_env}".bash
```
ryanhofdotgov, line 19: This script doesn't have the shell check source directive like the others?

match/build.bash
```diff
@@ -4,34 +4,38 @@
...
+source "$(dirname "$0")"/../tools/common.bash || exit
+# shellcheck source=./tools/build-common.bash
+source "$(dirname "$0")"/../tools/build-common.bash || exit
 
 run_deploy () {
   azure_env=$1
-  source $(dirname "$0")/../iac/env/${azure_env}.bash
-  source $(dirname "$0")/../iac/iac-common.bash
+  # shellcheck source=./iac/env/tts/dev.bash
+  source "$(dirname "$0")"/../iac/env/"${azure_env}".bash
```
ryanhofdotgov, line 15: Ditto

match/tools/test-lookup-api.bash
```diff
@@ -12,52 +12,56 @@
...
+# shellcheck source=./iac/iac-common.bash
+source "$(dirname "$0")"/../../iac/iac-common.bash || exit
 
 LOOKUP_API_FUNC_NAME="lookup_ids"
 
 main () {
   # Load agency/subscription/deployment-specific settings
   azure_env=$1
-  source $(dirname "$0")/../../iac/env/${azure_env}.bash
+  # shellcheck source=./iac/env/tts/dev.bash
```
ryanhofdotgov, line 25: Ditto

match/tools/test-lookup-api.bash
```diff
@@ -12,52 +12,56 @@
...
-      -g $MATCH_RESOURCE_GROUP \
-      -n $name \
-      --function-name $LOOKUP_API_FUNC_NAME \
+      -g "$MATCH_RESOURCE_GROUP" \
+      -n "$name" \
+      --function-name "$LOOKUP_API_FUNC_NAME" \
       --query invokeUrlTemplate \
       -o tsv)
-  endpoint_uri=$(echo $endpoint_uri | sed "s/{lookupid}/$lookup_id/")
+  # shellcheck disable=SC2001
```
ryanhofdotgov, line 55: Is there something that would prevent using bash string substitution per the recommendation in SC2001, instead of disabling shellcheck?

tools/authorize-cli.bash
```diff
@@ -12,13 +12,16 @@
...
-source $(dirname "$0")/common.bash || exit
+# shellcheck source=./tools/common.bash
+source "$(dirname "$0")"/common.bash || exit
 
 main () {
   # Load agency/subscription/deployment-specific settings
   azure_env=$1
-  source $(dirname "$0")/../iac/env/${azure_env}.bash
-  source $(dirname "$0")/../iac/iac-common.bash
+  # shellcheck source=./iac/env/tts/dev.bash
```
ryanhofdotgov, line 21: Ditto

tools/authorize-cli.bash
```diff
@@ -27,12 +30,12 @@ main () {
...
   # - https://github.com/Azure/azure-cli/blob/24e0b9ef8716e16b9e38c9bb123a734a6cf550eb/src/azure-cli-core/azure/cli/core/_profile.py#L65
   CLI_ID="04b07795-8ddb-461a-bbee-02f9e1bf7b46"
-
   object_id=$(\
     az ad app show \
-      --id $app_uri \
+      --id "$app_uri" \
       --query objectId \
       -o tsv)
+  # shellcheck disable=SC2016
```
ryanhofdotgov, line 38: Not quite following why shellcheck is getting tripped up by this? Do you have any insights here? Is it the JMESPath being passed to `--query`?

tools/build-common.bash
```diff
@@ -80,6 +80,10 @@ main () {
         c )
           ci_mode='true'
           ;;
+        * )
```
ryanhofdotgov, line 83: This change is out of scope for this issue, but would like to see implementation parity for the deploy command.

ryanhofdotgov (commented):
Whew – that's a lot of cleanup done – thanks! A few comments, and then after they are resolved I'd like to do one last review sweep.

iac/create-resources.bash
echappen: Yep, shellcheck flagged it as unused and I couldn't find it used in the codebase.

iac/create-resources.bash
echappen: My understanding is shellcheck sources the files you specify here as part of its code evaluation. I sourced all files from the project root. 

In this case with `$azure_env` being part of the filepath, shellcheck just needs an example file to source from, or else you'll get errors like `unbound variable` when using `$ENV` in the main file. For shellcheck's purposes, it doesn't matter which environment file it is.

match/tools/test-lookup-api.bash
echappen: I got tripped up on needing to escape the curly brackets in the first pass. Got it working!

tools/authorize-cli.bash
echappen: shellcheck kept wanting to evaluate `user_impersonation` as an expression instead of a string. Somehow couldn't get out of error whack-a-mole with this one.

tools/build-common.bash
echappen: Without it, shellcheck flagged this as `SC2220: Invalid flags are not handled. Add a *) case.`

echappen: Want me to disable it and add it back in another PR?

iac/create-state-storage-service-principal.bash
echappen: Good point, I'll look into how to make this work across repos

iac/create-resources.bash
ryanhofdotgov: OK. I understand better. Perhaps in the future we will have an explicit environment template file, which could be used instead of referencing `tts/dev`, but this is good.

iac/create-state-storage-service-principal.bash
ryanhofdotgov: See previous template comment, we can punt on that.

tools/authorize-cli.bash
ryanhofdotgov: Gotcha – yeah, this is an unusual case where we have backticks that are almost always used for sub-shell commands in bash, not JMESPath expressions.

tools/build-common.bash
ryanhofdotgov: No – it's cool here. Was just saying that would like the other option code immediately above this block to be structured the same (i.e., no reason for it to look different), and _that_ was out of scope for this issue.

echappen: noted, thanks!

iac/create-apim.bash
echappen: done, thanks!

echappen:
@ryanhofdotgov Alrighty, I got through all the outstanding items! Thanks for the thorough review.

ryanhofdotgov (approved):
👍 👍 

Not a blocker, but there were a few scripts that didn't consistently have a `shellcheck source=....` directive above their `source` line.  Any ideas why?

- upload.bash
- remove-external-network.bash
- add-front-door-to-app.bash 
- create-apim.bash

Also, I just realized the shellcheck is smart enough not to suggest double quotes around true constants that don't have spaces in their value; e.g,. `DB_NAME`.

echappen:
> there were a few scripts that didn't consistently have a shellcheck source=.... directive above their source line. Any ideas why?

Ah, I think it's because those scripts don't need to use those source files anymore.

ryanhofdotgov:
> Ah, I think it's because those scripts don't need to use those source files anymore.

Hmm... `verify_cloud` should definitely call on stuff there.

echappen:
True!

PR #856, opened 2021-05-14
Bump hosted-git-info dependency

kingcomma:
Should resolve the [vulnerability](https://github.com/18F/piipan/security/dependabot/dashboard/src/Piipan.Dashboard/package-lock.json/hosted-git-info/open) flagged by dependabot. In this case, `npm update` seemed to do the trick. Also updated `query-tool` which had the same vulnerability flagged by `npm audit` but not dependabot.


ryanhofdotgov (approved):
Any insights as to why dependabot couldn't produce the PR for this itself?

kingcomma:
My dependabot knowledge is lacking — do we need to explicitly list javascript as a package manager / ecosystem in `dependabot.yml`?

PR #857, opened 2021-05-14
use dynamic sql to set query parameters

echappen:


ryanhofdotgov (approved):
Thanks!

- Those particular tests that you had to modify – at some point would be good to revisit them and make them integration tests using the recent approach @kingcomma recently incorporated on the matching side – testing against the particular rows returned (for example) rather than the qualities of the query string.

- I just realized that the way I constructed parameterized SQL statements in the ETL code was not a good general example, as I was experimenting to see what it would take to get 90+% test coverage without integration tests.

If you look at the original version of that code, you'll see a simpler API call that is built into the command class – let's you skip all the verbose explicit typing:
https://github.com/18F/piipan/blame/ff408848a071d088302ee17f6908d5e2edc1592c/etl/src/Piipan.Etl/BulkUpload.cs#L113

echappen:
Refactoring this is taking longer than expected, so I'm going to merge this and create a new issue for integration tests and the simpler sql connection.

PR #858, opened 2021-05-14
Explicitly add npm to dependabot config for apps

ryanhofdotgov:
@kingcomma Great idea about dependabot config. Let's see if this will do it in future.


kingcomma (approved):
👍

Issue #864, opened 2021-05-17
Integration Tests for Metrics API


echappen:
Refactors from PR #857:

At some point would be good to revisit them and make them integration tests using the recent approach @kingcomma recently incorporated on the matching side – testing against the particular rows returned (for example) rather than the qualities of the query string.

I just realized that the way I constructed parameterized SQL statements in the ETL code was not a good general example, as I was experimenting to see what it would take to get 90+% test coverage without integration tests.

If you look at the original version of that code, you'll see a simpler API call that is built into the command class – let's you skip all the verbose explicit typing:
https://github.com/18F/piipan/blame/ff408848a071d088302ee17f6908d5e2edc1592c/etl/src/Piipan.Etl/BulkUpload.cs#L113

PR #872, opened 2021-05-18
Start integration tests for metrics API; Make Metrics constants DRY

echappen:
Partially addresses #864 

Since Metrics names are now being used in a lot of places, I hoisted them up to iac-common. This has implications for iac scripts.

### DIFF
--- a/iac/create-metrics-resources.bash
+++ b/iac/create-metrics-resources.bash
@@ -32,14 +32,10 @@ set_constants () {
   DASHBOARD_WAF_NAME=wafdashboard${ENV}
   # Metrics Collection Info
   COLLECT_APP_FILEPATH=Piipan.Metrics.Collect
-  COLLECT_APP_ID=metricscol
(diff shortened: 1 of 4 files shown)


ryanhofdotgov (approved):
LGTM, a couple of thoughts:

- didn't see shellcheck results in PR, is that expected? I think there is a couple of places where it would flag needing double quotes.
- I'd leave the smoke test shell script as is for now, and not go deeper with jq. Instead deferring those kinds of tests to a .NET integration test like in PR #744

echappen:
@ryanhofdotgov The shellcheck CI job passed, and I ran it manually on the new file with no errors. I'm still figuring out when shellcheck flags quotes and when it doesn't. Like you saw recently, I think it's trying to do something smart there.

ryanhofdotgov:
Ah... I was not thinking and was expecting a Shellcheck line item in the PR "8 checks passed" section. But I see it now in the details of build-and-deploy CircleCI job.

Issue #866, opened 2021-05-18
Add case number and participant ID to Query API


cjhskippy:
Provide additional information on participants to the Query API responses.  No changes to the query request are needed.

AC:
- [ ] Query API returns participant ID, if one is available
- [ ] Query API returns case identifier
- [ ] Query Tool displays case ID and participant ID in query response table

cjhskippy:
@echappen Correct.  I don't anticipate the need to query by case ID or participant ID at any time before MVP launch - maybe not ever.

PR #887, opened 2021-05-19
Add case id and participant id to match query response and query tool response table

echappen:
Closes Issue #866

Screen Shot 2021-05-20 at 1 51 52 PM

### DIFF
--- a/query-tool/src/Piipan.QueryTool/Pages/Shared/_ResultsTable.cshtml
+++ b/query-tool/src/Piipan.QueryTool/Pages/Shared/_ResultsTable.cshtml
@@ -10,6 +10,8 @@
             <th scope="col">DOB</th>
             <th scope="col">SSN</th>
             <th scope="col">State</th>
+            <th scope="col">Case Id</th>
+            <th scope="col">Participant Id</th>
         </tr>
     </thead>
     <tbody>
@@ -20,6 +22,8 @@
                     <td>@Html.DisplayFor(m => record.DateOfBirth)</td>
                     <td>@record.SocialSecurityNum</td>
                     <td>@record.StateName</td>
+                    <td>@record.CaseId</td>
+                    <td>@record.ParticipantId</td>
                 </tr>
             }
         </tbody>
--- a/docs/openapi/generated/duplicate-participation-api/openapi.md
+++ b/docs/openapi/generated/duplicate-participation-api/openapi.md
@@ -68,7 +68,9 @@ Queries all state databases for any PII records that are an exact match to the l
       "dob": "2019-08-24",
       "state_name": "string",
       "state_abbr": "string",
-      "exception": "string"
+      "exception": "string",
+      "case_id": "string",
+      "participant_id": "string"
     }
   ]
 }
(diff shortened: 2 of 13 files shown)


echappen:
@kingcomma @timoballard I'm very close to being ready with this PR. I've been stuck on this [This one failing test](https://app.circleci.com/pipelines/github/18F/piipan/1287/workflows/5027c00a-6fd8-4b7a-9a6a-72a2b5634eeb/jobs/2967). The test verifies that the record does have a CaseID ("CaseIdExample"), but the response match does not. I'm not sure where exactly the breakdown is happening. Any insight you could give would be helpful!

timoballard:
@echappen - is this possibly due to there not being a call to `ClearParticipants` at the top of the `ApiReturnsMatches` test? Maybe there's a lingering row from the previous test run (from before `CaseID` and `ParticipantId` were added)

kingcomma:
@echappen My suspicion is the same as yours. The per-state API uses [Dapper](https://github.com/DapperLib/Dapper) as an ORM of sorts for querying the database and mapping results to `PiiRecord` objects. So, I think the crux may be with Dapper not Npsql, but I haven't had the time to dig in too deep yet.

echappen:
@kingcomma Thanks! Didn't realize we were using Dapper for that so I'll check that out.

kingcomma:
👍  Looks like there's [some advice here](https://stackoverflow.com/questions/8902674/manually-map-column-names-with-class-properties). I just tested simply specifying aliases in camelcase and it seemed to clear up the issue. Not sure if this is the best approach, but at least solves the problem:

```
SELECT upload_id, first, last, middle, dob, ssn, exception, case_id CaseId, participant_id ParticipantId FROM participants ...
```

echappen:
Ok, this is ready for review at this point. Thank you all for your debugging help.

query-tool/src/Piipan.QueryTool/Pages/Shared/_ResultsTable.cshtml
```diff
@@ -10,6 +10,8 @@
             <th scope="col">DOB</th>
             <th scope="col">SSN</th>
             <th scope="col">State</th>
+            <th scope="col">Case Id</th>
+            <th scope="col">Participant Id</th>
```
kingcomma, line 14: *Tiny nit:* "Id" should probably be "ID" to match capitalization on the other headers (e.g., "SSN", "DOB").

kingcomma (approved):
👍 LGTM! Just one tiny nit that shouldn't hold up merging (we have UI polishing slated for the upcoming sprint anyway).

*Thought:* With other API changes coming down the pike, wanted to point out [this answer](https://stackoverflow.com/a/34536863) on the Dapper/naming issue. Flipping that setting should remove the need to add Dapper-specific aliases into the SQL statements. May be a more efficient approach if we find ourselves having to hardcode more aliases.

(Also, sorry for the delay on review — was hoping to wait until Code Climate checks were running again. Since that is blocked for now, don't want it to hold up any work.)

query-tool/src/Piipan.QueryTool/Pages/Shared/_ResultsTable.cshtml
echappen: Thanks, I'll change that in this next PR where I'm working in that same file

echappen:
> With other API changes coming down the pike, wanted to point out this answer on the Dapper/naming issue. Flipping that setting should remove the need to add Dapper-specific aliases into the SQL statements. May be a more efficient approach if we find ourselves having to hardcode more aliases.

If we're ok with being limited to that naming convention (db column name underscores = C# Pascal case) then I'd be 👍  with this option.

PR #905, opened 2021-05-21
Add benefits end date to participants database and bulk upload

### LINKED ISSUE #860: Track final benefits date when assigned
As a State agency worker, I want to see final date for which a participant received benefits when viewing a match so that I can avoid contacting another state or the applicant when the rules allow it.

If an applicant has closed their case in another state but their last benefits date has not yet passed, the eligibility worker in the initiating state should be able to pro-rate benefits to begin after that terminal date without having to perform further review.  Similarly, the worker should also be able to avoid taking action on a disclosure by the participant that they were previously receiving benefits in that state.

AC:
- [ ] Bulk upload process allows optional inclusion of final benefit date
- [ ] Bulk upload accepts records without last benefit date
- [ ] ~~Query API returns the final benefit date in the matched state(s) when provided~~

echappen:
Closes #860 
Since this changes the database schema, I'll drop participants database server and re-run iac on tts/dev before merging this. 

## Notes
Per [this discussion](https://github.com/18F/piipan/issues/860), I made the following choices:
- Database column name and C# object property are named `benefits_end_date` since it's technically a postgresql `date` column and a `DateTime` C# object. 
- However in the bulk upload csv, the field is named "benefits end month" since we're assuming the month is all states will be tracking. The import schema has a data type of [yearmonth](https://specs.frictionlessdata.io/table-schema/#types-and-formats) that seemed most appropriate. 
- Our system takes on the responsibility of determining the final day of benefits based on the provided month. I'm not sure when exactly this transfer from month-only to last-day-of-month should happen, but I think it should happen before it's recorded in the database. 
- ~I didn't convert the new csv field to ISO 8601 (per [this discussion](https://gsa-tts.slack.com/archives/C018P1ZJPEC/p1621541101091800?thread_ts=1621537600.087600&cid=C018P1ZJPEC)). figured we'd change all date fields at once. But if we want to do it here I can.~ See discussion below

### DIFF
--- a/etl/docs/csv/import-schema.json
+++ b/etl/docs/csv/import-schema.json
@@ -62,7 +62,6 @@
       {
         "name": "benefits end month",
         "type": "yearmonth",
-        "format": "%m/%Y",
         "description": "month and year when participant's benefits end"
       }
     ]
--- a/etl/docs/csv/import-schema.json
+++ b/etl/docs/csv/import-schema.json
@@ -22,7 +22,7 @@
       {
         "name": "dob",
         "type": "date",
-        "format": "%m/%d/%Y", 
+        "format": "%m/%d/%Y",
         "description": "Participant's date of birth",
         "constraints": {
             "required": true
@@ -58,6 +58,12 @@
         "name": "participant id",
         "type": "string",
         "description": "Participant's state-specific identifier. Must not be social security number or any personal identifiable information."
+      },
+      {
+        "name": "benefits end month",
+        "type": "yearmonth",
+        "format": "%m/%Y",
+        "description": "month and year when participant's benefits end"
       }
     ]
   }
--- a/etl/tests/Piipan.Etl.Tests/BulkUploadTests.cs
+++ b/etl/tests/Piipan.Etl.Tests/BulkUploadTests.cs
@@ -17,7 +17,8 @@ namespace Piipan.Etl.Tests
             var writer = new StreamWriter(stream);
             if (includeHeader)
             {
-                writer.WriteLine("last,first,middle,dob,ssn,exception,case id,participant id, benefits end month");
+                writer.WriteLine("last,first,middle,dob,ssn,exception,case id,participant id,benefits end month");
+
             }
             foreach (var record in records)
             {
--- a/etl/tests/Piipan.Etl.Tests/BulkUploadTests.cs
+++ b/etl/tests/Piipan.Etl.Tests/BulkUploadTests.cs
@@ -17,7 +17,7 @@ namespace Piipan.Etl.Tests
             var writer = new StreamWriter(stream);
             if (includeHeader)
             {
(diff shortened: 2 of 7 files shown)


etl/docs/csv/import-schema.json
```diff
@@ -58,6 +58,12 @@
         "name": "participant id",
         "type": "string",
         "description": "Participant's state-specific identifier. Must not be social security number or any personal identifiable information."
+      },
+      {
+        "name": "benefits end month",
+        "type": "yearmonth",
+        "format": "%m/%Y",
+        "description": "month and year when participant's benefits end"
```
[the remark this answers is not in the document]
echappen: Done!

etl/tests/Piipan.Etl.Tests/BulkUploadTests.cs
```diff
@@ -94,7 +97,7 @@ public void ReadAllFields()
         {
             var logger = Mock.Of<ILogger>();
             var stream = CsvFixture(new string[] {
-                "Last,First,Middle,01/01/1970,000-00-0000,Exception,CaseId,ParticipantId"
+                "Last,First,Middle,01/01/1970,000-00-0000,Exception,CaseId,ParticipantId,01/1970"
```
kingcomma, line 100: *Thought:* I can't come up with a scenario where this would bite us, but that this date would technically not pass local validation (using `validate.py`) but is correctly handled by the bulk upload application.

kingcomma (approved):
👍 LGTM!

etl/tests/Piipan.Etl.Tests/BulkUploadTests.cs
echappen: Thanks for pointing that out—I missed that since the tests pass with this format. I have some work in [another PR](https://github.com/18F/piipan/pull/930) that does some date validation in `PiiRecordMap` that could be useful for our other date fields (and would allow this to be caught in tests). Thoughts on [this approach](https://github.com/18F/piipan/pull/930/files#diff-b674da1eab2ed3b8e9fba00bf85f84b0b31673e317495b747bb072f39fbc7243R51)?

etl/tests/Piipan.Etl.Tests/BulkUploadTests.cs
```diff
@@ -17,7 +17,7 @@ static Stream CsvFixture(string[] records, bool includeHeader = true)
             var writer = new StreamWriter(stream);
             if (includeHeader)
             {
-                writer.WriteLine("last,first,middle,dob,ssn,exception,case id,participant id");
+                writer.WriteLine("last,first,middle,dob,ssn,exception,case id,participant id, benefits end month");
```
ryanhofdotgov, line 20:
```suggestion
                writer.WriteLine("last,first,middle,dob,ssn,exception,case id,participant id,benefits end month");

```

Issue #729, opened 2021-04-21
Stream all resource logs to an event hub


ryanhofdotgov:
As a security engineer, I want to integrate all resource logs from the system with a 3rd party Security Information and Event Management (SIEM) tool like Splunk. Azure solution is Azure Monitor with Event Hub.

Some reference docs to start with:
- https://docs.microsoft.com/en-us/azure/event-hubs/event-hubs-metrics-azure-monitor
- https://azure.microsoft.com/en-us/updates/new-app-service-integration-with-azure-monitor-preview/
- https://docs.microsoft.com/en-us/azure/event-hubs/event-hubs-diagnostic-logs
- https://docs.microsoft.com/en-us/azure/active-directory/reports-monitoring/tutorial-azure-monitor-stream-logs-to-event-hub

ryanhofdotgov:
From discussion with FastTrack:
* https://www.splunk.com/en_us/blog/tips-and-tricks/splunking-microsoft-azure-monitor-data-part-1-azure-setup.html
* https://www.splunk.com/en_us/blog/tips-and-tricks/splunking-microsoft-azure-monitor-data-part-2-splunk-setup.html

And also:
* https://splunkbase.splunk.com/app/3110/#/overview
* https://docs.splunk.com/Documentation/AddOns/released/MSCloudServices/About
* https://www.splunk.com/en_us/blog/tips-and-tricks/getting-microsoft-azure-data-into-splunk.html
* https://techcommunity.microsoft.com/t5/azure-sentinel/sending-enriched-azure-sentinel-alerts-to-3rd-party-siem-and/ba-p/1456976

ryanhofdotgov:
1st priority is stream Front Door logs to Event Hub.

ryanhofdotgov:
Removing "spike" in subject line, as I think we should be able to move beyond just the research phase in this sprint.

kingcomma:
Adding diagnostic settings (i.e., directing resource logs to event hub) via ARM template: https://docs.microsoft.com/en-us/azure/azure-monitor/essentials/resource-manager-diagnostic-settings

Using `az` CLI to discovery available log categories: https://docs.microsoft.com/en-us/cli/azure/monitor/diagnostic-settings/categories?view=azure-cli-latest#az_monitor_diagnostic_settings_categories_list

kingcomma:
@ryanhofdotgov With front door logs now set up, is there a priority for the remaining resources? If not, I'll just go through our list type-by-type.

[...]

PR #910, opened 2021-05-24
Create Event Hub resources for streaming logs

kingcomma:
- Creates an Event Hub namespace following our naming conventions: `{prefix}-evh-monitoring-{env}`
- Creates an event hub resource named `logs` intended as a single location for all resource logs for a given environment

Note: Event hubs have two SKUs, "Basic" and "Standard". I've chosen Standard, which gives us certain flexibility like the ability to "capture" logs in a storage account for debugging purposes.

First step for #729, to be followed by pointing resource logs to the event hub.


ryanhofdotgov (approved):
LGTM

PR #911, opened 2021-05-24
Add Benefits End Date to match query response, query results table, a…

### LINKED ISSUE #898: Query API returns final benefit date
As a state agency worker, I want to see the final date for which a participant received benefits when viewing a match so that I can avoid contacting another state or the applicant when it is not needed.

When moving to another state, a participant may terminate benefits but the last benefit date may still be in the future.  The participant may be eligible for benefits in their new state of residence as soon as the coverage period in the prior state expires.

AC:
In query API responses that include a match:
- [ ] If a final benefit date has been set for the participant, that date should be returned
- [ ] If no final benefit date has been set for the participant, that should be clear from the API response

Related to #860 - that's the same feature implemented in bulk uploads

echappen:
Relies on #905
Closes #898


ryanhofdotgov:
@echappen I was expecting that this would have the same date granularity as the input data in the CSV file; i.e., just month and year. Did you happen to get any [clarification of benefit granularity](https://github.com/18F/piipan/issues/860#issuecomment-845451513)?

echappen:
I asked LA on how specific they got with date. For them it's on the month-level. Is your comment suggesting that we keep the date as month-only in the database and as a property in C#?

ryanhofdotgov:
I was thinking that we should keep it YYYY-MM as far as all our external API users are concerned. We store it internally as a date field in the db (calculating the last day of the month), but that's just an implementation detail that we were "forced" to do, since PostgreSQL doesn't have a partial date type other than just storing the value as a string.

The output value in the JSON is really only as precise as the input value (YYYY-MM) in the CSV, so sending results back that have a higher precision might imply otherwise.

echappen:
Ok I see what you're saying now, thanks for clarifying. I'll change the json response to only return YYYY-MM and label it as benefits_end_month

echappen:
@ryanhofdotgov I'm very close to being done with this. I did the date-to-month conversion as early as I could at the database level. The last issue I'm having is that null values for `benefit_end_date` coming from the database are being serialized as "0001-01".  Does this look on the right track?

docs/openapi/generated/duplicate-participation-api/openapi.yaml
```diff
@@ -100,6 +100,11 @@ paths:
                         participant_id:
                           type: string
                           description: Participant's state-specific identifier. Must not be social security number or any personal identifiable information.
+                        benefits_end_month:
+                          type: string
+                          format: date
```
ryanhofdotgov, line 105: Not sure if our intended JSON format will validate correctly against this spec. 
[Swagger docs](https://swagger.io/docs/specification/data-models/data-types/) says `format: date` will expect full-date from [RFC 3339 ](https://datatracker.ietf.org/doc/html/rfc3339#section-5.6)(which uses ISO 8601).

That RFC says `full-date` is `full-date       = date-fullyear "-" date-month "-" date-mday`.

ryanhofdotgov (commented):
Yup – thanks!

ryanhofdotgov:
Oh – I would add an integration test that captures the null use case that you mention above.

docs/openapi/generated/duplicate-participation-api/openapi.yaml
echappen: Would the alternative be pattern instead of format then? It doesn't look like any formats match what we want:
```
benefits_end_month:
  type: string
  pattern: '^\d{4}-\d{2}$'

File: match/src/Piipan.Match.State/MatchResponse.cs [hunk 2]
```diff
@@ -57,6 +57,10 @@ public PiiRecord()
         [JsonProperty("participant_id")]
         public string ParticipantId { get; set; }
 
+        [JsonProperty("benefits_end_month")]
+        [JsonConverter(typeof(DateMonthConverter))]
+        public DateTime BenefitsEndMonth { get; set; }
```
ryanhofdotgov, line 62: I think your problem might be related to making this nullable. 
```suggestion
        public DateTime? BenefitsEndMonth { get; set; }
```

In general, we probably need to carefully consider null versus empty string and revisit ParticipantId and any other new fields that are optional. Similarly, on the CSV input side, we should be trimming white space and specifying null for empty string in database.

docs/openapi/generated/duplicate-participation-api/openapi.yaml
ryanhofdotgov: Yeah, that's close enough – if you wanted to get really fancy, you'd have to do something similar to https://www.regular-expressions.info/dates.html

match/src/Piipan.Match.State/MatchResponse.cs
echappen: Yep, that did it! I added some integration tests to cover that. I can make an issue to evaluate our overall handling of nulls vs empty strings.

docs/openapi/generated/duplicate-participation-api/openapi.yaml
echappen: Done!

match/src/Piipan.Match.State/MatchResponse.cs
echappen: I made it nullable in both state and orchestrator responses. Manually tested and looks gtg.

etl/docs/csv/example.csv
```diff
@@ -1,4 +1,4 @@
-﻿last,first,middle,dob,ssn,exception,case id,participant id, benefits end month
+last,first,middle,dob,ssn,exception,case id,participant id,benefits end month
```
kingcomma, line 1: *Chore:* Just flagging this merge conflict (now that column names have underscores) that Github also flagged.

match/docs/openapi/schemas/pii-record.yaml
```diff
@@ -38,3 +38,8 @@ PiiRecord:
     participant_id:
       type: string
       description: "Participant's state-specific identifier. Must not be social security number or any personal identifiable information."
+    benefits_end_month:
+      type: string
+      pattern: '^\d{4}-\d{2}$'
+      example: "2021-01"
+      description: "Participant's ending benefits month"
```
kingcomma, line 45: *Nit:* Since states are using these docs to understand how to handle response data, I'm wondering if this description could be clearer? @erinzstrenio or @cjhskippy — any opinions?

I think there's also a looming question of how much policy gets encoded in these docs. That's probably a question once we start including exception flags (i.e., do the docs need to detail how records flagged one way or another should be handled by state workers)?

match/src/Piipan.Match.State/Api.cs
```diff
@@ -97,7 +97,7 @@ internal static (string, dynamic) Prepare(MatchQueryRequest request, ILogger log
                 last = request.Query.Last,
                 first = request.Query.First,
             };
-            var sql = "SELECT upload_id, first, last, middle, dob, ssn, exception, case_id CaseId, participant_id ParticipantId FROM participants " +
+            var sql = "SELECT upload_id, first, last, middle, dob, ssn, exception, case_id CaseId, participant_id ParticipantId, to_char(benefits_end_date, 'YYYY-MM') BenefitsEndMonth FROM participants " +
```
kingcomma, line 100: Nice!

match/tests/Piipan.Match.State.IntegrationTests/ApiIntegrationTests.cs
```diff
@@ -172,6 +176,7 @@ public async void ApiMatchesIngoreFirstAndMiddle()
 
             // Assert
             Assert.Single(resultRecord.Matches);
+            Assert.Null(resultRecord.Matches[0].BenefitsEndMonth);
```
kingcomma, line 179: *Tiny nit:* The name of this test — `ApiMatchesIngoreFirstAndMiddle` — no longer accurately reflects what's being tested. This is likely overly cautious, but I could foresee a future where first and middle are no longer ignored, this test gets removed, and the `null` check on `BenefitsEndMonth` is dropped inadvertently.

*Suggestion:* I lean towards moving this check to its own test, which may also serve as a place to run the same check on other optional date fields (e.g., most recent benefits months).

query-tool/src/Piipan.QueryTool/PiiRecord.cs
```diff
@@ -59,5 +59,9 @@ public class PiiRecord : IQueryable
         [Display(Name = "Lookup ID")]
         [JsonPropertyName("lookup_id")]
         public string LookupId { get; set; }
+
+        [Display(Name = "Benefits End Month")]
+        [JsonPropertyName("benefits_end_month")]
+        public string? BenefitsEndMonth { get; set; }
```
kingcomma, line 65: *Question:* Was there some case that required this property to be declared nullable? There are other strings in this model that handle null values (e.g., participant ID) that aren't typed the same way. I think this is actually triggering a build warning:

```
PiiRecord.cs(65,22): warning CS8632: The annotation for nullable reference types should only be used in code within a '#nullable' annotations context.
```

kingcomma (commented):
Just a few small nits/suggestions (and a merge conflict to resolve).

FYI — I deployed these updates to our `-dev` environment for my testing. (New deploy scripts were so nice to have!)

query-tool/src/Piipan.QueryTool/PiiRecord.cs
ryanhofdotgov: I haven't looked into any of the details, but if it is helpful: DateTime is a struct, not a class There may be some nuance here that we need to understand better.

ryanhofdotgov: ^ To be more explicit: structs is a value type, like primitives like int, not a reference type. That's at least partly why the code was getting a value type with all the default initialization values.

match/tests/Piipan.Match.State.IntegrationTests/ApiIntegrationTests.cs
echappen: Good point! I split up these tests

query-tool/src/Piipan.QueryTool/PiiRecord.cs
echappen: To echo Ryan's comment, if the sql query returned null for `benefits_end_date`, the date would be converted to that default "0001" value we were seeing in the json response. Setting the property as nullable avoided that. 

To get rid of the warning @kingcomma saw, I added `#nullable enable` to the PiiRecord file. This spawned a bunch of other warnings for the rest of the PiiRecord fields:
```
Non-nullable property 'FirstName' is uninitialized. Consider declaring the property as nullable.
```
I made the fields nullable for now, but open to other ways of solving it.

kingcomma: I may be over-thinking this — I was assuming since this is the model for the Query Tool it can safely ignore the `null`/date issues (since it will only ever receive a string coming from the orchestrator API) and be set up to accept an optional string like the other fields (`MiddleName`, `ParticipantId`, `LookupId`) in the model.

Would just `public string BenefitsEndMonth { get; set; }` suffice here?

echappen: > Would just public string BenefitsEndMonth { get; set; } suffice here?

It could, although I find it confusing how these fields can be nullable in the json response but not marked as such here. I would expect this PiiRecord and match's PiiRecord to mimic each other. It feels like the nullable aspect of these fields gets obscured when operated on later.

kingcomma: I think this is all good fodder for #939. My main concern was consistency and avoiding a future developer having to wonder why one nullable string field is handled differently than another within the same model, which enabling nullable annotations addresses.

_Non-blocking thought:_ This has made me realize it may run into issues in the future with using `PiiRecord` as the model for both API requests and responses. For example, marking fields like `CaseId` and `State` as non-nullable (using the null-forgiving operator) is logical for responses since API responses will always contain values for those fields. But it breaks form submission since API requests will never contain values for those fields.

echappen: Using PiiRecord as the model for both API requests and responses may be the source of my confusion. For moving forward on this PR, I'll use your recommendation of keeping all fields required and removing the `#nullable enabled`. Then we can address all at once in #939.

match/docs/openapi/schemas/pii-record.yaml
cjhskippy: One possibility: 

The final month when the participant will receive benefits.  This field should be left empty if the participant will continue receiving benefits until the next recertification period.

erinzstrenio: Suggested edit to description: 

"The final month when a participant will receive benefits. Leave this field blank if the participant will continue receiving benefits until the next recertification period."

ryanhofdotgov (approved):
Thanks for pulling all these details together!

One final thought (that is not an ask to change this PR or even do specific follow up 😁 ), but just considerations if we have similar situations in future.

We have a lot of places in the code were we do translations back and forth between strings and dates. Our API boundaries can make for a lot of busy work: ETL -> DB, DB -> state API, state API -> orchestrator API, orchestrator API -> query tool.

It might be better to have a consistent internal representation (date object of some kind, set up in a particular way), even if we are converting to an ISO string at the API boundaries.

Here, dates are used in a few different ways:
- database date object calculates last day of month
- API .NET code (and their tests) use default of 1st day of month
- Query tool uses string as its internal representation rather than date object

Again, this doesn't impact behavior of resulting system, but just stood out.

Issue #861, opened 2021-05-17
Track recent benefit months


cjhskippy:
As a state agency worker, I want to know if an applicant actually received benefits in another state where they recently resided so that I can make an accurate determination of whether they are eligible in my state.

This may be especially relevant for a non-head-of-household in a still-open case.  For example, a child who spends some time with out-of-state relatives or shared custody.

Note that we believe all state agencies use the same definition of "benefit month" - that is, it's not necessary to track the specific days, we can say "June 2020" and it should mean the same thing for eligibility dates across all states.

AC:
- [ ] Bulk uploads include fields for the most recent 3 calendar months if the participant received benefits in their state.  These months would be consecutive, (example: February 2021, March 2021, April 2021).
- [ ] There must not be ambiguity about which months are covered during the upload.  This likely requires states to explicitly state which months they consider the "most recent" when they perform the upload.

PR #912, opened 2021-05-25
Add specs for recent benefits months in CSV import schema

echappen:
partially addresses #861 

I'd like to get consensus on these specs for the CSV import, since it'll inform how the data is parsed and stored. 

This new CSV column would look like this:
| recent benefit months |
|---|
| 2/2021,3/2021,4/2021 |
| 1/2021,3/2021,4/2021 |
| 1/2021,2/2021,3/2021 |

Does this look alright?

### DIFF
--- a/ddl/per-state.sql
+++ b/ddl/per-state.sql
@@ -29,7 +29,8 @@ CREATE TABLE IF NOT EXISTS participants(
 	upload_id integer REFERENCES uploads (id),
   	case_id text NOT NULL,
   	participant_id text,
-	benefits_end_date date
+	benefits_end_date date,
+    	recent_benefit_months date[]
 );
 
 COMMENT ON TABLE participants IS 'Program participant Personally Identifiable Information (PII)';
@@ -42,6 +43,7 @@ COMMENT ON COLUMN participants.exception IS 'Placeholder for value indicating sp
 COMMENT ON COLUMN participants.case_id IS 'Participant''s state-specific case identifier';
 COMMENT ON COLUMN participants.participant_id IS 'Participant''s state-specific identifier';
 COMMENT ON COLUMN participants.benefits_end_date IS 'Participant''s ending benefits date';
+COMMENT ON COLUMN participants.recent_benefit_months IS 'Participant''s recent benefit months';
 
 CREATE INDEX IF NOT EXISTS participants_ssn_idx ON participants (ssn, upload_id);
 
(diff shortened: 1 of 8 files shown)


ryanhofdotgov:
This seems like a good example of just always using ISO 6801 for formatting dates, as it already has a defined format; i.e. YYYY-MM

Using commas within a CSV field is allowed, but CSV generators/readers tend not to implement it super well. It requires the field be double-quoted. Space delimited might be better here.

ryanhofdotgov:
Tying it back to our other conversation about me regretting not using ISO 8601 from the get-to for the DOB:
I think I might open up a PR to deprecate the MM/DD/YYYY format, update the spec/test files to be YYYY-MM-DD, and update the ETL code to (temporarily) support both DOB formats.

echappen:
Yeah I was wondering about doing that conversion in a separate PR, instead of incorporating it into one of these other API changes. But as we continue to add dates, it's becoming apparent to just do it earlier than later.

echappen:
@ryanhofdotgov I also wanted to confirm that this description in the import schema is correct:

> Does not include current benefit month. Months do not need to be consecutive.

ryanhofdotgov:
@echappen Can you point me to where that text is from? I see: "These months would be consecutive"

echappen:
Ah, totally missed that. Will change.

kingcomma:
What is the expected value for this field for participants who have been active for less than 3 months in a state's system? I.e., if a participant is new to the state in `May 2021` and the state is submitting a bulk upload in `June 2021`, would that participant's `recent benefit months` field contain only one month (`5/2021`)?

If so, does the `format` or description need to be updated to reflect the possibility that the field would hold less than 3 dates.

kingcomma:
On the topic of date formatting, I think there may be another argument for ISO 8601 — I'm getting the following error when validating the `benefits end month` field (I must have missed this in prior review) against a value of `5/2021`:

```
CastError('Field "benefits end month" can\'t cast value "5/2021" for type "yearmonth" with format "%m/%Y"')
```

Upon re-reading the [table schema docs](https://specs.frictionlessdata.io/table-schema/#yearmonth) it looks like `yearmonth` does not support a `format` property.

echappen:
Good catch @kingcomma! I removed the formatting for the yearmonth column, and added some language to indicate that less than 3 months can be entered.

PR #917, opened 2021-05-25
Stream Front Door logs to Event Hub

kingcomma:
- Logs are streamed via "diagnostic settings" added via ARM template
- Added prefix and env parameters to front door ARM template so event hub resource name can be constructed

This work establishes a precedent of naming the [diagnostic setting](https://docs.microsoft.com/en-us/azure/azure-monitor/essentials/diagnostic-settings?tabs=CMD) `stream-logs-to-event-hub`. My current plan is for remaining resources which need their logs streamed to event hub to get a diagnostic setting with the same name.

### DIFF
--- a/iac/add-front-door-to-app.bash
+++ b/iac/add-front-door-to-app.bash
@@ -37,7 +37,9 @@ main () {
       frontDoorName="$front_door_name" \
       resourceGroupName="$resource_group" \
       resourceTags="$RESOURCE_TAGS" \
-      wafPolicyName="$waf_name"
+      wafPolicyName="$waf_name" \
+      prefix="$PREFIX" \
+      env="$ENV"
 
   script_completed
 }
--- a/iac/arm-templates/front-door-app-service.json
+++ b/iac/arm-templates/front-door-app-service.json
@@ -48,11 +48,24 @@
             "metadata": {
                 "description": "Describes if it is in detection mode or prevention mode at policy level."
             }
+        },
+        "prefix": {
+            "type": "string",
+            "metadata": {
+                "description": "Environment-specific resource naming prefix"
+            }
+        },
+        "env": {
+            "type": "string",
+            "metadata": {
+                "description": "Development environment indicator"
+            }
         }
     },
     "variables": {
         "location": "global", /* front-door location must be global */
-        "wafLocation": "global" /* waf location must be global */
+        "wafLocation": "global" /* waf location must be global */,
+        "eventhubName": "[concat(parameters('prefix'), '-evh-monitoring-', parameters('env'))]"
     },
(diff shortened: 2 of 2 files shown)


kingcomma:
Partially addresses #729.

ryanhofdotgov (approved):
LGTM, though wondering if there is any benefit to constructing names in the template, or constructing them in the IaC common script and passing them in here. Some of @echappen's recent changes have pushed more into the common script.

echappen:
I don't see an immediate need to reference the event hub names elsewhere in iac, but that could be possible in the future. Constructing those names outside the template would offer that flexibility.

kingcomma:
@echappen @ryanhofdotgov Good points — I meant to comment on this. I'm good moving the name construction out to `iac-common` with the next batch of log streaming updates.

FWIW, my inclusion of `prefix` and `env` parameters has been a bit of an experiment in what's possible within ARM templates. I'm not proposing this yet, but I've been mulling over what we could do with parameter files and what it would look like to define a couple JSON files with parameter values that are always passed to `az deployment group create`:

(1) `global`: A JSON file with names/values that are consistent across all our environments. E.g.:
```
{
  "baseNames": {
    "eventHub": "evh-monitoring",
    "metricsCollection": "func-metricscol",
    "perStateMatchAPI": "func-{state}match"
    [...]
}
```

(2) `env`: A JSON file with names and values that change based on environment. E.g.:
```
{
  "prefix": "tts",
  "name": "dev",
  "resourceGroups": {
    ...
  },
  "cloudName": "AzureCloud",
  "resourceTags": {
    "Project": "piipan"
  },
  [...]
}
```

Combined with `prefix` and `env` parameters (which could also be included in a parameters file or passed separately) ARM templates would be able to build all the various names and resource IDs as needed. E.g.:

```
azure_env="tts/dev"
az deployment group create \
  -g "rg-core-dev" \
  -f ./arm-templates/template.json \
  --parameters \
    global=@global.json \
    env="@env/${azure_env}.json" \
```

And then in the template itself:
```
"eventHubName": "[concat(parameters('env')['prefix'], '-', parameters('global')['baseNames']['eventHub'], '-', parameters('env')['name'])]"
```

In some sense, this would be just shifting code from bash to JSON. But I think there are a few potential benefits:

- Having environment details makes it easier to handle per-state resources exclusively in ARM templates (e.g., looping through each state and generating state-specific names like in [apim.json](https://github.com/18F/piipan/blob/dev/iac/arm-templates/apim.json#L63))
- Related resources can be more easily combined into a single template, _maybe_ leading to a one-template-per-resource-group approach
- `iac-common` from a the `azure-env` dependency

Issue #897, opened 2021-05-19
Query API returns the most recent benefit months


ryanhofdotgov:
As a state agency worker, I want to know if an applicant received benefits in another state where they recently resided so that I can make an accurate determination of whether they are eligible in my state.

This may be relevant for a non-head-of-household in a still-open case. For example, a child who spends some time with out-of-state relatives or shared custody.

Note that we believe all state agencies use the same definition of "benefit month" - that is, it's not necessary to track the specific days, we can say "June 2020" and it should mean the same thing for eligibility dates across all states.

AC:
For each match returned by the query API:
- [ ] The response indicates whether the participant received benefits within each of the most recent 3 months.
- [ ] Specify in the response which months are covered (Example: February 2021, March 2021, April 2021)

Related to #861 - that's the same feature implemented in bulk uploads.

cjhskippy:
Added the AC that we should explicitly name which months are covered.  An API consumer can't infer the "last 3 months" based on today's date - they'd have to know when the other state uploaded their bulk data (was the upload before or after the end of the month?)

echappen:
@ryanhofdotgov How do we want to represent this data in the database? Would it just remain the human-readable text provided by states, or would we convert each into a date?

ryanhofdotgov:
Let's store it as a native database date field – this allows easier time-based SQL queries against the data set. Not that I think we anticipate that right now – it's just the most natural approach.

ryanhofdotgov:
More specifically, with this approach, we would model it as a table column that is an array of dates.

echappen:
@ryanhofdotgov @cjhskippy  For the JSON response, I'm thinking we'd represent recent benefit months as an array of strings. Each item would be the same ISO year-month format as `benefits_end_month`. Does that sound good?

example:
```
recent_benefit_months: ["2021-04", "2021-03", "2021-02"]
```

ryanhofdotgov:
Sounds good – I would explicitly sort the dates in descending order at the API level and document that in the API spec.

PR #930, opened 2021-05-26
Add recent benefit months to bulk upload

### LINKED ISSUE #861: Track recent benefit months

echappen:
Closes #861 
Continuation of #912 
See discussion in #897 on making this table column an array of dates. 

### Notes
- Ideally I would like to convert the recent benefit months into a List of DateTimes upon conversion from CSV row into a `PiiRecord`, but I couldn't figure out how to do that. So instead, the property is kept as a string and formatted into a Postgresql array of dates at db insertion. 
- This was the first case I could find of a nullable field that also required validation in `PiiRecordMap`
- Since this changes the database schema, I'll drop the participants db server and re-run iac on tts/dev before merging

### DIFF
--- a/etl/docs/csv/example.csv
+++ b/etl/docs/csv/example.csv
@@ -1,11 +1,11 @@
-﻿last,first,middle,dob,ssn,exception,case id,participant id, benefits end month
-Farrington,Theodore,Carri,10/13/1931,000-12-3456,ReasonXYZ,caseid1,,5/2021
-Lynn,Wesley,Eura,8/01/1940,000-12-3457,,caseid2,,
-Cullen,S,,1/1/2015,000-12-3458,,caseid3,,
-Alger,Megan,Kamala,1/25/2020,000-12-3459,,caseid4,,
-Yu,,,11/2/1973,000-12-3460,,caseid5,,
-Wallen,Homer,Madonna,4/4/1913,000-12-3461,ReasonXYZ,caseid6,,
-Benner,Cristina,Wilton,07/22/1961,000-12-3462,,caseid7,,
-Whittaker,Troy,G,2/15/1985,000-12-3463,,caseid8,,
-Grier,Paulette,Tequila,12/24/1997,000-12-3464,,caseid9,,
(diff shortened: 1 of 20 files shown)


etl/docs/csv/example.csv
```diff
@@ -1,11 +1,11 @@
...
-Cullen,S,,1/1/2015,000-12-3458,,caseid3,,
-Alger,Megan,Kamala,1/25/2020,000-12-3459,,caseid4,,
-Yu,,,11/2/1973,000-12-3460,,caseid5,,
-Wallen,Homer,Madonna,4/4/1913,000-12-3461,ReasonXYZ,caseid6,,
-Benner,Cristina,Wilton,07/22/1961,000-12-3462,,caseid7,,
-Whittaker,Troy,G,2/15/1985,000-12-3463,,caseid8,,
-Grier,Paulette,Tequila,12/24/1997,000-12-3464,,caseid9,,
-Suarez,Amelia,Shavonda,1/19/1950,000-12-3465,,caseid10,participantid1,
+﻿last,first,middle,dob,ssn,exception,case id,participant id, benefits end month, recent benefit months
+Farrington,Theodore,Carri,10/13/1931,000-12-3456,ReasonXYZ,caseid1,,2021-05, 2021-04 2021-03 2021-02
```
echappen, line 2:
```suggestion
Farrington,Theodore,Carri,10/13/1931,000-12-3456,ReasonXYZ,caseid1,,2021-05,2021-04 2021-03 2021-02

```

PR #946, opened 2021-05-27
Remove whitespace from csv column headers

### LINKED ISSUE #935: piipan CSV column names shouldn't contain spaces
CSV libraries, tools, and custom code don't have a good reputation for handling spaces in column names:
https://www.google.com/search?q=csv+column+names+spaces

The recently added columns for case id, participants id, benefit end month, recent benefit months, etc. should be altered to have an underscore instead of a space.

echappen:
Closes #935


kingcomma (approved):
👍

PR #947, opened 2021-05-27
Keep bottom CUI banner flush with screen bottom

echappen:
Going from this:

Screen Shot 2021-05-27 at 11 33 51 AM

To this:

Screen Shot 2021-05-27 at 11 29 15 AM

Content will push banner down when taller than screen height. 

It wasn't "broken" per se, but now that we're demo-ing more often to more people, it may be visually bothersome.


query-tool/src/Piipan.QueryTool/wwwroot/sass/_uswds-theme-custom-styles.scss
```diff
@@ -22,3 +22,12 @@ i.e.
 .cui {
   @include u-bg('cyan-5');
 }
+
+/*
+This keeps bottom CUI banner flush with screen bottom when main content is short,
+and pushes it down when main content is longer than screen height.
+82px is height of the CUI banner.
```
ryanhofdotgov, line 29: My front end experience is limited, so just reacting to the hard-coded 82px – I believe because the page design is responsive,  the height of the CUI banner going to change depending on the width of the browser window.

Examples:

1115x82:
Screen Shot 2021-05-28 at 11 07 16 AM

610x107:
Screen Shot 2021-05-28 at 11 07 01 AM

ryanhofdotgov: Less common ways this height value might change include when the dev statically changes the font size in CSS situations, or when the browser user changes their font size dynamically for accessibility purposes:
Screen Shot 2021-05-28 at 11 16 05 AM

Example:
1358x124:
Screen Shot 2021-05-28 at 11 14 58 AM

echappen: Agreed the hard coded height isn’t my favorite. Since I don’t imagine much usage with screens smaller than 13”, I figured this could be good enough while being as sparse with the custom code as I could be. 

In the past I’ve used these alternatives:

1. Creating a [flexbox “column” container](https://philipwalton.github.io/solved-by-flexbox/demos/sticky-footer/) around all page content with three top-level elements: one for the top banner, one for the bottom banner, and one for the main page content in the middle. The middle one expands to fill the empty vertical space and pushes bottom banner down. 

2. Making bottom banner position fixed to the bottom of screen. In order to prevent bottom banner from covering main content at smaller heights, bottom padding is adding to the main container that’s around the same height as the bottom banner (basically the same problem of determining the height of a dynamic element). 

4. Setting a min-height to main content that’s around the height of the largest screen being used. Kind of like a basic footer where you’d have to scroll to see bottom banner on smaller screens. 

5. Javascript to handle dynamic height changes

The only alternative I’d really recommend is #1, but it’s more custom css and would require markup changes. Want me to try that?

kingcomma: If it's helpful — It may be possible to pull of (1) with the existing USWDS utility tokens for [flex](https://designsystem.digital.gov/utilities/flex/) and [grid](https://designsystem.digital.gov/utilities/layout-grid/#auto-layout-columns).

ryanhofdotgov (approved):
@echappen I'm out of my depth here so defer to you on best tradeoff. I say LGTM 👍

echappen:
I'll give it a shot with USWDS flexbox like Michael mentions and see if that ends up being a better solution. Will timebox to 1 hour.

echappen:
Pushed a flexbox solution to this [PR](https://github.com/18F/piipan/pull/983)

Issue #445, opened 2021-03-03
Reject rows missing a first name


cjhskippy:
The NAC will require a first name for any matching to be performed.  We should update the bulk upload API and related documentation to reject these rows.

This is not strictly MVP material.  Rejecting these rows and flagging them for FNS to learn about where data gets messy is valuable, but will not impact any match results.

- [ ] Bulk upload rejects rows that do not include a first name
- [ ] Bulk upload provides error message with reason for rejecting rows missing a first name
- [ ] Bulk upload API documentation is updated
- [ ] Update examples, see https://github.com/18F/piipan/pull/951#discussion_r643343415

kingcomma:
@cjhskippy I was looking into issues around handling first names and saw this was closed by me 14 days ago. I think that was by accident and this should be reopened. Does that sound right to you?

cjhskippy:
No longer relevant.  Now that the NAC performs matches against a deidentified participant hash and avoids collection of PII, this can be closed.

PR #951, opened 2021-05-28
Add example requests and responses for DupPart API

### LINKED ISSUE #899: Add example schemas to OpenAPI docs
Per feedback from some states, add explicit examples to OpenAPI docs where request/response format may be unclear. Specifically:

- [ ] `/query` example responses for: 
    - [ ] single match
    - [ ] multiple matches
    - [ ] no match
- [ ] `/query` example requests for:
    - [ ] only required fields
    - [ ] all fields (required + optional)
- [ ] `/lookup_ids` example responses for:
    - [ ] stored queries with only required fields
    - [ ] stored queries with all fields (required + optional)

kingcomma:
- Regenerated documentation
- Tried to keep examples as close as possible to their associated models to minimize the chance they fall out of date
- Widdershins does not currently support multiple request examples for markdown, only displays first example

The primary change here is adding examples to the API documentation used by states: https://github.com/18F/piipan/blob/413b021c906658a44732348e1a358a1bc457ff71/docs/openapi/generated/duplicate-participation-api/openapi.md

Closes #899, but will need to be kept up to date with scheduled API updates (included #911).

### DIFF
--- a/docs/openapi/generated/duplicate-participation-api/openapi.yaml
+++ b/docs/openapi/generated/duplicate-participation-api/openapi.yaml
@@ -47,6 +47,24 @@ paths:
                       type: string
                       format: date
                       description: PII record's date of birth
+            examples:
+              All fields:
+                description: A request with values for all fields
+                value:
+                  query:
+                    first: string
+                    middle: string
+                    last: string
+                    ssn: 000-00-0000
+                    dob: '1970-01-01'
+              Only requried fields:
+                description: A request with values only for required fields
+                value:
+                  query:
+                    first: string
+                    last: string
+                    ssn: 000-00-0000
+                    dob: '1970-01-01'
       responses:
         '200':
           description: 'Matching PII records, if any exist'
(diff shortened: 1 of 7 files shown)


kingcomma:
My goal is to get these examples in front of states as soon as possible. But if others feel it's better to hold off on this until the API changes stabilize and include it in the release for the next sprint, I'm ok with that too.

ryanhofdotgov (approved):
Is the plan that we would merge this in, then update for all the changes that get merged by this sprint's end? (e.g., remove state_name, rename state_abbr to state)

Nitpick: I'd give a real example of a state or one of our defined example ones (EA, EB, EC). A separate issue that I'll open, but I'd like for our APIs to use the capitalized versions of the state abbreviations in the responses (and even in our environment variables that we set). We use them in lowercase form in resource identifiers, but seems more natural to list them in uppercase elsewhere.

kingcomma:
@ryanhofdotgov I merged #954 and rebased this branch to include the `state_name` work. I also included a change to use the example state abbreviations (`ea`, `eb`, `ec`). Those were small edits, but re-requesting review all the same.

docs/openapi/generated/duplicate-participation-api/openapi.yaml
```diff
@@ -101,6 +119,52 @@ paths:
...
+                        middle: string
+                        last: string
+                        ssn: 000-00-0000
+                        dob: '1970-01-01'
+                        state: eb
+                        state_abbr: eb
+                        exception: string
+                        case_id: string
+                        participant_id: string
+                      - first: null
```
ryanhofdotgov, line 158: Did we make `first` optional? Can it ever be `null`?

kingcomma: Speaking only to current functionality: `first` is an optional field in the bulk upload process. It's required by the DupPart API for requests, but can be `null` in responses given our current requirements.

This does lead to an unintuitive discrepancy: our [example bulk upload CSV](https://github.com/18F/piipan/blob/dev/etl/docs/csv/example.csv#L6) has a line with only last name, but to return that result via the DupPart API a state would need to submit PII with a first name.

ryanhofdotgov: Hmm... I think that's a detail that got missed in the churn. Product owner specifically talked about requiring first name in bulk upload and that's what we've been telling our state partners recently. You correctly flagged this issue as being closed pre-maturely:
https://github.com/18F/piipan/issues/445

PR #961, opened 2021-06-01
Update Query Tool to use new state API field

### LINKED ISSUE #953: Update Query Tool to use new `state` API field
Once #895 is complete, the Query Tool will need to be updated to drop `StateName` and swap `State` for `StateAbbr`.

kingcomma:
- Remove state_name field from models and templates
- Use state field in templates
- Update tests

Follow on work from #954.
Closes #953.


ryanhofdotgov (approved):
LGTM

Issue #969, opened 2021-06-01
Stream resource logs to an event hub


cjhskippy:
Resource logs to stream:

- [ ] Per-state ETL Functions (#1017)
- [x] Participant records PostgreSQL Server (#975)
- [x] Core PostgreSQL Server (#975)
- [x] Query Tool and Dashboard App Service instances (#962)
- [x] Metrics API and Collection Functions (#1015)

PR #975, opened 2021-06-02
Stream PSQL logs to event hub

kingcomma:
- Participant records DB
- Metrics "core" DB
- Only streams `PostgreSQLLogs` logs, does not include performance-related `QueryStoreRuntimeStatistics` or `QueryStoreWaitStatistics` logs

Partially addresses #969


ryanhofdotgov (approved):
LGTM

PR #976, opened 2021-06-02
Revert monitoring to accommodate AzureUSGoverment

ryanhofdotgov:
- Event Hub API version 2021-01-01-preview not supported in AzureUSGoverment:
  https://portal.azure.us/#blade/HubsExtension/ArmExplorerBlade
- Azure Monitor is not supported for Azure Blob Storage in AzureUSGovernment, reverts #950:
  https://docs.microsoft.com/en-us/azure/storage/blobs/monitor-blob-storage?tabs=azure-portal

### DIFF
--- a/iac/arm-templates/blob-storage.json
+++ b/iac/arm-templates/blob-storage.json
@@ -10,12 +10,6 @@
         },
         "storageAccountName": {
             "type": "string"
-        },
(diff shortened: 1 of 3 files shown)


kingcomma (approved):
👍

PR #981, opened 2021-06-02
Change working dir to app tree before publishing

### LINKED ISSUE #949: Target specific projects when deploying web apps
For both `dashboard` and `querytool`, `build.bash deploy [...]` publishes all projects listed in the solution file. This can result in multiple projects — e.g., `Piipan.QueryTool.csproj` and `Piipan.QueryTool.Tests.csproj` — being published side-by-side in `artifacts`, and consequently zipped together and deployed as a single app.

Additionally, `build.bash deploy [...]` publishes to the root `artifacts` directory, potentially placing files side-by-side with previous builds on developer's local machines.

To resolve:

- [ ] Specify a specific `.csproj` when running `dotnet publish` in `run_deploy`
- [ ] Publish to a random/unique directory for each run of `build.bash deploy [...]` (perhaps using a timestamp or an approach similar to [random_password](https://github.com/18F/piipan/blob/dev/iac/iac-common.bash#L72-L74) for determining the directory name)

[Related slack thread](https://gsa-tts.slack.com/archives/C018P1ZJPEC/p1622210338169800?thread_ts=1622209102.169400&cid=C018P1ZJPEC).

ryanhofdotgov:
Closes #949.

This avoids problem where solution file publishes both src and test trees resulting in a failed deploy. It doesn't solve the problem exactly as #949 suggests, but it should get the job done and is similar to how the deploy is written for our function apps.

### DIFF
--- a/dashboard/build.bash
+++ b/dashboard/build.bash
@@ -20,6 +20,12 @@ run_deploy () {
   verify_cloud
   set_constants
 
+
+  # Don't publish at solution level as it will publish both src and tests
+  # to the same directory resulting in a failed deployment:
+  # https://github.com/dotnet/sdk/issues/7238
+  pushd "$(dirname "$0")"/src/Piipan.Dashboard
+
   echo "Publishing project"
   dotnet publish -o ./artifacts
   echo "Deploying to Azure Environment ${azure_env}"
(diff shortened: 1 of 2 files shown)


kingcomma (approved):
👍 

*Non-blocking thought:* this creates an `artifacts` directory in `src/{ProjectName}` rather than at the top-level of the sub-system (as suggested in the [project structure example](https://gist.github.com/davidfowl/ed7564297c61fe9ab814)). Git ignores all artifacts directories though, so it's not issue.

PR #982, opened 2021-06-02
Make first name required in CSV

### LINKED ISSUE #445: Reject rows missing a first name

ryanhofdotgov:
Closes #445


ryanhofdotgov:
Putting this in the ice box as user story motivating #445 needs to be explored further.

PR #983, opened 2021-06-02
Use flexbox to keep bottom CUI banner sticky

echappen:
Alternative to #947 that uses Flexbox and avoids hard-coded heights. I couldn't get it working with USWDS utilities so had to write something custom.


ryanhofdotgov:
Nice 👍 – I'll give @kingcomma an opportunity to review since for CSS I'm just good for the obvious stuff.

kingcomma (approved):
LGTM!

_Non-blocking thought:_ In terms of USWDS utility classes, I _think_ this could be achieved by wrapping all the content in a container like `<div class="minh-viewport display-flex flex-column">` and then adding `flex-fill` to the container around `#main-content`. Something like:

```
<div class="minh-viewport display-flex flex-column">
    <partial name="_UsaBanner" />
    <partial name="_CUI" />
    <div class="usa-overlay"></div>
    <header class="usa-header usa-header--extended"> ... </header>
    <div class="container margin-bottom-3 flex-fill"> ... </div>
    <partial name="_CUI" />
</div>
```

echappen:
Ah I forgot about `minh-viewport`, thanks!

PR #984, opened 2021-06-02
Convert dob in bulk import to ISO 8601 format

### LINKED ISSUE #936: DoB field in CSV bulk import format should use ISO 8601
Existing bulk import format uses MM/DD/YYYY as it's date format. Let's just use ISO 8601 for consistency with the JSON API as well as just all around best practice for machine parseable data formats. Don't let [Excel issues](https://github.com/18F/piipan/blob/dev/etl/docs/bulk-import.md#notes) hold us back.

Implementation notes:
- See if it is low effort to make the change in a backwards compatible manner. i.e., continue to accept MM/DD/YYYY in ETL and validator for a couple of sprints, so all of the state's test code doesn't stop working
- Update docs to just specify ISO 8601 format
- Update our example CSV to use ISO 8601
- After waiting a sprint, update validator to only accept 8601, then after another sprint, update ETL to only accept 8601

echappen:
Closes #936


ryanhofdotgov (approved):
Just needs a clean up on the conflict in the .csv file (I think the optional byte order marker that Excel will put there was removed in a recent PR).

Otherwise 👍 – the .NET library and CSVHelper code just does the permissive thing by automatically parsing US style dates along with ISO 8601 ones. I'll open up a ticket to tighten that up eventually, so we only accept ISO 8601 in the ETL code to be consistent with the validator.

PR #985, opened 2021-06-02
Add benefits_end_month to openapi examples

kingcomma:
With #898 complete, adds new `benefits_end_month` field to examples.


kingcomma:
@echappen Examples snuck into the mix at the end of last sprint (#951) but I forgot to include them when reviewing #911.

echappen:
Thanks for catching that, and for adding these examples!

echappen (approved):

PR #988, opened 2021-06-03
Add recent benefit months to bulk upload

### LINKED ISSUE #861: Track recent benefit months

echappen:
Closes #861 
See discussion in #897 on making this table column an array of dates. 

### Notes
- Since this changes the database schema, I'll drop the participants db server and re-run iac on tts/dev before merging

### DIFF
--- a/etl/docs/csv/import-schema.json
+++ b/etl/docs/csv/import-schema.json
@@ -63,6 +63,12 @@
         "name": "benefits_end_month",
         "type": "yearmonth",
         "description": "The final month when a participant will receive benefits. Leave this field blank if the participant will continue receiving benefits until the next recertification period."
+      },
+      {
+        "name": "recent_benefit_months",
+        "type": "string",
+        "format": "%Y-%m %Y-%m %Y-%m",
+        "description": "List of up to the last 3 months that participant received benefits, formatted as yearmonths separated by spaces. Does not include current benefit month. Months need to be consecutive. Fewer than 3 months can be entered for newer participants."
       }
     ]
   }
--- a/etl/docs/csv/import-schema.json
+++ b/etl/docs/csv/import-schema.json
@@ -67,8 +67,10 @@
       {
         "name": "recent_benefit_months",
         "type": "string",
-        "format": "%Y-%m %Y-%m %Y-%m",
-        "description": "List of up to the last 3 months that participant received benefits, formatted as yearmonths separated by spaces. Does not include current benefit month. Months need to be consecutive. Fewer than 3 months can be entered for newer participants."
+        "description": "List of up to the last 3 months that participant received benefits, formatted as yearmonths separated by spaces. Does not include current benefit month. Months need to be consecutive. Fewer than 3 months can be entered for newer participants.",
+        "constraints": {
+          "pattern": "([0-9]{4}-[0-9]{2} ?){0,3}"
+        }
       }
     ]
   }
--- a/etl/docs/csv/import-schema.json
+++ b/etl/docs/csv/import-schema.json
@@ -69,7 +69,7 @@
         "type": "string",
         "description": "List of up to the last 3 months that participant received benefits, formatted as yearmonths separated by spaces. Does not include current benefit month. Months need to be consecutive. Fewer than 3 months can be entered for newer participants.",
         "constraints": {
-          "pattern": "([0-9]{4}-[0-9]{2} ?){0,3}"
+          "pattern": "([0-9]{4}-[0-9]{1,2} ?){0,3}"
         }
       }
     ]
--- a/etl/src/Piipan.Etl/BulkUpload.cs
+++ b/etl/src/Piipan.Etl/BulkUpload.cs
@@ -64,11 +64,6 @@ namespace Piipan.Etl
             }
         }
 
-        public static DateTime LastDayOfMonth(DateTime dateTime)
-        {
-            return new DateTime(dateTime.Year, dateTime.Month, DateTime.DaysInMonth(dateTime.Year, dateTime.Month));
-        }
-
         internal static IEnumerable<PiiRecord> Read(Stream input, ILogger log)
         {
             var reader = new StreamReader(input);
@@ -148,12 +143,6 @@ namespace Piipan.Etl
 
                 foreach (var record in records)
                 {
-                    if (record.BenefitsEndDate.HasValue)
-                    {
-                      DateTime benefitsEndDate = record.BenefitsEndDate.Value;
-                      record.BenefitsEndDate = LastDayOfMonth(benefitsEndDate);
-                    }
-
                     using (var cmd = factory.CreateCommand())
                     {
                         cmd.Connection = conn;
--- a/etl/src/Piipan.Etl/BulkUpload.cs
+++ b/etl/src/Piipan.Etl/BulkUpload.cs
@@ -170,7 +170,7 @@ namespace Piipan.Etl
                         AddWithValue(cmd, DbType.String, "case_id", record.CaseId);
                         AddWithValue(cmd, DbType.String, "participant_id", (object)record.ParticipantId ?? DBNull.Value);
                         AddWithValue(cmd, DbType.DateTime, "benefits_end_date", (object)record.BenefitsEndDate ?? DBNull.Value);
-                        AddWithValue(cmd, DbType.Object, "recent_benefit_months", (object)FormatDatesAsPgArray(record.RecentBenefitMonths) ?? DBNull.Value);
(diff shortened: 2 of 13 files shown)


etl/docs/csv/import-schema.json
```diff
@@ -63,6 +63,12 @@
         "name": "benefits_end_month",
         "type": "yearmonth",
         "description": "The final month when a participant will receive benefits. Leave this field blank if the participant will continue receiving benefits until the next recertification period."
+      },
+      {
+        "name": "recent_benefit_months",
+        "type": "string",
+        "format": "%Y-%m %Y-%m %Y-%m",
```
ryanhofdotgov, line 70: If a CSV only has one or two dates, what will validate.py report against it?

etl/src/Piipan.Etl/BulkUpload.cs
```diff
@@ -187,5 +188,22 @@ static void AddWithValue(DbCommand cmd, DbType type, String name, object value)
...
+            if (String.IsNullOrEmpty(input)) return "{}";
+            string formatted = "";
+            string[] datestrings = input.Split(' ');
+            List<string> formattedDateStrings = new List<string>();
+            formatted += "{";
+            foreach (var datestring in datestrings)
+            {
+              if (String.IsNullOrEmpty(datestring)) continue;
+              DateTime date = Convert.ToDateTime(datestring);
+              formattedDateStrings.Add(date.ToString("yyyy-MM-dd"));
```
ryanhofdotgov, line 202: Will the full date persisted to the db be 1st of the month or the last of the month? I'd do the same as benefit_end_date for consistency, as well as it representing inclusivity through the whole month.

Just for easier low-level inspection in a table, I might sort the dates in descending order before saving to database.

etl/src/Piipan.Etl/PiiRecordMap.cs
```diff
@@ -43,6 +46,28 @@ public PiiRecordMap()
...
+            Map(m => m.RecentBenefitMonths).Name("recent_benefit_months")
+                .Validate(field => {
+                  if (String.IsNullOrEmpty(field.Field)) return true;
+
+                  string[] formats={"yyyy-MM", "yyyy-M"};
+                  string[] dates = field.Field.Split(' ');
+                  foreach (string date in dates)
+                  {
+                    DateTime dateValue;
+                    var result = DateTime.TryParseExact(
```
ryanhofdotgov, line 59: Nice! This is the approach I was thinking we would add to all the date fields eventually.

etl/src/Piipan.Etl/PiiRecord.cs
```diff
@@ -24,5 +25,6 @@ public class PiiRecord
         public string? ParticipantId { get; set; }
         public string? Exception { get; set; }
         public DateTime? BenefitsEndDate { get; set; }
+        public string? RecentBenefitMonths { get; set; }
```
ryanhofdotgov, line 28: I think one way to have a strongly typed approach instead of just string would be to use something like what is described here:
- https://josef.codes/csvhelper-read-column-value-into-a-list-property/
- https://stackoverflow.com/questions/60966194/using-csvhelper-to-read-cell-content-into-list-or-array

etl/src/Piipan.Etl/BulkUpload.cs
echappen: These are the first of the month so I'll change them to the last to be more consistent.

etl/docs/csv/import-schema.json
echappen: It will not. I added a few more examples of <3 dates to the example csv.

ryanhofdotgov: Hmm... something isn't quite right, but not sure if it is in the schema, the tableschema code, or our validate.py. 

For example, try adding a bogus string like `xxyy` or `2021` to a `recent_benefit_months` cell in the CSV. Expectation is that it should throw an error, but it is silent.

echappen: Thanks for checking, I can look into it.

etl/src/Piipan.Etl/PiiRecord.cs
echappen: Those posts were a bit outdated, but they led me to figuring out how to use a version of [this recommendation](https://joshclose.github.io/CsvHelper/examples/configuration/class-maps/type-conversion/) for custom type conversion from the docs.

etl/docs/csv/import-schema.json
echappen: It seems using type `string` with a date-like formatting wasn't working. I switched to using a regex pattern constraint, which will catch non-dates but still allows scenarios like fewer than 3 dates and null values. The regex syntax is a little weird—for example, I had to put a real whitespace instead of `\s?`

etl/docs/csv/import-schema.json
```diff
@@ -63,6 +63,14 @@
         "name": "benefits_end_month",
         "type": "yearmonth",
         "description": "The final month when a participant will receive benefits. Leave this field blank if the participant will continue receiving benefits until the next recertification period."
+      },
+      {
+        "name": "recent_benefit_months",
+        "type": "string",
+        "description": "List of up to the last 3 months that participant received benefits, formatted as yearmonths separated by spaces. Does not include current benefit month. Months need to be consecutive. Fewer than 3 months can be entered for newer participants.",
+        "constraints": {
+          "pattern": "([0-9]{4}-[0-9]{2} ?){0,3}"
```
ryanhofdotgov, line 72: I think since you are accepting "yyyy-M" in C# below, we should be consistent and accept that here as well.

ryanhofdotgov: (Or alternatively, strictly accept "yyyy-MM" in both places)

etl/src/Piipan.Etl/BulkUpload.cs
echappen: I added a Helper class to `Piipan.Shared` that I thought could starting housing DateTime parsing and conversions for month-only dates. Would like to eventually use it for BenefitsEndMonth as well. Let me know if this is on the right track.

ryanhofdotgov: That LGTM – thanks!

echappen:
@ryanhofdotgov I created another converter for `benefits_end_month` that uses the new `MonthEndDateTime` Helper. Also added similar DateTime validation. This is ready for a final review. Once approved I'll run iac on tts/dev to update the database, and then merge once that's successful.

etl/src/Piipan.Etl/Converters.cs
```diff
@@ -0,0 +1,41 @@
...
+using System.Linq;
+using System.Collections.Generic;
+using CsvHelper;
+using CsvHelper.Configuration;
+using CsvHelper.TypeConversion;
+using Piipan.Shared.Helpers;
+
+namespace Piipan.Etl
+{
+	public class ToDateTimeConverter : DefaultTypeConverter
```
ryanhofdotgov, line 11: Nitpick: The name is a bit generic for the very specific last day of month from partial date conversion that it does. I'd be inclined to rename or have comment block for these 2 classes here. Perhaps `MonthEndConverter` and `YearMonthConverter` or something like that.

We should define some minimal guidelines across the repo, but a short comment block for the classes/methods is probably warranted too. Generally, all public classes/all public methods is a good guideline, but I think our confusion about how to reference methods in external test assemblies has muddied that.

etl/src/Piipan.Etl/Converters.cs
```diff
@@ -0,0 +1,41 @@
...
+	{
+		public override object ConvertFromString(string text, IReaderRow row, MemberMapData memberMapData)
+		{
+			if (String.IsNullOrEmpty(text)) return null;
+            return MonthEndDateTime.Parse(text);
```
ryanhofdotgov, line 16: Nitpick: Looks like maybe some of the files in this PR have not been [formatted by omnisharp](https://github.com/18F/piipan/blob/1284f25851be575128655324e51af92ac8bff7a1/docs/source-guidelines.md#source-code-conventions)?

ryanhofdotgov (approved):
I left a couple of nitpicks, but we can move forward with the PR as-is and address in a follow-up if you like. Thanks!

PR #1015, opened 2021-06-07
Stream logs for metrics Functions to Event Hub

kingcomma:
Uses existing CLI-based configuration approach.

There has been some change in the available log categories over the past several days, and this work may need to be updated to include new/changed categories at some point in the future.

Partially addresses #969.

### DIFF
--- a/iac/create-metrics-resources.bash
+++ b/iac/create-metrics-resources.bash
@@ -147,6 +147,32 @@ EOF
     --subnet "$FUNC_SUBNET_NAME" \
     --vnet "$VNET_NAME"
 
+  # Configure log streaming for function app
+  metrics_collect_function_id=$(\
+    az functionapp show \
+      -n "$METRICS_COLLECT_APP_NAME" \
+      -g "$METRICS_RESOURCE_GROUP" \
+      -o tsv \
+      --query id)
+  hub_rule_id=$(\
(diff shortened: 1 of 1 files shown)


iac/create-metrics-resources.bash
```diff
@@ -147,6 +147,32 @@ EOF
...
+    az functionapp show \
+      -n "$METRICS_COLLECT_APP_NAME" \
+      -g "$METRICS_RESOURCE_GROUP" \
+      -o tsv \
+      --query id)
+  hub_rule_id=$(\
+    az eventhubs namespace authorization-rule list \
+      --resource-group "$RESOURCE_GROUP" \
+      --namespace-name "$EVENT_HUB_NAME" \
+      --query "[?name == 'RootManageSharedAccessKey'].id" \
```
ryanhofdotgov, line 161: Thought: seeing this `RootManageSharedAccessKey` here and in the portal the other day makes me think the project should have a short doc of the big picture on how the logging is intended to work here. In the same vein as your authentication doc, but probably much less in-depth.

ryanhofdotgov (approved):
LGTM

iac/create-metrics-resources.bash
kingcomma: Added this as a task #1016

PR #1023, opened 2021-06-08
Return recent benefit months in Match query responses

### LINKED ISSUE #897: Query API returns the most recent benefit months

echappen:
Closes #897 

Also addresses some naming items from [previous PR](https://github.com/18F/piipan/pull/988#pullrequestreview-677746071)

### Notes:
- Expands `Piipan.Shared` to other projects
- Adds new `Piipan.Match.Shared` project to reduce duplication of Json converter code
- @timoballard I tried using `IEnumerable` for `RecentBenefitMonths` in the State `MatchResponse` but couldn't get around the error `Cannot create an instance of the abstract class or interface 'IEnumerable<DateTime>'`. I figured I'm probably missing something there that can be refactored in a future PR. 

iac won't need to be run for this PR, but I'll deploy the match subsystem to `tts/dev` before merging.


ryanhofdotgov (approved):
LGTM – thanks for the needed refactor and for drilling down into these details that make internal date handling consistent 👍

echappen:
match subsystem has been deployed to `tts/dev` so I shall merge!

ryanhofdotgov:
@echappen Since the .NET code itself will be automatically deployed once the PR is closed, there's no strict need to manually deploy ahead of time. With the IaC stuff it's a bit different since we have no automation set up for that.

PR #1028, opened 2021-06-09
Enable Easy Auth for Query Tool, Dashboard

### LINKED ISSUE #169: OpenID Authentication for query tool
As a security officer, I want users to prove their identity when accessing the NAC so that nobody is able to use the NAC for unauthorized reasons.

At this time, we can treat the access groups as mutually exclusive: users with access to the query tool / lookup IDs will not also have access to the metrics app.  Users with access to the metrics app will not have access to the query tool.

- [ ] Users must login through an OpenID provider before gaining access to the NAC query tool form or results
- [ ] Attempts to access Query tool without credentials returns 403 access denied
- [ ] Granting a user access to the query tool does not automatically grant them access to the metrics app, and vice versa

### LINKED ISSUE #276: OpenID (eAuth) integration for metrics website
As a FNS Business Owner, I want to know that only authorized users are able to access the metrics website for the NAC.

All authorized users for the metrics app will be at FNS, and we anticipate using eAuth for this purpose.

At this time, we can treat the access groups as mutually exclusive: users with access to the query tool / lookup IDs will not also have access to the metrics app. Users with access to the metrics app will not have access to the query tool.

AC:
- [ ] Access to the website is verified through an OpenID provider
- [ ] Attempts to access the website without authenticating return a 403 Access Denied error
- [ ] Granting a user access to the metrics website does not automatically grant them access to the query tool and vice versa.

timoballard:
Resolves #169 
Resolves #276

Updates IaC to configure Easy Auth for the Query Tool and Dashboard. Temporarily disables Front Door for the two apps. OIDC provider for dev/test is an Azure B2C Tenant which is not configured as part of IaC ([see here for B2C setup details](https://github.com/18F/piipan/blob/dev/docs/create-ad-b2c-idp.md))


iac/env/tts/dev.bash
```diff
@@ -23,3 +23,8 @@ CLOUD_NAME=AzureCloud
 
 # used to create API Management resources
 APIM_EMAIL=noreply@tts.test
+
+OIDC_ISSUER_URI="https://ttsb2cdev.b2clogin.com/ttsb2cdev.onmicrosoft.com/v2.0/.well-known/openid-configuration?p=B2C_1_si"
+
+DASHBOARD_APP_CLIENT_ID=e7e769ad-e9bc-4c5f-8c3e-ebaf6cf9cacb
+QUERY_TOOL_APP_CLIENT_ID=0c3baef3-c2f5-41b8-b62a-eb0083e8771f
```
timoballard, line 30: I don't _love_ this - suggestions for a better approach to managing/specifying app client IDs in IaC - given that app registration isn't automated - are welcome!

iac/arm-templates/dashboard-app.json
```diff
@@ -143,6 +130,33 @@
...
+            "properties": {
+                "platform": {
+                    "enabled": true
+                },
+                "globalValidation": {
+                    "requireAuthentication": true,
+                    "unauthenticatedClientAction": "RedirectToLoginPage"
+                },
+                "identityProviders": {
+                    "azureActiveDirectory": {
```
ryanhofdotgov, line 151: I think what we actually need is the OIDC identity provider (not the Azure AD one) – here are the OIDC-specific details:
https://docs.microsoft.com/en-us/azure/app-service/configure-authentication-provider-openid-connect

We will need to manage the IdP client secret somewhere in a key vault I suspect and have the ARM template read it out of there. May have to do a 2 stage process, where we just set a random value in IaC, then manually update it and run the ARM template again.

ryanhofdotgov (approved):
Looks good 👍

Notes from our convo:
- Automated mechanism to set up a secret in Key Vault and set App Service environment variable from that will come in a separate PR. Tracking task in https://github.com/18F/piipan/issues/1039

- We have to manually set IDP_CLIENT_SECRET in App Service environment variable for both dashboard and query tool after every run of IaC until automation is established

PR #1032, opened 2021-06-10
Add protect_location field to bulk upload import

### LINKED ISSUE #850: Clarify how to share vulnerability status in bulk uploads
As a state agency IT worker, I want clear guidance to specify an individual's status as vulnerable so that I know how to send information to the NAC.  I also want the data to be stored in a standard way across all states so that I know how to make sense of data that other states uploaded.

We are defining "vulnerable" as a person who would be put at risk by having their location information exposed.

AC:
- [ ] Create a "protect location" field that will accept an individual's vulnerability status
- [ ] Include yes/no/unknown options
- [ ] "Protect location" field is optional - if a row is missing that field, the upload is still accepted
- [ ] Update documentation with guidance on the "Protect location" field (but note that final rule is still pending)

echappen:
Closes #850 

## Notes
- I was unable to get a mishandled Boolean field to throw a Validation Exception. They seem to be caught at the conversion level. After some googling, I [found](https://stackoverflow.com/a/22998705) that a custom validator may be the only way to catch validation errors with booleans. A custom validator didn't seem necessary at this time, since acceptable values for true and false are provided [in other ways](https://joshclose.github.io/CsvHelper/examples/configuration/attributes/). 

### Adds
- Adds optional `protect_location` boolean field to Bulk Upload API 
- Adds `protect_location` column to example upload csv

## Deploy
- Since database schema is changed, I'll merge once iac has been run on tts/dev for these changes.

### DIFF
--- a/etl/docs/csv/import-schema.json
+++ b/etl/docs/csv/import-schema.json
@@ -71,6 +71,23 @@
         "constraints": {
           "pattern": "([0-9]{4}-[0-9]{1,2} ?){0,3}"
         }
+      },
+      {
+        "name": "protect_location",
+        "type": "boolean",
+        "description": "Set true to indicate location protection for vulnerable individuals. Null or empty values will be treated as true throughout the system.",
+        "trueValues": [
+          "true",
+          "True",
+          "TRUE",
+          "1"
+        ],
+        "falseValues": [
+          "false",
+          "False",
+          "FALSE",
+          "0"
+        ]
       }
     ]
   }
--- a/etl/docs/csv/import-schema.json
+++ b/etl/docs/csv/import-schema.json
@@ -77,16 +77,10 @@
         "type": "boolean",
         "description": "Set true to indicate location protection for vulnerable individuals. Null or empty values will be treated as true throughout the system.",
         "trueValues": [
-          "true",
-          "True",
-          "TRUE",
-          "1"
+          "true"
         ],
         "falseValues": [
-          "false",
-          "False",
-          "FALSE",
-          "0"
+          "false"
         ]
       }
     ]
--- a/etl/tests/Piipan.Etl.Tests/BulkUploadTests.cs
+++ b/etl/tests/Piipan.Etl.Tests/BulkUploadTests.cs
@@ -17,7 +17,7 @@ namespace Piipan.Etl.Tests
             var writer = new StreamWriter(stream);
             if (includeHeader)
             {
-                writer.WriteLine("last,first,middle,dob,ssn,exception,case_id,participant_id,benefits_end_month,recent_benefit_months");
+                writer.WriteLine("last,first,middle,dob,ssn,exception,case_id,participant_id,benefits_end_month,recent_benefit_months,protect_location");
             }
             foreach (var record in records)
             {
@@ -57,7 +57,8 @@ namespace Piipan.Etl.Tests
                   new DateTime(2021, 5, 31),
                   new DateTime(2021, 4, 30),
                   new DateTime(2021, 3, 31)
-                }
+                },
+                ProtectLocation = true
             };
         }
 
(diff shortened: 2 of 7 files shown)


ryanhofdotgov:
@echappen To clarify, when you say "I was unable to get a mishandled Boolean field to throw a Validation Exception", you're just talking about that specific type of exception type – it still throws an Exception, but it's a TypeConverterException. Kind of how invalid dates will throw a ReaderException.

etl/docs/csv/import-schema.json
```diff
@@ -71,6 +71,23 @@
         "constraints": {
           "pattern": "([0-9]{4}-[0-9]{1,2} ?){0,3}"
         }
+      },
+      {
+        "name": "protect_location",
```
ryanhofdotgov, line 76: We should limit the values to just `true` and `false`. This simplifies it and is a defensive tactic to limit exposure to implementation quirks on both sides of the client/server equation down the line.

In same vein, I double-checked what [ISO 8601 says about dates](https://en.wikipedia.org/wiki/ISO_8601), and has a principle of:
"Each date and time value has a fixed number of digits that must be padded with leading zeros."
So when we validate `recent_benefit_months` we should just stick with `([0-9]{4}-[0-9]{2} ?){0,3}`, instead of making zero padding optional. I'll open up a ticket to tidy that up.

etl/tests/Piipan.Etl.Tests/BulkUploadTests.cs
```diff
@@ -164,6 +167,23 @@ public void ExpectFieldValidationError(String inline)
...
+            var records = BulkUpload.Read(stream, logger);
+            Assert.Throws<CsvHelper.TypeConversion.TypeConverterException>(() =>
+            {
+                foreach (var record in records)
+                {
+                    ;
+                }
+            });
+        }
+
```
ryanhofdotgov, line 186: One thing that I missed when originally writing these tests was to test when an optional CSV field was completely omitted from the CSV input entirely; i.e., the named column doesn't exist.

Would you be able to add a test that checks that in the context of the `protect_location` field? You'll probably want to create a new CsvFixureXXX method that takes [named arguments](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/named-and-optional-arguments)

ryanhofdotgov (commented):
Thank you for the change log section for this PR! 🥇

echappen:
@ryanhofdotgov Correct—I didn't know if we were valuing certain exceptions over others.

etl/tests/Piipan.Etl.Tests/BulkUploadTests.cs
echappen: Correct—I didn't know if we were valuing certain exceptions over others.

echappen:
> when you say "I was unable to get a mishandled Boolean field to throw a Validation Exception", you're just talking about that specific type of exception type? 

@ryanhofdotgov Correct—I didn't know if we were valuing certain exception types over others.

ryanhofdotgov:
> > when you say "I was unable to get a mishandled Boolean field to throw a Validation Exception", you're just talking about that specific type of exception type?
> 
> @ryanhofdotgov Correct—I didn't know if we were valuing certain exception types over others.

Cool – yeah, I think CSVHelper would be marginally better if it had a consistent "exception interface", but since I don't think we can really recover differently based on the underlying exception type, any exception at all seems A-OK.

etl/tests/Piipan.Etl.Tests/BulkUploadTests.cs
echappen: [Added](https://github.com/18F/piipan/pull/1032/commits/8d9da965ca00c855d2011f5e66c6e1f0e029901f#diff-a1bc982406a22fce11811c15276f30cce6e1c191aabbcabc4cea6b503bb7f4daR207)

ryanhofdotgov: Ah – the new test you added is a test that isn't being done, but not the one I had in mind.

The CSV header has named columns (e.g,. `last,first,middle...`); the existing `CsvFixture` method sets it for every fixture it generates. As we start adding all these optional columns, it's also allowable for the CSV to not have the named column in its header at all. Right now for the most part, our tests assume that the header is correct as we are mostly testing at how we process the values.

For example:
At time 0: states send a CSV with required columns A,B,C
At time 1: we've deployed code that supports new optional column D
At time 2: states send a CSV with only required columns A,B,C – this upload should still be successful

But no need to hold up the PR. It's been a while since a wrote much code for the project, so let me take a stab at the test I had in mind to verify the time 2 scenario works.

ryanhofdotgov (approved):

PR #1033, opened 2021-06-10
Add protect_location to match query api response

### LINKED ISSUE #989: Add protect location from matching state to Query API responses
As a SNAP applicant who could be harmed by the disclosure of my location, I want this vulnerability to be shared when states are notified about potential duplicate participation so that matching states treat my location with the appropriate secrecy.

Note: scope of this issue is to pass through the protect location info shared in the matching state's bulk uploads.  Capturing the status at the querying state has been separated into issue #1040

AC:

- [ ] Query API includes an optional "protect location" field (should be named consistently with the bulk upload field name from #850)
- [ ] Field accepts values of yes/no
- [ ] API documentation is updated, including requirement that response values that are unspecified need to be treated as if the response was yes

echappen:
Closes #989 

### Adds
- Adds protect_location field to matches in Duplicate Participation query responses

### DIFF
--- a/docs/openapi/generated/duplicate-participation-api/openapi.md
+++ b/docs/openapi/generated/duplicate-participation-api/openapi.md
@@ -164,7 +164,7 @@ Status Code **200**
 |»» participant_id|string|false|none|Participant's state-specific identifier. Must not be social security number or any personal identifiable information.|
 |»» benefits_end_month|string|false|none|Participant's ending benefits month|
 |»» recent_benefit_months|[string]|false|none|List of up to the last 3 months that participant received benefits, in descending order. Each month is formatted as ISO 8601 year and month. Does not include current benefit month.|
-|»» protect_location|boolean¦null|false|none|Location protection flag for vulnerable individuals. Null values are treated as true, and individual's location info will not be shared with matching state.|
+|»» protect_location|boolean¦null|false|none|Location protection flag for vulnerable individuals. True values indicate that the individual’s location must be protected from disclosure to avoid harm to the individual. Apply the same protections to true and null values.|
 
 <aside class="warning">
 To perform this operation, you must be authenticated by means of one of the following methods:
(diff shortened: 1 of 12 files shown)


docs/openapi/generated/duplicate-participation-api/openapi.md
```diff
@@ -161,6 +164,7 @@ Status Code **200**
 |»» participant_id|string|false|none|Participant's state-specific identifier. Must not be social security number or any personal identifiable information.|
 |»» benefits_end_month|string|false|none|Participant's ending benefits month|
 |»» recent_benefit_months|[string]|false|none|List of up to the last 3 months that participant received benefits, in descending order. Each month is formatted as ISO 8601 year and month. Does not include current benefit month.|
+|»» protect_location|boolean¦null|false|none|Indicates location protection for vulnerable individuals. Null values provided to the system are treated as true.|
```
ryanhofdotgov, line 167: I think "Indicates location protection for vulnerable individuals" could use some wordsmithing with @erinzstrenio and/or @cjhskippy.

Do we actually treat null values as true? I don't think the code does that, which I think is what we want. Here's @cjhskippy's comment on that; there's a nuance between what we store/return at API level and what we would do in the query tool UI:
https://github.com/18F/piipan/issues/850#issuecomment-854228103

erinzstrenio: How about: "Location protection flag for vulnerable individuals. Null values are treated as true, and participant's location info will not be shared with matching state [_or whatever the implications are if the flag will trigger workflow changes_]."

echappen: My understanding is the API's don't need to act on the `protect_location` field for now, but we'd display null values as yes's (or something other than null) in the Query Tool.

ryanhofdotgov: ^ Can we remove the "Null values..." bit from the API docs as that's a UX product decision for the query tool, rather than part of our "contract" with the state systems?

ryanhofdotgov (approved):

docs/openapi/generated/duplicate-participation-api/openapi.md
erinzstrenio: Revised language per team discussion in Slack: 
"Location protection flag for vulnerable individuals. True values indicate that the individual’s location must be protected from disclosure to avoid harm to the individual. Apply the same protections to true and null values."

PR #1035, opened 2021-06-11
Specify CSV fields that are optional

ryanhofdotgov:
This permits the CSV header to be omitted from CSV file for optional fields, which will be important for preserving backwards compatibility.

Consistency should be maintained between optional fields in import-schema.json, which C# members are marked as nullable, and what maps are Optional.

*Note*: we want validate.py to still look for – and fail – if all of our currently defined columns are not present, as we want devs to move towards all the current columns. May have to revisit once the API has reached "1.0" status and support validation of previous versions.

### Adds
- For backwards compatibility, columns, not just field values, are optional when CSV schema does not list them as required. validator.py will still fail if all defined columns are not present.

### DIFF
--- a/etl/src/Piipan.Etl/PiiRecordMap.cs
+++ b/etl/src/Piipan.Etl/PiiRecordMap.cs
@@ -21,10 +21,10 @@ namespace Piipan.Etl
             });
 
             Map(m => m.First).Name("first")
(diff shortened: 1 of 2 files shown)


echappen (approved):
Thanks for doing this!

PR #1041, opened 2021-06-11
Wordsmithing protect location field description

ryanhofdotgov:
Should generally match query response API description


echappen (approved):
👍

PR #1054, opened 2021-06-14
Log managed identity when executing match API requests

### LINKED ISSUE #970: Stream managed identity logs to an event hub
Azure Active Directory (needs research — assumes managed identity usage logging happens here)

kingcomma:
Covers all HTTP triggered Functions with EasyAuth enabled:

- State query call
- Orchestrator query call
- Orchestrator lookup_ids call

The log message displays as `Executing request from user (null)` in a couple contexts:
- If `HttpRequest.HttpContext.User` is not set, as is the case in local development where the Function does not sit behind EasyAuth
- If `HttpRequest` does not contain an `HttpContext` object, as is the case when running tests with a mocked `HttpRequest`

Partially addresses #970.

### DIFF
--- a/match/src/Piipan.Match.Orchestrator/Api.cs
+++ b/match/src/Piipan.Match.Orchestrator/Api.cs
@@ -42,6 +42,8 @@ namespace Piipan.Match.Orchestrator
             [HttpTrigger(AuthorizationLevel.Anonymous, "post", Route = null)] HttpRequest req,
             ILogger log)
         {
+            log.LogInformation("Executing request from user {User}", req.HttpContext?.User.Identity.Name);
+
             var incoming = await new StreamReader(req.Body).ReadToEndAsync();
             var request = Parse(incoming, log);
             if (request.Query == null)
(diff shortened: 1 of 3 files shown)


ryanhofdotgov:
@kingcomma I don't think we've talked about Easy Auth in the same context as managed identities before – are managed identities implemented under the hood with Easy Auth?

ryanhofdotgov (approved):
LGTM, but it is missing the Function Apps from the Metrics subsystem, so doesn't completely close #970

kingcomma:
@ryanhofdotgov When EasyAuth is in play (e.g., when the orchestrator us calling the state APIs), the "managed identity" is the system-assigned identity of the client resource (aka, the [service principal](https://github.com/18F/piipan/blob/dev/docs/securing-internal-apis.md#configuring-the-client)). Authentication of the identity by EasyAuth happens under the hood and the HTTP request that reaches the Function contains data about the identity.

kingcomma:
On the topic of metrics Functions, as currently configured they do not use any authentication method and thus the log message would always display as `Executing request from user (null)`. Is the plan to implement some form of authentication in the future? Happy to fold those functions into this PR either way.

ryanhofdotgov:
@kingcomma Let's fold those log messages in to this PR and fix the auth for those Function Apps in the upcoming sprint.

kingcomma:
@ryanhofdotgov I added logging to `Piipan.Metrics.Api`. I did not add logging to `Piipan.Metrics.Collect`. We're currently using Event Grid triggers there, but that might change with #1055. I've updated that ticket to include logging once the final implementation is complete.

I've also clarified that this PR only partially addresses #970.

I think that covers everything.

ryanhofdotgov (approved):
LGTM

PR #1058, opened 2021-06-14
Add recent benefit months and protect location to query tool results

### LINKED ISSUE #851: Share vulnerability status in FNS website queries / responses
As a SNAP applicant who could be harmed by the disclosure of my location, I want this vulnerability to be shared when states are notified about potential duplicate participation so that matching states treat my location with the appropriate secrecy.

AC:
- [ ] Query website form includes a yes/no field for whether the applicant is vulnerable.
  - [ ] Field should not have a default value - it should be explicitly set
  - [ ] Vulnerable status is an optional field.
- [ ] Query website matches display the vulnerability status provided by the matching state, translated from the stored boolean value in #850:
  - [ ] true is displayed as "Yes"
  - [ ] false is displayed as "No"
  - [ ] empty / null is displayed as "Yes" (not a typo - workers need to treat uncertain responses identically to yes responses)

### LINKED ISSUE #869: Add recent benefit months to query website
Update the query website to include the same data provided in #861

AC:
When a match is found using the query website:
- [ ] match results include the same recent benefit months shared in the bulk upload.

echappen:
Closes #869 
Closes #851 

- null `protect_location` values will show up as "Yes"; true as "Yes"; false as "No"
- No `recent_benefit_months` is shown as an empty column

@cjhskippy @erinzstrenio See image below. Thoughts on wording for these new columns?

Also, the results table is getting pretty scrunched. A common way to handle that is by giving the table a fixed width and making it horizontally scrollable. This results in some of the trailing data being hidden unless the user scrolls. I'd welcome any ideas on the design of that, or we can hold off until Julia gets back. 

Screen Shot 2021-06-14 at 2 07 53 PM

### DIFF
--- a/query-tool/src/Piipan.QueryTool/PiiRecord.cs
+++ b/query-tool/src/Piipan.QueryTool/PiiRecord.cs
@@ -59,5 +59,27 @@ namespace Piipan.QueryTool
         [Display(Name = "Benefits End Month")]
         [JsonPropertyName("benefits_end_month")]
         public string BenefitsEndMonth { get; set; }
+
+        [Display(Name = "Recent Benefit Months")]
+        [JsonPropertyName("recent_benefit_months")]
+        public string[] RecentBenefitMonths { get; set; } = new string[0];
+
+        public string RecentBenefitMonthsDisplay
+        {
+            get { return String.Join(", ", this.RecentBenefitMonths); }
+        }
+
+        [Display(Name = "Protect Location")]
+        [JsonPropertyName("protect_location")]
+        public bool? ProtectLocation { get; set; }
+
+        public string ProtectLocationDisplay
+        {
+            get
+            {
+                if (this.ProtectLocation == null) return "Yes";
+                return (bool)this.ProtectLocation ? "Yes" : "No";
+            }
+        }
     }
 }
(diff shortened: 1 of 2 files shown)


cjhskippy:
Thanks Eleni!  "Protect Location" is the best naming I can come up with in the absence of real usability testing.  I'd avoid trying to tweak the display for now.  It seems likely to me that we will need to make some more drastic changes in the not-too-distant future.

erinzstrenio:
Some edit suggestions: 

Benefits End —> Benefits end month
Recent Benefit Months —> Recent benefits months
Protect Location —> Protected location for vulnerable individuals

The last one might be too wordy? Also debated on whether it should say "most recent benefits month" instead of just "recent..."

kingcomma:
On the UI question about handling the table size, the [UX/UI recs doc](https://docs.google.com/document/d/1vdjSyvKMzKIqyZ0JM3QAlv_3X3NfLSPjGwVIi2D00y8/edit#) has a card-based example that probably solves this long-term.

query-tool/src/Piipan.QueryTool/PiiRecord.cs
```diff
@@ -59,5 +59,27 @@ public class PiiRecord : IQueryable
         [Display(Name = "Benefits End Month")]
         [JsonPropertyName("benefits_end_month")]
         public string BenefitsEndMonth { get; set; }
+
+        [Display(Name = "Recent Benefit Months")]
+        [JsonPropertyName("recent_benefit_months")]
+        public string[] RecentBenefitMonths { get; set; } = new string[0];
```
ryanhofdotgov, line 65: Not for this PR, or perhaps not ever, but it would be neat to see if we could universally apply the `DisplayFormat` mechanism to all the fields, so we separate the data type from the display format. While we want to use ISO 8601 for all the APIs, there may be a better display representation for humans.

ryanhofdotgov (approved):
LGTM, aside from the tweaks that @erinzstrenio has suggested.

PR #1072, opened 2021-06-15
Create Event Hub log receiver and assign role in IaC

### LINKED ISSUE #1064: Add IaC mechanism to grant access to Event Hub log receiver
- [ ] Create app registration for an external log reader account if does not exist
- [ ] Explicitly grant Azure Event Hubs Data Receiver role on the event hub instance
- [ ] Name of reader account needs to be environment specific

Reference:
https://docs.microsoft.com/en-us/azure/event-hubs/authenticate-application

kingcomma:
- Add env-specific `$SIEM_RECEIVER` variable to tts/dev and tts/test
- Create app registration if it does not already exist (avoid resetting password on existing registration)
- Assign relevant role to receiver in event hub ARM template

Closes #1064

### DIFF
--- a/iac/arm-templates/event-hub-monitoring.json
+++ b/iac/arm-templates/event-hub-monitoring.json
@@ -23,7 +23,7 @@
         "namespace": "[concat(parameters('prefix'), '-evh-', variables('basename'), '-', parameters('env'))]",
         // Azure Event Hubs Data Receiver
         "receiverRoleId": "a638d3c7-ab3a-418d-83e6-5f17a39d4fde",
-        "receiverRoleName": "[guid(parameters('receiverId'), variables('roleId'), resourceGroup().id)]"
+        "receiverRoleName": "[guid(parameters('receiverId'), variables('receiverRoleId'), resourceGroup().id)]"
     },
     "resources": [
         {
--- a/iac/arm-templates/event-hub-monitoring.json
+++ b/iac/arm-templates/event-hub-monitoring.json
@@ -8,6 +8,9 @@
         "prefix": {
             "type": "String"
         },
+        "receiverId": {
+            "type": "String"
+        },
         "resourceTags": {
             "type": "object"
         },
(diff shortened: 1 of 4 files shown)


iac/arm-templates/event-hub-monitoring.json
```diff
@@ -44,7 +50,23 @@
                 "messageRetentionInDays": 1,
                 "partitionCount": 1,
                 "status": "Active"
-            }
+            },
+            "resources": [
+                {
+                    "type": "Microsoft.EventHub/namespaces/eventhubs/providers/roleAssignments",
+                    "apiVersion": "2020-04-01-preview",
```
ryanhofdotgov, line 57: @kingcomma Can you use the Resource Explorer in USGovernmentCloud per FastTrack's recent recommendation to see if this version is available there?

Would be cool to verify that is a usable approach, rather than just running it there and seeing if it succeeds/fails.

ryanhofdotgov (approved):

ryanhofdotgov:
@kingcomma Would you also drop a comment here after you've run the updated IaC (or relevant portion of the IaC) against `fns/dev`?

iac/arm-templates/event-hub-monitoring.json
kingcomma: Absolutely! I double checked this but didn't mention it in the initial PR comment. Resource Explorer was super helpful.

The only thing that's a little wonky here is this role assignment is a sub-resource with a type of `Microsoft.EventHub/namespaces/eventhubs/providers/roleAssignments`. As I understand it, that correlates to a top-level resource type of `Microsoft.Authorization/roleAssignments` in Resource Explorer.

kingcomma:
😞 Running this prompted a couple naming issues that I didn't catch after a refactor. Corrected those with d4478ac.

As for the role assignment, an interesting issue popped up. Role assignments have a `name` property that must be a valid GUID. ARM templates have a `guid` function that generates a stable GUID based on passed parameters. This works great if the role assignment is only ever created by the ARM template (i.e., with the same GUID). If the role assignment is first created via a different method or with a different GUID, we'll get an error:

```
Deployment failed. Correlation ID: --. {
  "error": {
    "code": "RoleAssignmentExists",
    "message": "The role assignment already exists."
  }
}
```

I had to remove the existing role assignment (which was created via the Portal on our call with FastTrack) before the updated IaC would work. It now works consistently and repeatably.

ryanhofdotgov (approved):

PR #1075, opened 2021-06-15
Clear app firewall rules for dashboard and query tool

ryanhofdotgov:
Removing the ipSecurityRestrictions property in ARM template does not clear any existing firewall rules, but leaves it unchanged. Need to explicitly set the property to an empty array.

Firewall rules can be cleared since Easy Auth is enabled since 622b1a9b17b2fb63abf366f282130ed97587b91d


kingcomma (approved):
👍

Issue #368, opened 2021-02-18
Allow states to query for households in one API call


ryanhofdotgov:
As a state system integrator, I want to be able to submit multiple PII records in one query to the NAC. This would allow me to more easily process entire households without having to perform multiple web service API calls.

Implementation notes:
* small-scale, household level "bulk" queries for near-real time responses (max household size should be further researched, but likely maxes around 50.  )

AC:
- [ ] Queries can accept up to 50 PII records (our estimate for a maximally-large household)
- [ ] Query responses include the results for every person in the query
- [ ] (optional) Measure and report on a typical response time for queries with 50 PII records

Issue was originally paired with #865 (support overnight processing of recertifications in a single API call)

ryanhofdotgov:
Needs to be split up

cjhskippy:
Spun off additional issue of #865 (batch queries for the purpose of overnight recertifications)

echappen:
> Query responses include the results for every person in the query

Does this mean:
1. There's a list of matches for every person in the query?
1. There's a Lookup ID for every person in the query?

cc @ryanhofdotgov @cjhskippy

echappen:
And by "person" I mean for each unique PII object in the query

kingcomma:
Conceptually, I had been thinking of this as the same approach we currently take with the only difference being the orchestrator running multiple match queries rather than a single query. In that case, the orchestrator would return an array of match response objects (each with their own `lookup_id` and `matches` properties).

I've also been wondering if we should just default to accepting an array for both requests and responses, whether the request contains a single applicant or multiple household members? The trade off there is a bigger "break" but perhaps a better experience if state systems don't have to conditionally check for an object or an array.

ryanhofdotgov:
+1 to breaking things now and using an array for both query and response to reduce variations in how the API is used.

ryanhofdotgov:
@echappen I would do this work in 2 steps – update the spec, give us an opportunity to look at it for a day or two, then do the implementation work.

PR #1082, opened 2021-06-16
Change match query API schema to accept and return arrays

echappen:
Satisfies schema/documentation changes only for #368 

I found it easiest to review the changes by viewing the `docs/openapi/generated/duplicate-participation-api/openapi.md` file for this branch. 

Summarizing the different areas we talked about:

Switching to a Bulk API means:
1. we need both item-level and top-level error handling
1. we need to properly relate people requested to their results
1. metadata may become more important 
1. parameter and property names will change

All these things pushed us to think about standardizing our response schemas across our APIs. We're considering [JSON API](https://github.com/18F/piipan/pull/1119) for this purpose. 

### Error handling
See the schema for how top-level and item-level errors are handled. 

For item-level errors, we're using Amazon's Bulk API approach to having either a result or errors object for each person in the response. 

Note that JSON API has either a `data` or `errors` property in its top-level response, but not both.

### Relating People to their Results
Also takes the Amazon approach of using an index for each result and errors object. See the schema for more details on index.

### Metadata
JSON API has an optional top-level `meta` property. Since we can't think of useful metadata for this endpoint at the moment, we're going to leave it off of this response. 

### Naming
Apart from the top-level JSON API spec names, we decided to be as semantic as we could with API naming so it's easier for states to understand. The PII is representing persons, so that's what we call them in the API. 

Since using JSON API spec names for responses but not requests was a bit illogical, I've conformed the request body to use `data` instead of `persons`. I also updated the proposed [ADR](https://github.com/18F/piipan/pull/1119/files) to reflect that.

### DIFF
--- a/docs/openapi/generated/duplicate-participation-api/openapi.md
+++ b/docs/openapi/generated/duplicate-participation-api/openapi.md
@@ -268,8 +268,8 @@ Status Code **200**
 |»»» state|string|false|none|State/territory two-letter postal abbreviation|
 |»»» state_abbr|string|false|none|State/territory two-letter postal abbreviation. Deprecated, superseded by `state`.|
 |»»» exception|string|false|none|Placeholder for value indicating special processing instructions|
-|»»» case_id|string|false|none|Participant's state-specific case identifier|
-|»»» participant_id|string|false|none|Participant's state-specific identifier. Must not be social security number or any personal identifiable information.|
+|»»» case_id|string|false|none|Participant's state-specific case identifier. Can be the same for multiple participants.|
+|»»» participant_id|string|false|none|Participant's state-specific identifier. Is unique to the participant. Must not be social security number or any PII.|
 |»»» benefits_end_month|string|false|none|Participant's ending benefits month|
 |»»» recent_benefit_months|[string]|false|none|List of up to the last 3 months that participant received benefits, in descending order. Each month is formatted as ISO 8601 year and month. Does not include current benefit month.|
 |»»» protect_location|boolean¦null|false|none|Location protection flag for vulnerable individuals. True values indicate that the individual’s location must be protected from disclosure to avoid harm to the individual. Apply the same protections to true and null values.|
--- a/docs/openapi/generated/duplicate-participation-api/openapi.md
+++ b/docs/openapi/generated/duplicate-participation-api/openapi.md
@@ -70,6 +70,7 @@ Queries all state databases for any PII records that are an exact match to the l
 {
   "data": [
     {
+      "index": 0,
       "lookup_id": "string",
       "matches": [
         {
@@ -103,6 +104,7 @@ Queries all state databases for any PII records that are an exact match to the l
 {
   "data": [
     {
+      "index": 0,
       "lookup_id": null,
       "matches": []
     }
@@ -116,6 +118,7 @@ Queries all state databases for any PII records that are an exact match to the l
 {
   "data": [
     {
+      "index": 0,
       "lookup_id": "string",
       "matches": [
         {
@@ -163,6 +166,7 @@ Queries all state databases for any PII records that are an exact match to the l
 {
   "data": [
     {
+      "index": 0,
       "lookup_id": "string",
       "matches": [
         {
@@ -182,6 +186,7 @@ Queries all state databases for any PII records that are an exact match to the l
       ]
     },
     {
+      "index": 1,
       "lookup_id": "string",
       "matches": [
         {
@@ -218,6 +223,7 @@ Status Code **200**
 |Name|Type|Required|Restrictions|Description|
 |---|---|---|---|---|
 |» data|array|false|none|none|
+|»» index|integer|true|none|Index of match query request item that match query response item corresponds to, starting from 0. For every match query item provided in the request, a match response item is returned, even if no matches are found. Index is derived from the implicit order of match query request items provided in the request.|
 |»» lookup_id|string¦null|false|none|the identifier of the match request|
 |»» matches|[object]|false|none|none|
 |»»» first|string|false|none|First name|
--- a/docs/openapi/generated/duplicate-participation-api/openapi.md
+++ b/docs/openapi/generated/duplicate-participation-api/openapi.md
@@ -44,15 +44,23 @@ Queries all state databases for any PII records that are an exact match to the l
 
 ```json
 {
-  "query": {
-    "first": "string",
-    "middle": "string",
-    "last": "string",
-    "ssn": "000-00-0000",
-    "dob": "1970-01-01"
-  }
+  "query": [
+    {
+      "first": "string",
+      "middle": "string",
+      "last": "string",
+      "ssn": "000-00-0000",
+      "dob": "1970-01-01"
+    }
+  ]
 }
 ```
+
+<h3 id="query-for-matches-parameters">Parameters</h3>
+
+|Name|In|Type|Required|Description|
+|---|---|---|---|---|
+|query|body|[[#/paths/~1query/post/requestBody/content/application~1json/schema/properties/query/items](#schema#/paths/~1query/post/requestbody/content/application~1json/schema/properties/query/items)]|true|none|
 
 > Example responses
 
@@ -60,26 +68,30 @@ Queries all state databases for any PII records that are an exact match to the l
 
 ```json
 {
-  "lookup_id": "string",
-  "matches": [
+  "data": [
     {
-      "first": "string",
-      "middle": "string",
-      "last": "string",
-      "ssn": "000-00-0000",
-      "dob": "1970-01-01",
-      "state": "ea",
-      "state_abbr": "ea",
-      "exception": "string",
-      "case_id": "string",
-      "participant_id": "string",
-      "benefits_end_month": "2021-01",
-      "recent_benefit_months": [
-        "2021-05",
-        "2021-04",
-        "2021-03"
-      ],
-      "protect_location": true
+      "lookup_id": "string",
+      "matches": [
+        {
+          "first": "string",
+          "middle": "string",
+          "last": "string",
+          "ssn": "000-00-0000",
+          "dob": "1970-01-01",
+          "state": "ea",
+          "state_abbr": "ea",
+          "exception": "string",
+          "case_id": "string",
+          "participant_id": "string",
+          "benefits_end_month": "2021-01",
+          "recent_benefit_months": [
+            "2021-05",
+            "2021-04",
+            "2021-03"
+          ],
+          "protect_location": true
+        }
+      ]
     }
   ]
 }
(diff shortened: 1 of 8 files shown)


ryanhofdotgov:
Here are some quick, initial thoughts:
- How should the API consumer associate query input elements with their corresponding match response? You could try to rely on knowledge of the data type and the current matching algorithm and tell them to use the SSN value to pair up input/output. However, it might be better to have 1:1 array element responses and explicitly send back an "empty" type.
- I'd avoid naming things like `data`. Maybe `household` instead? Even `query` has bugged me from the outset. 
- There are implications to lookup API semantics – it was supposed to save the original query input, so what does that mean when the input is a whole household.

echappen:
@ryanhofdotgov Lots of food for thought, thank you! Happy to sync on this by video if that's more efficient, but some initial responses:

> How should the API consumer associate query input elements with their corresponding match response? You could try to rely on knowledge of the data type and the current matching algorithm and tell them to use the SSN value to pair up input/output. However, it might be better to have 1:1 array element responses and explicitly send back an "empty" type.

^ Idea: we could include the original query item for every match item, like so:

```
lookip_id: "XYZ",
query_item: {
  // one person's PII from original query 
},
matches: [
  // matches related to that query item
]
```

So for every query item in the request, there would be a match item in the response. 

> I'd avoid naming things like data. Maybe household instead? Even query has bugged me from the outset.

Household feels a little too specific to me — will a list of PII for a single query always be a household?

> There are implications to lookup API semantics – it was supposed to save the original query input, so what does that mean when the input is a whole household.

Thinking out loud for what it would mean to have 1 lookup ID for multiple PII records:

Say there are matches on just a subset of the individuals in the original match request. Like a household with 12 people and only 2 of them have matches. 

State 1 sends the Lookup ID to State 2. State 2 makes a Lookup request and gets back the original PII for all 12 individuals. Which ones have matches?

To find out, State 2 enters all 12 invidiuals back into a match query. The hope is that they've entered all the data consistently, and this match query returns the same data as the original match request from State 1. 

So it seems the connection between query and match gets fuzzier as more members are added to the query. 

Conversely, for 1 Lookup ID for every PII record, states could communicate about only the records with matches by passing only the Lookup ID's in question. The match query created from the Lookup ID responses could be easier for the EW — less PII to input for each query may mean better success at getting back the match. 

The downside to doing this is that State 2 wouldn't know about the other members of the household. But if those members don't have matches, is that still important context?

ryanhofdotgov:
Re: Household naming – that's definitely the intent of allowing multiple queries at once – to support (re)certification of a single household with multiple members, which maps to a single SNAP application/case. I am hoping state integrators do not misuse it for other purposes.

Upper bound of household size drives the maximum number of query elements we will accept before rejecting the API call. Casual googling showed that [a family of 36 exists in 21st century United States](https://www.today.com/parents/move-over-duggars-meet-family-34-children-1d80298344), so 50 is on the same order. We expect that most multi person household queries would be much smaller.

echappen:
> Re: Household naming – that's definitely the intent of allowing multiple queries at once – to support (re)certification of a single household with multiple members, which maps to a single SNAP application/case. I am hoping state integrators do not misuse it for other purposes.

I'm anticipating this schema would be used for both household and individual queries. Responses to an individual query would have a response with 1 match item in it. In this case, the term 'household' makes less sense to me. Or are you implying we should have different response schemas for household vs individual queries?

> Upper bound of household size drives the maximum number of query elements we will accept before rejecting the API call. Casual googling showed that a family of 36 exists in 21st century United States, so 50 is on the same order. We expect that most multi person household queries would be much smaller.

I was thinking of it as more of a human usability issue than a computational one. I sense this fuzziness I described may become an issue for EW's to enter data even at much smaller households. 

Every time states pass a Lookup ID back and forth, each EW would have to enter multiple PII records to see the match data. I would expect a household of ~30 individuals to be prohibitively cumbersome to enter every time, but I think it could be cumbersome even at ~5 people.

ryanhofdotgov:
I'll send out an invite to the whole engineering team to chat as I think it will good to tease these details out more.

Although it was in my initial quick take, I'm not really that bullish on the `household` name, but just to clarify on intended use... A SNAP application is for a household, which can have one or more individuals. Each application will be associated with a case in the state-level benefits program system. So the same API would be use regardless of how many folks were in the household.

Here's an [application from California](https://www.cdss.ca.gov/cdssweb/entres/forms/English/SAWS2PLUS.pdf), and a [recertification form from California](https://www.cdss.ca.gov/cdssweb/entres/forms/english/cf37.pdf) as examples, just to see the nomenclature they use a little more directly. (CalFresh is the local name for SNAP).

echappen:
That's really helpful context, thanks! My mental model until now has been very individual-centric and I need to change that 😄

echappen:
Documenting what we came to consensus on in our sync:
- Should be 1-1 relationship between Lookup ID and single PII record in query (so each PII record in a Match Query will have its own corresponding Lookup ID)
- If a case ID or participant ID is provided in the match data, then Lookup ID is unnecessary for inter-state communication because they can just send the other state's case id or participant id instead. 
- Lookup ID data should remain "pure" (not be populated with match-specific data) to avoid implementation churn by states.

cjhskippy:
One minor dissent: case ID by itself is not enough for inter-state communication to identify an individual because each case can involve multiple participants.  The other state would need the combination of case ID + participant ID, or case ID + Lookup ID.

It's my hope that Lookup ID, or some match ID is as convenient as any of the other options available to states.

Also worth noting that if the initiating state only shares the participant ID and case ID, the matching state has no discretion to evaluate whether the match involves the same person.  They have to trust that the initiating state compared the two names to assess whether there's any likelihood that it's really the same person.

echappen:
@ryanhofdotgov @cjhskippy @kingcomma Given Thursday's [discussion](https://github.com/18F/piipan/pull/1082#issuecomment-863482989), are there any other changes needed for this schema before I start implementing?

echappen:
> How should the API consumer associate query input elements with their corresponding match response?

After discussing with our Engineering team and other members of TTS Engineering, and researching other API's, we decided to use Amazon's `index` [approach](https://docs.aws.amazon.com/comprehend/latest/dg/get-started-batch.html) to help API consumers with this issue. 

API schema and docs have been updated with an additional `index` for each match response item, as well as a description for the match query response's `data` array. @cjhskippy @erinzstrenio let me know if these descriptions are good to go. They can be found [here](https://github.com/18F/piipan/blob/51fd16e4040903258eab925e31d93352c75b8c7e/docs/openapi/generated/duplicate-participation-api/openapi.md).

docs/openapi/generated/duplicate-participation-api/openapi.md
```diff
@@ -44,42 +44,55 @@ Queries all state databases for any PII records that are an exact match to the l
...
+  ]
 }
 ```
 
+<h3 id="query-for-matches-parameters">Parameters</h3>
+
+|Name|In|Type|Required|Description|
+|---|---|---|---|---|
+|query|body|[[#/paths/~1query/post/requestBody/content/application~1json/schema/properties/query/items](#schema#/paths/~1query/post/requestbody/content/application~1json/schema/properties/query/items)]|true|none|
+
```
kingcomma, line 64: *Question:* do you know why this parameters section was inserted by Widdershins? It doesn't show up in previous docs and the Widdershins config doesn't appear to have changed. I poked a bit and couldn't find an answer.

docs/openapi/generated/duplicate-participation-api/openapi.md
```diff
@@ -44,42 +44,55 @@ Queries all state databases for any PII records that are an exact match to the l
...
-  }
+  "query": [
+    {
+      "first": "string",
+      "middle": "string",
+      "last": "string",
+      "ssn": "000-00-0000",
+      "dob": "1970-01-01"
+    }
+  ]
```
kingcomma, line 55: *Thought (non-blocking):* It's a bummer that [Widdershins doesn't support multiple request examples](https://github.com/Mermade/widdershins/issues/192)...

*Suggestion:* In light of that, I think it might be helpful for states if the description for this example explicitly calls out that this is an example request for a single individual / participant. Maybe something like "A request with a single individual?" (needs some wordsmithing).

docs/openapi/generated/duplicate-participation-api/openapi.md
```diff
@@ -44,42 +44,55 @@ Queries all state databases for any PII records that are an exact match to the l
...
 
+<h3 id="query-for-matches-parameters">Parameters</h3>
+
+|Name|In|Type|Required|Description|
+|---|---|---|---|---|
+|query|body|[[#/paths/~1query/post/requestBody/content/application~1json/schema/properties/query/items](#schema#/paths/~1query/post/requestbody/content/application~1json/schema/properties/query/items)]|true|none|
+
 > Example responses
 
 > A query returning a single match
```
kingcomma, line 67: *Suggestion:* Similar to above, would it be more helpful to state implementors if this description was more specific — something like "A query for a single individual returning a single match"?

docs/openapi/generated/duplicate-participation-api/openapi.md
```diff
@@ -89,49 +102,108 @@ Queries all state databases for any PII records that are an exact match to the l
...
+      "index": 0,
+      "lookup_id": null,
+      "matches": []
+    }
+  ]
 }
 ```
 
-> A query returning multiple matches
+> A query returning multiple matches for one record
```
kingcomma, line 115: *Suggestion:* Another one — perhaps something like "A query for a single individual returning multiple matches"?

docs/openapi/generated/duplicate-participation-api/openapi.md
```diff
@@ -89,49 +102,108 @@ Queries all state databases for any PII records that are an exact match to the l
...
+          "benefits_end_month": null,
+          "protect_location": null
+        }
+      ]
+    }
+  ]
+}
+```
+
+> A query returning one match for multiple records
```
kingcomma, line 163: Same as above.

kingcomma (commented):
Added some inline comments, mostly nits/suggestions around example descriptions. (I commented on the generated doc, hope that's ok).

*Suggestion:* I think it would also be helpful for states if we show an example response of a query for multiple individuals with a match for one/some but not all (i.e., an `index` but no `lookup_id` and an `matches` array).

docs/openapi/generated/duplicate-participation-api/openapi.md
echappen: This might be a result of me changing query from an object to an array in the request schema. I'll see if I can remove this.

echappen: Couldn't figure out how to remove this table, or alter the query row to remove that giant `Type` link. It may be looking weird because we're using the `--shallowSchemas` parameter, which doesn't resolve the data from the `Type` link within the table. If that was removed and the `--resolve` flag was used, then the table would look something like this: 

<h3 id="query-for-matches-parameters">Parameters</h3>

|Name|In|Type|Required|Description|
|---|---|---|---|---|
|query|body|[[#/paths/~1query/post/requestBody/content/application~1json/schema/properties/query/items](#schema#/paths/~1query/post/requestbody/content/application~1json/schema/properties/query/items)]|true|none|
|» first|body|string|true|PII record's first name|
|» middle|body|string|false|PII record's middle name|
|» last|body|string|true|PII record's last name|
|» ssn|body|string|true|PII record's social security number|
|» dob|body|string(date)|true|PII record's date of birth|

That gives more context to the `Type` link, but probably isn't ideal either.

echappen:
@erinzstrenio Changed my mind and just put all your edits in this PR. To recap all your edits:

"The API for the Duplicate Participation system where matching and lookups will occur"
^ The API where matching and lookups will occur

"Matching PII records, if any exist"
^ Successful response. Returns match response items.

"the identifier of the match request"
^ The identifier of the match request item, if a match is present. This ID can be used for looking up the PII of the original match request item.

"Index of match query request item that match query response item corresponds to, starting from 0."
^ The index of the request item that the response item corresponds to, starting from 0.

"Participant's state-specific case identifier" (case_id)
^ Participant's state-specific case identifier. Can be the same for multiple participants.

participant_id
^ Is unique to the participant

personal identifiable information
^ PII

get the original match data related to a Lookup ID
^ Capitalize sentence

Last Response code table
200 - Successful response. Returns original match query request item.

Not Found
^ lowercase F

echappen:
Thinking through error handling——this is mostly about batch API's, but I think this pattern could be extended to current and future API's.

For item-level errors in a batch request, the [Amazon](https://docs.aws.amazon.com/comprehend/latest/dg/get-started-batch.html) example has a results list and an errors list.
- A result item pertaining to a request item can be in either list, but not in both.
- Both result and error items use an index to correspond to the items in the original request.
- Errors and results lists are always present in the data object, even if they're empty.

Example of an entire response body:

```
{
  meta: {...},
  data: {
    results: [
      {
        index: 0,
        lookup_id: "xyz",
        matches: [...]
      },
      {
        index: 1,
        lookup_id: "abc",
        matches: [...]
      }
    ],
    errors: [
      {
        index: 2,
        code: "InternalServerException",
        message: "Unexpected Server Error. Please try again."
      }
    ]
  }
}

```

I like this approach because it may be easier for a client to do `data.errors.present?`, rather than iterating through each result to see if any errors exist.

If there's an HTTP or other top-level error for the entire request, then instead of a data property, you'd have an errors property, which would be an array:

```
{
  meta: {...},
  errors: [
    {
      code: "InternalServerException",
      message: "Unexpected Server Error. Please try again."
    }
  ]
}
```

In other words, at top-level, there's either a data or an errors property, but not both. An errors property that's present would never be empty, or else it wouldn't be there. This is in line with the JSON API standard, which is proposed [here](https://github.com/18F/piipan/pull/1119).

Extending this to single-resource API requests: `data` would be an object, but `errors` would still be an array. In this case, errors returned could be resource-related and not just HTTP-level errors.

echappen:
I updated the PR description to reflect all the areas we've talked about and are trying to address in this schema. Schema is ready for another review.

echappen:
I updated the error objects—both http-level and item-level—to conform more to JSON API [spec](https://jsonapi.org/format/#error-objects). The spec has no required fields for error objects, but I'm thinking we'd want to make `status` required for http-level errors, and `index` required for item-level errors.

echappen:
@ryanhofdotgov This schema is updated and ready for review

ryanhofdotgov (approved):
LGTM 👍

Issue #1091, opened 2021-06-17
Clarify process for testing secured remote APIs locally


kingcomma:
As a developer, I want to know the process for connecting to remote APIs that are secured by Easy Auth from my local machine (e.g., while testing a locally running Dashboard app). Currently this documentation is spread around several subsystems making it difficult to understand.

- [ ] Document process in a central location (i.e., `securing-internal-apis.md`)
- [ ] Detail process for using `assign-app-role` and `authorize-cli` scripts
- [ ] Switch `authorize-cli` to use same input parameters as `assign-app-role`

Issue #1055, opened 2021-06-14
Apply managed identity approach to the metrics Function Apps


ryanhofdotgov:
- [x] Only Dashboard can access metrics API (#1084 and #1092)
- [ ] Only event from Event Hub can trigger Collector
- [ ] ~Log managed identity data in Collector (see #1054)~

kingcomma:
Event Hub identities / authentication: https://docs.microsoft.com/en-us/azure/event-grid/security-authentication and https://docs.microsoft.com/en-us/azure/event-grid/managed-service-identity

kingcomma:
Using an HTTP trigger (webhook) for event delivery: https://docs.microsoft.com/en-us/azure/azure-functions/functions-bindings-http-webhook

ryanhofdotgov:
Saw [this comment ](https://github.com/18F/piipan/pull/1054#issuecomment-860882578) in #1054 and just want to clarify that the primary goal is just to ensure that only the Event Hub event can trigger the collector (e.g., can't be invoked by an unauthenticated HTTP client), but do not actually know if managed identity is the best and typical way to accomplish that.

kingcomma:
[Service tags](https://docs.microsoft.com/en-us/azure/event-grid/network-security#service-tags) seem like a potential solution for limiting a Function to only Event Grid, but the option doesn't appear to be available in Azure Government (it's not specifically documented one way or another).

Commercial portal:

<img width="312" alt="Screen Shot 2021-06-24 at 5 05 15 PM" src="https://user-images.githubusercontent.com/3246955/123332744-e8631c80-d50e-11eb-8696-71b37157f702.png">

Government portal:

<img width="319" alt="Screen Shot 2021-06-24 at 5 05 06 PM" src="https://user-images.githubusercontent.com/3246955/123332755-ec8f3a00-d50e-11eb-9104-fbe0ce535c09.png">

kingcomma:
With more experimenting, service tags may be usable on AzureUSGovernment. I can set an access restriction that references a service tag via the CLI with `az functionapp config access-restriction add -n ... -g ... --priority 100 --service-tag AzureEventGrid` and the rule will show up in the Portal:

<img width="321" alt="Screen Shot 2021-06-25 at 2 39 22 PM" src="https://user-images.githubusercontent.com/3246955/123470988-24f05000-d5c3-11eb-90e4-4c5a2d70d7ba.png">

kingcomma:
[...]

PR #1092, opened 2021-06-17
Make authenticated requests when calling metrics API from dashboard

kingcomma:
- Inject shared `AuthorizedJsonApiClient` instead of `HttpClient`
- Adapt `ParticipantUploadRequest` to work with `AuthorizedJsonApiClient`
- Update tests and doc

Partially addresses #1055.

I have a follow up issue to refine the documentation around local testing (#1091).

### DIFF
--- a/dashboard/docs/dashboard.md
+++ b/dashboard/docs/dashboard.md
@@ -10,7 +10,12 @@ To run the app locally:
     dotnet dev-certs https --trust
 ```
 
-2. Run the app using the `dotnet run` CLI command:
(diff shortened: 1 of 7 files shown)


kingcomma:
The changes to `Startup.cs` cause a slight drop in test coverage. If others are on board, I'd like to defer trying to cover that new code until we figure out an optimal way for unit testing our Razor apps and Function apps with DI.

echappen (approved):
Sorry I missed this earlier. Thanks for doing this 👍

Issue #333, opened 2021-02-09
Replace key/vault access in Metrics resources with Managed Identity


echappen:
I know it's already been discussed, but just noting the future work to implement the use of a managed identity (or system-assigned identity) for connecting directly to the database without the need for getting/storing secrets with key vault.

_Originally posted by @kingcomma in https://github.com/18F/piipan/pull/298#discussion_r573088872_

kingcomma:
@cjhskippy This appears to be a duplicate of #409. I'm closing for now, but happy to reopen if I'm mistaken.

PR #1099, opened 2021-06-22
Isolate core database creation in IaC

kingcomma:
Now that the core database is used outside of just the metrics sub-system, move database (and cluster) creation out of `create-metrics-resources.bash`:

- Move core database creation to own bash script
- Rename relevant resources (`s/metrics/core/`)
- Use existing `kv-core` keyvault
- Update metrics applications with renamed `secretName` (necessary until remaining work in #333 is complete)

OmniSharp caught some formatting adjustments on save.

Precursor for #333


iac/create-resources.bash
```diff
@@ -582,6 +582,9 @@ main () {
   # Establish metrics sub-system
   ./create-metrics-resources.bash "$azure_env"
 
+  # Core database server and schemas
+  ./create-core-databases.bash "$azure_env"
```
echappen, line 586: Just checking that this database doesn't need to be created before other metrics resources above.

kingcomma: Correct — I had to confirm this as well. The metrics functions require a database connection string, but that can be constructed by `pg_connection_string()` / `az postgres show-connection-string` without an existing database.

echappen (approved):
LGTM, thanks for refactoring this!

PR #1100, opened 2021-06-22
Configure core database with managed roles

kingcomma:
- Pull common procedures out to functions
- Bring roles and privileges closer to how they are handled in `participant-records`, though maintains use of `public` schema
- Create managed roles for Metrics API (read) and Metrics Collect (read/write)

Partially addresses #409. Would definitely appreciate a keen eye on the roles and privileges.

Subsequent updates will switch applications to use managed identities when connecting to database and remove vestigial Key Vault references.

### DIFF
--- a/iac/create-core-databases.bash
+++ b/iac/create-core-databases.bash
@@ -6,12 +6,146 @@ source "$(dirname "$0")"/../tools/common.bash || exit
 set_constants () {
   DB_SERVER_NAME=$PREFIX-psql-core-$ENV
   DB_ADMIN_NAME=piipanadmin
+  SUPERUSER=$DB_ADMIN_NAME
+
   DB_NAME=metrics
   DB_TABLE_NAME=participant_uploads
+
   VAULT_NAME=$PREFIX-kv-core-$ENV
-  # Name of secret used to store the PostgreSQL metrics server admin password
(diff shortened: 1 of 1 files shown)


kingcomma:
Updated IaC has been run against `tts/dev`

iac/create-core-databases.bash
```diff
@@ -6,14 +6,148 @@ source "$(dirname "$0")"/../tools/common.bash || exit
...
+  PSQL_OPTS=(-v ON_ERROR_STOP=1 -X -q)
+  TEMPLATE_DB=template1
+
+  # Name of Azure Active Directory admin for PostgreSQL server
+  PG_AAD_ADMIN=piipan-admins
+
   PRIVATE_DNS_ZONE=$(private_dns_zone)
 }
 
+init_db () {
```
ryanhofdotgov, line 26: thought: at some point in the future, it would be nice to unify the database approaches so we can have a common set of db functions

ryanhofdotgov (approved):
👍

iac/create-core-databases.bash
kingcomma: 💯  I had thought about taking a stab at that in this PR, but it the scope of the work started to outweigh the scope of this ticket.

PR #1102, opened 2021-06-23
Change App Service instances to Windows

### LINKED ISSUE #1042: Change all app service instances to Windows-under-the-hood plans
Change all app service instances to Windows-under-the-hood plans:
- there are specific networking features/visualizations that aren't available in Linux-under-the-hood plans
- allows us to be consistent with how we deploy other components in the system (which we changed to Windows for similar lack-of-parity reasons)

ryanhofdotgov:
Closes #1042. App Service instances under Linux to not have as many logging options and do not have the networking UI in the portal that makes it easy to understand the networking config.

Updated IaC has been run on tts/dev.

Deploy notes:
- manually drop existing app service instances as cannot update in place due to platform change/service plan change
- rerun IaC
- manually drop obsolete Linux service plan `piipan-app-plan`

### DIFF
--- a/iac/arm-templates/query-tool-app.json
+++ b/iac/arm-templates/query-tool-app.json
@@ -40,6 +40,7 @@
                     "value": "dotnetcore"
                 }
             ],
+            // Use 64-bit runtime on Windows for consistency with local Linux and macOS dev environments
             "use32BitWorkerProcess": false,
             "ipSecurityRestrictions": [],
             "ftpsState": "Disabled",
@@ -67,6 +68,7 @@
             },
             "kind": "windows",
             "properties": {
+                // Must be false for Windows deployments
                 "reserved": false
             }
         },
--- a/iac/arm-templates/query-tool-app.json
+++ b/iac/arm-templates/query-tool-app.json
@@ -31,7 +31,16 @@
         "appName": "[concat(parameters('appName'))]",
         "sku": "S1",
         "baseSiteConfig": {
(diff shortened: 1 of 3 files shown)


iac/arm-templates/query-tool-app.json
```diff
@@ -31,7 +31,16 @@
...
+            // .NET 3.1 Core uses "v4.0", .NET 5 uses "v5.0"
+            "netFrameworkVersion": "v4.0",
+            "metadata": [
+                {
+                    // .NET 3.1 Core uses "dotnetcore", .NET 5 uses "dotnet"
+                    "name": "CURRENT_STACK",
+                    "value": "dotnetcore"
+                }
+            ],
+            "use32BitWorkerProcess": false,
```
kingcomma, line 43: Just curious, was this necessary for the switch to Windows or included for a different reason?

kingcomma (approved):
👍 Thanks for digging through so much for this "small" change!

iac/arm-templates/query-tool-app.json
ryanhofdotgov: Yeah, I added a few more comments for you to re-review. In short, there is only [64-bit .NET runtimes on Linux](https://devblogs.microsoft.com/aspnet/64-bit-asp-net-core-on-azure-app-service/) and macOS.

kingcomma: Thanks!

kingcomma (approved):

Issue #1111, opened 2021-06-25
Deliver Event Grid events to Metrics Collector using managed identity


kingcomma:
Consistent with our zero-trust approach, deliver events from Event Grid to the Metrics Collector Function app using a managed identity.

- [ ] Configure Event Grid (see: https://docs.microsoft.com/en-us/azure/event-grid/enable-identity-custom-topics-domains)
- [ ] Enable app service authentication on Metrics Collector
- [ ] Log managed identity information when Metrics Collector is called

This ticket is an extension of #1055.

kingcomma:
Turns out the required approach isn't as straightforward as what I outlined above. @ryanhofdotgov I'm a little stuck trying to figure out how best to achieve our goal of a zero-trust approach to Azure Functions with Event Grid triggers. Would you agree that the trade-offs outlined below are not acceptable? If so, is the current approach sufficient or should we research other options?

Using managed identities to delivery events (taking the same approach as we've taken with our HTTP triggered Function apps that run our APIs) requires two changes that don't feel like acceptable tradeoffs:

1. Switching the Function to an HTTP trigger and changing the Event Grid subscription to use a ["webhook" endpoint type rather than an "Azure Function" endpoint type](https://docs.microsoft.com/en-us/azure/event-grid/handler-functions)
2. Implementing custom application-level code that performs a [validation handshake with Event Grid](https://docs.microsoft.com/en-us/azure/event-grid/webhook-event-delivery#endpoint-validation-with-event-grid-events) (this step is handled automatically by Azure if using the "Azure Function" endpoint type)

As a reminder, the current approach we are taking for delivering metrics events to the metrics collection Function app is to:

1. Add an access restriction on the Function app that limits inbound traffic to the `AzureEventGrid` service tag
2. Let Event Grid do it's [built-in behavior](https://docs.microsoft.com/en-us/azure/event-grid/security-authentication#:~:text=Access%20keys%20are%20fetched%20using%20Event%20Grid%20service%20principal%27s%20credentials.%20The%20permissions%20are%20granted%20to%20Event%20Grid%20when%20you%20register%20the%20Event%20Grid%20resource%20provider%20in%20their%20Azure%20subscription.) of grabbing the Function's baked-in access key and using it to authenticate event delivery

ryanhofdotgov:
[...]

PR #1112, opened 2021-06-25
Restrict Metrics Collector to Event Grid at network level

### LINKED ISSUE #1055: Apply managed identity approach to the metrics Function Apps

kingcomma:
- Conditionally add access restriction for `AzureEventGrid` service tag
- Requires AZ CLI >= 2.23, earlier versions [throw error](https://github.com/Azure/azure-cli/issues/16661#issuecomment-839551317) when using `--service-tag` parameter (could also be resolved by switching to an ARM template)

Closes #1055. Implementing managed identities has been adding to the backlog as #1111.

### DIFF
--- a/docs/iac.md
+++ b/docs/iac.md
@@ -4,7 +4,7 @@
 
 All prerequisites are available in [Azure Cloud Shell](https://docs.microsoft.com/en-us/azure/cloud-shell/overview).
 
-- [Azure Command Line Interface (CLI)](https://docs.microsoft.com/en-us/cli/azure/install-azure-cli)
+- [Azure Command Line Interface (CLI)](https://docs.microsoft.com/en-us/cli/azure/install-azure-cli) >= 2.23.0
 - [Azure Functions Core Tools](https://docs.microsoft.com/en-us/azure/azure-functions/functions-run-local)
 - [.NET Core 3.1 SDK](https://dotnet.microsoft.com/download)
 - `bash` shell, `/dev/urandom` – included in macOS, Linux, Git for Windows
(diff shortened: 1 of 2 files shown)


ryanhofdotgov:
All that work for me get a version less than 2.22 installed and now you tell me I need 2.23 😆

kingcomma:
🤦 I wasn't thrilled with this realization. I'm basing 2.23 entirely on the linked GitHub issue. On my machine, 2.16 did not work but 2.25 did. Might be worth confirming it doesn't work for you before going through the headache.

Also, honestly, it's probably a similar amount of work to convert this Function app creation to an ARM template as it is for us to fool around with CLI versions. Happy to make that switch if it seems better.

ryanhofdotgov:
Nah, let's keep moving forward 🙂 – would you update the iac.md doc in this PR to indicate >= 2.23 is required?

kingcomma:
Doc updated. I also confirmed the necessary change was part of [2.23.0](https://docs.microsoft.com/en-us/cli/azure/release-notes-azure-cli?tabs=azure-cli#may-06-2021):

> az webapp/functionapp config access-restriction add/remove: Support for Service Tag, Http headers and multi-source rules.

ryanhofdotgov (approved):

kingcomma:
IaC run against `tts/dev`.

Issue #701, opened 2021-04-16
Use PostgreSQL in place of Table Storage for lookup API


kingcomma:
As a developer, I want a database solution for the lookup API that fits within our existing architecture and does not increase cognitive load by introducing new tools and approaches.

As a system owner, I want a database solution with automated backups (which are not available using Table Storage).

Task consists of:
- [ ] Create a new database within the `core` cluster and bind it to the orchestrator as part of the IaC process
- [ ] Update orchestrator to use the new database for storing and retrieving lookup IDs
- [x] Quick ADR write-up, why we thought table records were a good quick start, why we changed to our default RDBMS.

ryanhofdotgov:
@kingcomma I think actually we want to leave the participant-records db cluster as is, to more strictly isolate the bulk PII. Instead elevate the metrics db cluster to a catch-all/"core" cluster.

I anticipate using it with separate databases for:
* metrics
* lookup API
* to be determined status API
* to be determined collaboration work

As backup retention policy is set at the cluster level, this keeps databases with similar retention needs together.

ryanhofdotgov:
Added ADR as part of task (but can be split off as necessary)

cjhskippy:
Closed in connection with removal of Lookup API.

PR #1118, opened 2021-06-29
Namespace `DB_NAME` and inline `DB_TABLE_NAME`

kingcomma:
Prep work to better support creating multiple databases within `create-core-databases.bash`.

Precursor to #701.

### DIFF
--- a/iac/create-core-databases.bash
+++ b/iac/create-core-databases.bash
@@ -8,8 +8,7 @@ set_constants () {
   DB_ADMIN_NAME=piipanadmin
   SUPERUSER=$DB_ADMIN_NAME
 
-  DB_NAME=metrics
-  DB_TABLE_NAME=participant_uploads
+  METRICS_DB_NAME=metrics
 
   VAULT_NAME=$PREFIX-kv-core-$ENV
   PG_SECRET_NAME=core-pg-admin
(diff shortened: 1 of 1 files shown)


ryanhofdotgov:
@kingcomma Can you share a little bit about what the shape of this script might look like afterwards? I might be inclined to create a common cluster set up script, but separate per-database set up scripts to keep the size of each bash script smaller.

kingcomma:
I've been kicking this around a bit. Long term, I think this could look similar to what you described. Specifically:

- A common cluster set up script
- A common set of database tools for creating databases, roles, configuring AAD (perhaps part of `iac-common`)
- A chunk of commands in the IaC for each database:
  - Create database
  - Create roles, configure managed identity
  - Run an external DDL to set up the remaining resources (or maybe this gets cut out in favor of some other database change management approach)

Near term — to meet the needs of #701 — I had admittedly planned on shoehorning the lookup resource creation (a single db, table, and role) into this script and defer longer term tasks. But if we think it's worth spending the time on those tasks now, I'm more than happy to do so.

ryanhofdotgov (approved):
I'm OK with that 👍 

We will probably need a 3rd database in the core cluster "soon" and we can address then.

PR #1119, opened 2021-06-29
Add ADR for JSON API proposal

echappen:
A discussion came up while doing #1082 about possibly standardizing our response schemas across our API's. Here's an initial proposal to use a subset of the JSON API schema for this purpose.

### DIFF
--- a/docs/adr/0015-use-json-api-spec-for-apis.md
+++ b/docs/adr/0015-use-json-api-spec-for-apis.md
@@ -1,0 +1,45 @@
+# 1. Use a subset of the JSON API Specification for all API responses
+
+Date: 2021-06-29
+
+## Status
+
+Proposed
+
+## Context
+
+While our API endpoints serve difference purposes, they generally need to return similar data:
+
+1. metadata about the request/response
+1. HTTP-level error messages
+1. the actual payload
+
+We've been adding API endpoints organically as needed, which has resulted in different schemas and approaches to returning these data types.
+
+Standardizing response schemas across our API's can:
+- remove the need for API developers to reinvent this wheel for every API and endpoint
+- help clients integrate with the API faster
+
+While we use the [OpenAPI Specification](https://swagger.io/specification/) for our API schemas, the purpose of OpenAPI is only to descibe schemas consistently, not to determine what the schema should be. For the latter, other specifications exist. It'd be up to us to pick one or create our own.
+
+## Decision
+
+We adopt the following subset of the [JSON API Specification](https://jsonapi.org/) for a [top-level](https://jsonapi.org/format/#document-top-level) response schema:
+
+A response document *MUST* contain at least one of the following top-level members:
+
+- *data*: the document's "primary data"
+- *errors*: an array of error objects
+- *meta*: a meta object that contains non-standard meta-information.
+
+The members data and errors *MUST NOT* coexist in the same document.
+
+Primary data *MUST* be either:
+
+- a single resource object or null, for requests that target single resources
+- an array of resource objects or an empty array ([]), for requests that target resource collections
+
+## Consequences
+
+- We'll need to adjust our existing API endpoints to return this top-level schema.
+- Breaking changes to the API
--- a/docs/adr/0016-use-json-api-spec-for-apis.md
+++ b/docs/adr/0016-use-json-api-spec-for-apis.md
@@ -1,4 +1,4 @@
-# 1. Use a subset of the JSON API Specification for all API responses
+# 16. Use a subset of the JSON API Specification for all API responses
 
 Date: 2021-06-29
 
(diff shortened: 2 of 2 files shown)


docs/adr/0015-use-json-api-spec-for-apis.md
```diff
@@ -0,0 +1,45 @@
...
+
+Standardizing response schemas across our API's can:
+- remove the need for API developers to reinvent this wheel for every API and endpoint
+- help clients integrate with the API faster
+
+While we use the [OpenAPI Specification](https://swagger.io/specification/) for our API schemas, the purpose of OpenAPI is only to descibe schemas consistently, not to determine what the schema should be. For the latter, other specifications exist. It'd be up to us to pick one or create our own.
+
+## Decision
+
+We adopt the following subset of the [JSON API Specification](https://jsonapi.org/) for a [top-level](https://jsonapi.org/format/#document-top-level) response schema:
```
ryanhofdotgov, line 27: There's not as much of an advantage of adopting a subset of JSON API vs just adopting what it requires entirely; i.e., you want to be able to just grab a JSON API client code and expect it to work with our APIs.

Are we really just implementing a subset, or are we just going to initially be concerned about what is required of JSON API and not worry about implementing optional bells and whistles?

ryanhofdotgov, line 27:
```suggestion
We will adopt the following subset of the [JSON API Specification](https://jsonapi.org/) for a [top-level](https://jsonapi.org/format/#document-top-level) response schema:
```

echappen: I was thinking only of our immediate need for a consistent response schema, but I agree adopting the spec fully would be more beneficial.

There are certain required aspects of JSON API we haven’t discussed which may influence our decision to adopt it fully:

1. [Content negotiation](https://jsonapi.org/format/#content-negotiation), specifically requiring clients to send a specific JSON API Content Type header
2. Primary data in the `data` field must be resource objects. [Resource objects](https://jsonapi.org/format/#document-resource-objects) must have an `id` and `type` field

I think we can accommodate these requirements, but it would mean states would need to conform to the spec as well.

docs/adr/0015-use-json-api-spec-for-apis.md
```diff
@@ -0,0 +1,45 @@
+# 1. Use a subset of the JSON API Specification for all API responses
```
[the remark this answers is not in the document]
echappen: No sweat!

ryanhofdotgov:
I just read the JSON API spec (for the first time really), and I agree with you – it doesn't make sense to adopt it wholesale, just use elements of it for inspiration.

JSON API, according to one of its authors, is all about ["incrementally fetching and updating a graph over HTTP"](https://twitter.com/wycats/status/925847356532125696/photo/1), with focus on incremental fetching, updating, and leveraging HTTP caching, and for the duplicate participation API, we really don't care about any of that.

Some open questions though on the very tiny subset we're choosing to incorporate:
-  the ADR says `data` must be a list of resource objects – what's a resource object? Is it exactly how JSON API describes it where each has a `type` and an `id`?
- the ADR says we're applying this approach to API responses, but JSON API applies the principles to POST body requests too. 
- should we incorporate any of their recommendations (e.g., [naming](https://jsonapi.org/recommendations/#naming)) that are outside the spec?

echappen:
> what's a resource object? Is it exactly how JSON API describes it where each has a type and an id?

As far as I can tell, yes. It's just an object with both an `id` and `type` field at minimum. 

> the ADR says we're applying this approach to API responses, but JSON API applies the principles to POST body requests too.

If we adopt the spec for requests, then the request data would need to be resources objects as well. Since we're still fuzzy on what exactly our request data is going to look like, it may be best to hold off on conforming requests to the spec.   

> should we incorporate any of their recommendations (e.g., naming) that are outside the spec?

I'm ok with those naming requirements and recommendations, although it seems snake case has been our convention so far for json properties (which is what I think "members" refers to).

ryanhofdotgov:
Sorry – I think I'm confused. Is the approach described in this ADR meant to match up with the API spec changes in #1082? 

I was thinking this ADR was really just about adding `data` and `errors` as top level elements, but not sure if you are also proposing adding the properties of `type`, `id`, `attributes`, etc. to make everything under `data` fit the mold of a [JSON API resource object](https://jsonapi.org/format/#document-resource-objects).

I'll throw something on your calendar to chat.

echappen:
@ryanhofdotgov Updated the ADR based on our convo today.

ryanhofdotgov (approved):
This is great – thanks for weighing all the options, tradeoffs, and bike shed colors with us and writing it up!

Issue #628, opened 2021-04-02
Lookup API provides querying state


cjhskippy:
As a state agency worker responsible for processing match results, I want to know which state to contact when I am notified of a match that was initiated by another state's query.

AC:
- [ ] Lookup API returns the state that initiated the match related to that match ID

Depends on #683

echappen:
@cjhskippy @ryanhofdotgov Can you confirm this is the workflow we're solving for here?

Workflow:
1. EW is notified of a match; receives a Lookup ID
2. EW performs lookup ID request; receives PII from original request
3. EW takes that PII and performs a Match query; receives match data that includes the match records for their state and 1-to-N number of records from other states

If there's only one other state in the match response, then this state is the one to be contacted. 

If there are multiple other states, then it's ambiguous which state performed the original match query. Including the state in the response for step 2 above will make this clear and possibly save an extra step for the EW.

Do I have this right?

cjhskippy:
Closed in connection with removal of Lookup API.

PR #1120, opened 2021-06-29
Create table in core database cluster for Lookup API

kingcomma:
- Create new `lookup` database and `lookups` table
- Create AAD managed db role for orchestrator Function
- Modify `create_managed_role` to accept resource group as a parameter

Specific feedback on the `lookups` table schema is welcomed. I envision #628 adding a `state` column.


kingcomma:
Closing this as we shift focus to re-prioritized work.

PR #1123, opened 2021-06-30
Automate configuration of IDP client secrets

### LINKED ISSUE #1039: Semi-automated mechanism for establishing IdP secret for App Service instances
Semi-automated mechanism to set up App Service environment variable from that will come in a separate PR. Follow up to #1028.

Implementation notes:
- Design for future scenario where we have multiple load-balanced instances of the query tool (dashboard too, but that's less likely)
- If setting secret in CLI, don't pass secret via CLI parameter, use `read`, etc. ([like our AzCopy example](https://github.com/18F/piipan/blob/1284f25851be575128655324e51af92ac8bff7a1/etl/docs/uploads.md#azcopy))
- Maybe just a bash script that iterates over the dashboard or query tool instances using their SysTag (see approach elsewhere in script), then set the env variable with value collected at prompt? Run once for dashboard, a second time (second password) for query tool?

timoballard:
Closes #1039 

Adds `configure-oidc.bash`, which, given an environment and application name, sets the `IDP_CLIENT_SECRET` after pulling the current value from the appropriate vault. This approach assumes the existence of a Key Vault which contains an OIDC secret, both following standard naming conventions. 

`configure-oidc.bash` is run as part of IaC for the Dashboard and Query Tool applications, but can be run as a one-off to accommodate updated client secrets.

### DIFF
--- a/docs/iac.md
+++ b/docs/iac.md
@@ -62,12 +62,6 @@ The following environment variables are pre-configured by the Infrastructure-as-
 | `LookupConnectionString` | Azure Storage Account connection string for accessing Table Storage service used for storing lookup IDs. | Piipan.Match.Orchestrator |
 | `LookupTableName` | Name of the Table Storage table where lookup IDs are stored.  | Piipan.Match.Orchestrator |
 
-#### Manually configured
-The following environment variables are not configured by the Infrastructure-as-Code scripts, and thus need to be manually configured after each run.
-| Name | Value | Used by |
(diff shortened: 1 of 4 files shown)


echappen (approved):
LGTM 👍

timoballard:
Going to briefly mark this as a draft -- need to update `IaC` documentation and account for potentially having multiple app instances.

PR #1132, opened 2021-07-01
ADR on using Azure Table storage as NoSQL datastore

kingcomma:
Provides context around our decision to avoid Table storage moving forward, our preference for PostgreSQL, and evaluation criteria if deviating from our preference.

Partially addresses #701.

### DIFF
--- a/docs/adr/0015-do-not-use-table-storage.md
+++ b/docs/adr/0015-do-not-use-table-storage.md
@@ -1,0 +1,29 @@
+# 15. Do not use Table storage
+
+Date: 2021-07-01
+
+## Status
+
+Accepted
+
+## Context
+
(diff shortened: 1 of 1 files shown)


docs/adr/0015-do-not-use-table-storage.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Decision
+
+We will avoid Table storage as a datastore solution.
+
+## Consequences
+
+Moving forward, we will prefer PostgreSQL—[our default database](0004-default-database.md)—or investigate [Cosmos DB](https://docs.microsoft.com/en-us/azure/cosmos-db/introduction) as an alternative NoSQL solution.
+
+Before deviating from PostgreSQL, we will evaluate the following criteria:
```
echappen, line 23: I really like this part 👍

echappen (approved):

docs/adr/0015-do-not-use-table-storage.md
kingcomma: All credit goes to @ryanhofdotgov 😄

PR #1153, opened 2021-07-02
Hack to get Azure logging for App Service working under Windows

ryanhofdotgov:


ryanhofdotgov:
@timoballard I forgot that I didn't merge this in – was looking to see if FastTrack had a different suggestion. Were you able to get log entries to appear for the query tool without this?

timoballard:
@ryanhofdotgov - Nope- I just [borrowed this same approach](https://github.com/18F/piipan/blob/39c09db3836e588ca801da38fd073a07d046dc02/query-tool/src/Piipan.QueryTool/Program.cs#L25-L36) in my branch

PR #1154, opened 2021-07-02
Configure CIS Azure Benchmark policy as part of IaC

### LINKED ISSUE #828: Add CIS Azure Benchmark to Azure policy
Enable CIS Azure Benchmark in Azure policy in TTS subscription to identify additional security backlog items, ideally through an IaC mechanism.

timoballard:
Closes #828 

Adds `configure-cis-policy.bash`, which assigns the CIS Microsoft Azure Foundations Benchmark to both the core and match resource groups.

### DIFF
--- a/iac/configure-cis-policy.bash
+++ b/iac/configure-cis-policy.bash
@@ -1,4 +1,10 @@
 #!/bin/bash
+#
+# Assigns the CIS Microsoft Azure Foundations Benchmark policy set
+# to the core and match resource groups. Assumes an Azure user with
+# the Global Administrator role has signed in with the Azure CLI.
+#
+# usage: configure-cis-policy.bash <azure-env>
 
 # shellcheck source=./tools/common.bash
 source "$(dirname "$0")"/../tools/common.bash || exit
--- a/iac/configure-cis-policy.bash
+++ b/iac/configure-cis-policy.bash
@@ -10,7 +10,9 @@
 source "$(dirname "$0")"/../tools/common.bash || exit
 
 set_constants () {
-  CIS_POLICY_DISPLAY_NAME="[Preview]: CIS Microsoft Azure Foundations Benchmark v1.3.0"
+  # https://docs.microsoft.com/en-us/azure/governance/policy/samples/cis-azure-1-3-0
+  # The policy "name" is the UUID of the set-definition
+  CIS_POLICY_SET_DEFINITION_NAME="612b5213-9160-4969-8578-1518bd2a000c"
 }
 
 main () {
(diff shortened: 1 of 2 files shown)


iac/configure-cis-policy.bash
```diff
@@ -0,0 +1,49 @@
...
+
+  echo "Assigning $cis_policy_set_definition_name to $MATCH_RESOURCE_GROUP"
+  az policy assignment create \
+    --policy-set-definition "$cis_policy_set_definition_name" \
+    --resource-group "$MATCH_RESOURCE_GROUP" \
+    --name "cis-1_3-$MATCH_RESOURCE_GROUP" \
+    --location "$LOCATION" \
+    --assign-identity
+
+  echo "Assigning $cis_policy_set_definition_name to $METRICS_RESOURCE_GROUP"
```
echappen, line 40: $METRICS_RESOURCE_GROUP is the same as $RESOURCE_GROUP at this point, so this action is likely redundant.

timoballard: Oh didn't catch that - thanks @echappen! 

I see `METRICS_RESOURCE_GROUP` used elsewhere in IaC - do you think we should deprecate it entirely, or are we keeping it around in case it takes on a different value from `RESOURCE_GROUP` in the future and/or in other environments?

echappen: We don't need to keep it around—we just renamed it in the ENV files as a short-term solution. I'll make an issue to remove the name from IAC, to eliminate any future confusion.

timoballard:
After some digging, I was only able to target existing policy set-definitions by either hard coding the set-definition `name` (the `name` field is actually a `uuid`) or hard coding the `displayName` ("[Preview]: CIS Microsoft Azure Foundations Benchmark v1.3.0", in this case). Neither of those seems like a totally stable value, but I opted to target the `displayName` because it's less opaque. Open to other ideas!

iac/configure-cis-policy.bash
```diff
@@ -0,0 +1,47 @@
...
+# to the core and match resource groups. Assumes an Azure user with
+# the Global Administrator role has signed in with the Azure CLI.
+#
+# usage: configure-cis-policy.bash <azure-env>
+
+# shellcheck source=./tools/common.bash
+source "$(dirname "$0")"/../tools/common.bash || exit
+
+set_constants () {
+  CIS_POLICY_DISPLAY_NAME="[Preview]: CIS Microsoft Azure Foundations Benchmark v1.3.0"
```
ryanhofdotgov, line 13: I'd use the UUID with an appropriate comment or reference to documentation to where it is defined; we've had to take a similar approach elsewhere:
* [Azure Event Hubs Data Receiver role](https://github.com/18F/piipan/blob/0945cdcc2565a4062cad240e6621c74568d055fa/iac/arm-templates/event-hub-monitoring.json#L25)
* [Azure CLI](https://github.com/18F/piipan/blob/0945cdcc2565a4062cad240e6621c74568d055fa/tools/authorize-cli.bash#L31)
* [Storage Blob Data Contributor role](https://github.com/18F/piipan/blob/0945cdcc2565a4062cad240e6621c74568d055fa/iac/create-state-storage-service-principal.bash#L33)

We know for sure that the display name will change when it gets taken out of Preview, but there's at least the possibility the UUID will stay the same. We can also ask FastTrack if they know if the uuid will be stable.

timoballard: Thanks @ryanhofdotgov! I think "stable" was probably the wrong term here - my main concern was that the `uuid` in Azure Commercial will differ from the `uuid` in Azure Gov. I suppose in that case though, we'd just have to move it into the env configs instead of hard coding it.

ryanhofdotgov: Ah, good point – I've just verified the uuids are the same across `AzureCloud` and `AzureUSGovernment`.

ryanhofdotgov (approved):

PR #1156, opened 2021-07-06
Change rule for consecutive recent benefit months in import schema

echappen:
Recent benefit months do not need to be consecutive. 

For example, if current month is July and an individual received benefits in March, April, and June (but not May), data can be `2021-03 2021-04 2021-06`

## Changelog

Relaxing bulk upload rule for `recent_benefit_months` requiring months needing to be consecutive. Recent Benefit Months no longer need to be consecutive.


ryanhofdotgov (approved):
Thanks!

PR #1188, opened 2021-07-14
Disallow usage of all-access APIM keys

### LINKED ISSUE #684: Prohibit use of master keys across all APIM policies.
APIM automatically creates a "Built-in all-access subscription" with keys that can be used to access any API. This facilitates using the Portal's testing functionality, but poses a significant security risk. These keys can not be disabled but can be prohibited at the API level through a [policy directive](https://medium.com/vipps/azure-api-management-has-a-master-key-ab4658cafe02).

This directive should be added to all policies.

Potentially also relevant if we begin using APIM "products." When a new product is created a corresponding subscription for that product is also created. This subscription can be suspended/deleted, but more research is needed to confirm that is sufficient.

timoballard:
Closes #684 

Adds `apim-policy.xml` prevents the usage of the automatically generated all-access APIM subscription keys. Attempts to use these keys for API calls results in the same `401` response that would be returned with an invalid subscription key.

Updates IaC to apply `apim-policy.xml` to the APIM globally.

### DIFF
--- a/iac/apim-policy.xml
+++ b/iac/apim-policy.xml
@@ -1,0 +1,22 @@
+<!-- Generated via IaC -->
+
+<policies>
+    <inbound>
+        <choose>
+            <when condition="@(context.Subscription != null &amp;&amp; context.Subscription.Id == &quot;master&quot;)">
+                <return-response>
+                    <set-status code="401" reason="Access Denied" />
+                    <set-header name="Content-Type" exists-action="override">
+                        <value>application/json; charset=UTF-8</value>
+                    </set-header>
+                    <set-body>{"statusCode": 401, "message": "Access denied due to invalid subscription key. Make sure to provide a valid key for an active subscription."}</set-body>
+                </return-response>
+            </when>
+        </choose>
+    </inbound>
+    <backend>
+        <forward-request />
+    </backend>
+    <outbound />
+    <on-error />
+</policies>
--- a/iac/apim-policy.xml
+++ b/iac/apim-policy.xml
@@ -2,6 +2,7 @@
 
 <policies>
     <inbound>
+        <!-- APIM issues a "master" key that is scoped to the entire instance, can be used as a valid API key for any API, and cannot be disabled. Prevent this key from being used. As a consequence, the "test" feature in the Azure portal, which relies on this key, is no longer usable. -->
         <choose>
             <when condition="@(context.Subscription != null &amp;&amp; context.Subscription.Id == &quot;master&quot;)">
                 <return-response>
--- a/iac/arm-templates/apim.json
+++ b/iac/arm-templates/apim.json
@@ -43,6 +43,9 @@
         },
         "eventHubName": {
             "type": "String"
+        },
+        "apimPolicyXml": {
+            "type": "String"
         }
     },
     "variables": {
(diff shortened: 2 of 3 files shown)


iac/apim-policy.xml
```diff
@@ -0,0 +1,22 @@
+<!-- Generated via IaC -->
+
+<policies>
+    <inbound>
+        <choose>
```
kingcomma, line 5:
```suggestion
        <!-- APIM issues a "master" key that is scoped to the entire instance, can be used as a valid API key for any API, and cannot be disabled. Prevent this key from being used. As a consequence, the "test" feature in the Azure portal, which relies on this key, is no longer usable. -->
        <choose>
```

*Suggestion:* I think we should document this policy briefly since it may not be intuitive to a future developer. I suggested some language just to be helpful, don't feel beholden to it.

iac/arm-templates/apim.json
```diff
@@ -119,6 +122,18 @@
                 "versioningScheme": "Segment"
             }
         },
+        {
+            "type": "Microsoft.ApiManagement/service/policies",
```
kingcomma, line 126: *Question:* Just so I understand, if I'm looking at this in the portal the policy would be applied at the "All APIs" level?

iac/arm-templates/apim.json
```diff
@@ -119,6 +122,18 @@
                 "versioningScheme": "Segment"
             }
         },
+        {
```
kingcomma, line 125: *Thought:* The [test feature](https://docs.microsoft.com/en-us/azure/api-management/api-management-howto-api-inspector) in the APIM portal displays debugging information that can be handy for diagnosing issues related to our APIs. I wonder if it's worth [conditionally deploying](https://docs.microsoft.com/en-us/azure/azure-resource-manager/templates/conditional-resource-deployment) this policy so we have access to the test features in at least `tts/dev`?

Tagging @ryanhofdotgov since I'm not sure if this is an acceptable security tradeoff.

iac/arm-templates/apim.json
timoballard: Correct!

iac/arm-templates/apim.json
timoballard: Yeah that's a good thought. If this approach to APIM testing is relatively infrequent, it's pretty straightforward (albeit a bit clunky) to disable it temporarily from the portal by commenting out this policy directive in the code editor

kingcomma: Also a good point! I'm good with that approach for now and if it becomes a burden we can take a stab at the conditional deployment approach.

kingcomma (approved):
👍 Thanks for picking this up!

iac/arm-templates/apim.json
ryanhofdotgov: My hot take is that if we are conditionally disabling a security mechanism for convenience in a developer environment, I'd want some process/tooling to really make sure it wasn't accidentally disabled in production (e.g., a documented test plan and perhaps integration test); i.e., more than just a conditional in an ARM template checking for a certain value.

If you're right there in the portal I can see how it is convenient, but just for clarity, does this APIM feature allow you to do anything that you can't do with a developer-specific subscription key and Postman?

kingcomma: In my experience, the primary benefit of the feature is that it conveniently presents debugging information that would otherwise be spread across multiple resources / log streams (but which would technically be available to a developer). For example, it offers a [trace](https://docs.microsoft.com/en-us/azure/api-management/api-management-howto-api-inspector#review-trace-information) pane that shows you information about a request from start to finish — the request body, the policy transformations that were applied, the request object that was forwarded on to the backend, the backend response, etc.

timoballard: Thanks for the input on this!

ryanhofdotgov: Ah, that trace function is pretty cool.

PR #1199, opened 2021-07-20
Update query tool to handle household match requests/responses

### LINKED ISSUE #1121: Fix query tool after breaking changes in API
Update the query tool to be able to continue working now that API supports querying for an entire household.

No UI changes are required - when this is done, the website will still only accept one person at a time.

- [ ] Query tool works after #368 is merged

echappen:
Closes #1121 
Relies on #1098

This is a quick fix to ensure there are no breaking changes from the API switch to households.
No new UI was added.
Validation and other data-level errors aren't being shown in the UI.


kingcomma (approved):
👍

Issue #1170, opened 2021-07-08
Log APIM subscription key name in Orchestrator


kingcomma:
Follow on to #683. Once the APIM subscription name is available as an HTTP header, add logging to the Orchestrator application for calls to the match query method (but not the lookup API).

Consider using middleware approach established in #1207.

PR #1200, opened 2021-07-21
Add state username to Duplicate Participation API requests

### LINKED ISSUE #889: Add state username to duplicate participation API
The near real-time APIs are designed to be incorporated into a state backoffice system and triggered by actions by a state eligibility worker. (This is in contrast to the bulk upload API, which is intended to be automated).

To support cross-organizational auditing (AU-16), extend the DupPart API to include a field for the state system to supply the state worker username on whose behalf they are making the query for. Make it optional – states SHOULD specify the username, not MUST.

AC:
- [ ] States can provide the username to match query API calls
~~States can provide username to lookup API calls~~ no longer needed since lookup API is going away
- [ ] Username for match API is forwarded to audit logs
- [ ] API documentation is updated

echappen:
Closes #889
Relies on #1098 and #1082

I added the username as a headers param for consistency between `get` and `post` requests. 

### Adds 
Adds header parameter to include a state username for Duplicate Participation API


match/src/Piipan.Match.Orchestrator/Api.cs
```diff
@@ -146,6 +149,12 @@ public Api(IAuthorizedApiClient apiClient, ITableStorage<QueryEntity> lookupStor
         {
             log.LogInformation("Executing request from user {User}", req.HttpContext?.User.Identity.Name);
 
+            string? username = req.Headers["X-username"];
+            if(username is string)
+            {
+                log.LogInformation("with username {Username}", username);
+            }
+
```
kingcomma, line 157: A couple questions for my own benefit stemming from my lack of familiarity with AU-16.

*Question (non-blocking):* Is there a difference from a compliance standpoint if this message is only logged some times vs. always logged (with `null` or another string indicating the header is not present)?

*Question (non-blocking):* To differentiate from the different user-level audit logging, is it helpful if this message is explicit about this being an _external_ user? Once #1170 is complete, we'll be logging "user" info at three levels:
- Managed identity
- APIM subscription name
- External username

kingcomma (commented):
@echappen It looks like this only applies to Lookup calls. Is the plan to add Match calls in a subsequent PR?

echappen:
@kingcomma I think the match check got removed in my big squash/merge. I'll re-add.

match/src/Piipan.Match.Orchestrator/Api.cs
echappen: My reading of AU-16(1) is that some sort of user identifying info is required in logs. Since #889 specifies username as optional, I'm taking that to mean the other "user" logging is sufficient for what's required, and a username would just be supplementary. But I'd defer to @ryanhofdotgov on that.

ryanhofdotgov: Here's the motivating scenario:
- an API call is being made hundreds of times second, over and over for the same values. It's outside normal usage.
- we want to be able to look at logs and see the source of the API call, under what APIM subscription account
- once there, we want to support cross-organizational audit by also supplying the external user name

The managed identity involved is good to have for completeness and to make sure there isn't security issue around that, but the APIM and external user id is the most important.

If the API call is being made as a direct result of an EW clicking a button, then integrators should be able to easily populate that field. But sometimes the API call will under some type of automated control. In which, case I would hope the user agent who help in an audit.

ryanhofdotgov:
At risk of bike-shedding, it would be good to make the header name more descriptive and to disambiguate it from a value that would be used for authentication/authorization.

I've seen `X-On-Behalf-Of` used elsewhere. And if you want to go for a less descriptive name but stick with an official HTTP/1.1 header, [`From` is almost exactly what we set out to do](https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.22). Only difference is that it is described as an email address, though we could be less strict on that.

echappen:
@ryanhofdotgov Thanks for the ideas. Looking into `X-On-Behalf-Of`... Azure docs describe an ["On Behalf Of" flow](https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-on-behalf-of-flow) for their OAuth process, so using that name may be misleading or conflict with their own terminology. 

I'm ok with using the more official `From` and specifying in our docs that it can be some text other than email. W3 docs list that the value SHOULD be an email address, so we'd still be compliant there.

match/docs/openapi/lookup/id.yaml
```diff
@@ -4,6 +4,12 @@ get:
     - "Lookup"
   summary: "Get the original match data related to a Lookup ID"
   description: "User can provide a Lookup ID and receive the match data associated with it"
+  parameters:
+    - in: header
+      name: From
+      schema:
+        type: string
+      description: "The username of the individual making the request"
```
ryanhofdotgov, line 12:
```suggestion
      description: "As in the HTTP/1.1 RFC, used for logging purposes as a means for identifying the source of invalid or unwanted requests. The interpretation of this field is that the request is being performed on behalf of the state government-affiliated person whose email address (or username) is specified here. It is not used for authentication or authorization."
```
We should clarify some more; I've copied and modified the `From` description from the HTTP RFC in the suggestion above.

match/docs/openapi/orchestrator/query.yaml
```diff
@@ -4,6 +4,12 @@ post:
     - "Match"
   summary: "Search for all matching PII records"
   description: "Queries all state databases for any PII records that are an exact match to the last name, date of birth, and social security number of persons provided in the request."
+  parameters:
+    - in: header
+      name: From
+      schema:
+        type: string
+      description: "The username of the individual making the request"
```
ryanhofdotgov, line 12:
```suggestion
      description: "As in the HTTP/1.1 RFC, used for logging purposes as a means for identifying the source of invalid or unwanted requests. The interpretation of this field is that the request is being performed on behalf of the state government-affiliated person whose email address (or username) is specified here. It is not used for authentication or authorization."
```
We should clarify some more; I've copied and modified the `From` description from the HTTP RFC in the suggestion above.

match/src/Piipan.Match.Orchestrator/Api.cs
```diff
@@ -47,6 +50,12 @@ public Api(IAuthorizedApiClient apiClient, ITableStorage<QueryEntity> lookupStor
             {
                 log.LogInformation("Executing request from user {User}", req.HttpContext?.User.Identity.Name);
 
+                string? username = req.Headers["From"];
+                if (username is string)
+                {
+                    log.LogInformation("with username {Username}", username);
```
ryanhofdotgov, line 56:
```suggestion
                    log.LogInformation("on behalf of {Username}", username);
```

match/src/Piipan.Match.Orchestrator/Api.cs
```diff
@@ -146,6 +155,12 @@ public Api(IAuthorizedApiClient apiClient, ITableStorage<QueryEntity> lookupStor
         {
             log.LogInformation("Executing request from user {User}", req.HttpContext?.User.Identity.Name);
 
+            string? username = req.Headers["From"];
+            if(username is string)
+            {
+                log.LogInformation("with username {Username}", username);
```
ryanhofdotgov, line 161:
```suggestion
                log.LogInformation("on behalf of {Username}", username);
```

match/tests/Piipan.Match.Orchestrator.Tests/LookupTests.cs
```diff
@@ -65,9 +67,15 @@ public async void SuccessfulApiCall()
         {
             // Arrange
             var api = ConstructMockedApi();
+            var mockRequest = new Mock<HttpRequest>();
+            var headers = new HeaderDictionary(new Dictionary<String, StringValues>
+            {
+                { "X-username", "foobar"}
```
ryanhofdotgov, line 73:
```suggestion
                { "From", "foobar"}
```

ryanhofdotgov (approved):
A few suggestions to clarify documentation and to make the logs easier to understand by someone external to the development team.

Issue #1125, opened 2021-06-30
Design spike: sign-in page/sign-out page for dashboard/query tool


ryanhofdotgov:
Consider and spec out what user experience we want for the dashboard/query tool applications with respect to sign-in and sign-out. 

Currently the Easy Auth-protected dashboard and query tool apps do not have a public sign-in page and will immediately redirect to the OIDC identity provider. 

Also, currently they do not have a custom logout page, but just show the default Easy Auth logout page, which looks like this:
<img width="400" alt="Screen Shot 2021-06-30 at 3 42 18 PM" src="https://user-images.githubusercontent.com/37100189/124040434-c4de1d00-d9b9-11eb-81dd-95fe13325a5c.png">

Depending on the outcome of this spike, follow-on technical implementation might include:
- [ ] exposing public sign-in page, explicitly disabling Easy Auth for that URL
- [ ] exposing public custom sign-out page, explicitly disabling Easy Auth for that URL
- [ ] updating logout URL to specify `post_logout_redirect_uri`, per [Customize sign-in and sign-out in Azure App Service authentication documentation](https://docs.microsoft.com/en-us/azure/app-service/configure-authentication-customize-sign-in-out#sign-out-of-a-session)

juliaklindpaintner:
Let me know if you want UX/UI input on this!

ryanhofdotgov:
@juliaklindpaintner Absolutely – this ticket has you in mind, but will likely need you to pull in @timoballard for a bit to get you started.

PR #1226, opened 2021-07-28
Add authentication exception for health probes

### LINKED ISSUE #1131: Configure Easy Auth exception for Front Door health check probe
Front Door health check probe will fail against dashboard/query tool when easy auth is enabled. Various blog post seem to indicate that solution is to create a page that is excluded from Easy Auth that Front Door can "ping". Awaiting FastTrack feedback.

Could be related to follow on of #1125 (see implementation notes).

kingcomma:
- Add an empty, static HTML file to each app
- Add static file to excludedPaths in ARM template auth settings
- Point Front Door to excluded path

Depending on the outcome of #1125, this approach can be easily adapted by pointing to a new path and removing the static HTML file.

Closes #1131.

### DIFF
--- a/iac/arm-templates/dashboard-app.json
+++ b/iac/arm-templates/dashboard-app.json
@@ -161,7 +161,8 @@
                 "globalValidation": {
                     "requireAuthentication": true,
                     "unauthenticatedClientAction": "RedirectToLoginPage",
-                    "redirectToProvider": "oidcProvider"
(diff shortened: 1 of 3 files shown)


timoballard (approved):
LGTM!

kingcomma:
Updated IaC has been run against `tts/dev`

PR #1241, opened 2021-08-02
Extract OIDC user claims in Dashboard

### LINKED ISSUE #1126: Extract OIDC claims/user info for display and logging in the dashboard
Using the information conveyed to the dashboard through its configured identity provider and Easy Auth OIDC mechanism:
- extract out the currently signed in user's email address and display in upper right corner of application, to the left of the sign out link
- extract out all the claims of the currently signed in user and log it to the App Service log at the `INFO` level

Implementation notes:
- Reference: [Work with user identities in Azure App Service authentication](https://docs.microsoft.com/en-us/azure/app-service/configure-authentication-user-identities)
- Will likely need to use [MaximeRouiller.Azure.AppService.EasyAuth](https://github.com/MaximRouiller/MaximeRouiller.Azure.AppService.EasyAuth)
- Follow on work is to [add same functionality to query tool](#1127)

timoballard:
Closes #1126 

Extracts the authenticated user's email from OIDC claims and displays in the upper right.

Reuses much of #1207 via `Piipan.Shared`


dashboard/src/Piipan.Dashboard/Startup.cs
```diff
@@ -42,6 +49,20 @@ public void ConfigureServices(IServiceCollection services)
...
+            services.AddEasyAuth();
+            
+            services.AddDistributedMemoryCache();
+            services.AddSession();
+
+            services.AddTransient<IClaimsProvider, ClaimsProvider>();
+
+            if (_env.IsDevelopment())
+            {
+                var mockFile = $"{_env.ContentRootPath}/mock_user.json";
```
ryanhofdotgov, line 63: Could you add the bit that excludes this file from builds? Thanks! Unlikely to ever cause any problems, just a defensive programming.

timoballard: Good catch - thanks!

ryanhofdotgov (approved):
Thanks!

PR #1242, opened 2021-08-02
Remove lookup api from query tool

### LINKED ISSUE #1234: Update Query Tool to remove Lookup API
Now that Lookup API is removed, we should adjust the Query Tool to prevent breaking changes. 

_Originally posted by @kingcomma in https://github.com/18F/piipan/issues/1228#issuecomment-889946723_

echappen:
Closes #1234


query-tool/src/Piipan.QueryTool/IQueryable.cs
```diff
@@ -6,6 +6,5 @@ namespace Piipan.QueryTool
 {
     public interface IQueryable
     {
-        string LookupId { get; set; }
     }
 }
```
kingcomma, line 10: *Chore:* I think the `IQueryable` class (and .cs file) can be dropped entirely now? Same goes for dropping the inheritance in `PiiRecord.cs`.

kingcomma (commented):
*Chore:* I think the `IQueryable` class (and the associated .cs file) can be dropped entirely

timoballard (approved):

PR #1302, opened 2021-08-04
Add Sign out link to app services

### LINKED ISSUE #1124: Add sign-out link to App Services that signs out of Easy Auth
Add a sign-out link to upper right corner of:
- [x] Dashboard
- [x] Query tool

that simply calls `/.auth/logout` per the [Customize sign-in and sign-out in Azure App Service authentication documentation](https://docs.microsoft.com/en-us/azure/app-service/configure-authentication-customize-sign-in-out#sign-out-of-a-session).

Note:
At this time, we are only logging out of the app itself, not all eAuth-authenticated apps.

timoballard:
Closes #1124 

Adds a sign out menu item to the upper right of both the Dashboard and Query Tool. The link calls `/.auth/logout`, which terminates the app service session, but does not sign the user out of the identity provider.

## Notes
- There is no authentication layer in local development mode, so clicking the sign out link currently results in a 404. We could conditionally configure a page at the `/.auth/logout` route when in development mode, but as of now that is not being done.


echappen (changes requested):
[request] I don't anticipate this being a big use case, but the dropdown is hidden on screen widths 1024px and below. I think we can defer on styling for now (or use USWDS's Mobile Menu), but it'd be good to have some way to log out at smaller sizes.

echappen (approved):
👍 Thanks!

PR #1327, opened 2021-08-06
Log APIM subscription name in Orchestrator

### LINKED ISSUE #1170: Log APIM subscription key name in Orchestrator

kingcomma:
- Log the custom subscription name header established in #1171 
- Remove nullable reference type (and directive) to clear up compiler warnings

If/when we make the move to .NET 5/6 we should be able to clean up all this logging using either some custom [middleware](https://docs.microsoft.com/en-us/azure/azure-functions/dotnet-isolated-process-guide#middleware) or perhaps something similar to the [ASP.NET `AddHttpLogging` method](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/http-logging/?view=aspnetcore-6.0#http-logging-options) (if it exists for Function apps).

Closes #1170

### DIFF
--- a/match/src/Piipan.Match.Orchestrator/Api.cs
+++ b/match/src/Piipan.Match.Orchestrator/Api.cs
@@ -13,8 +13,6 @@ using Newtonsoft.Json;
 using Piipan.Match.Shared;
 using Piipan.Shared.Authentication;
 
-#nullable enable
-
 namespace Piipan.Match.Orchestrator
 {
     /// <summary>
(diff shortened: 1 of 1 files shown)


ryanhofdotgov (approved):
LGTM

PR #1328, opened 2021-08-06
Add OIDC claim-based authorization policy to Dashboard

### LINKED ISSUE #1128: Add OIDC claim-based policy enforcement to dashboard
After #1126 and #1127 is completed, add a mechanism to the dashboard to limit access to OIDC authenticated users that have one or more specific claims for:

- identity assurance level
- authentication assurance level
- presence of specific roles

Implementation notes:
- Needs to be flexible to permit configurable key/value pairs
- Needs to be flexible around number of claims being checked (e.g., 0 or 1 or 3, etc.)
- Can likely keep boolean logic simple; i.e., all rules are AND'ed together to get a authz decision
- Will need to exact same mechanism (with different config values) in query tool as well

timoballard:
Closes #1128

Adds a default authorization policy to the Dashboard that requires the user to have particular claims in order to access the application. The required claims are configurable via `appsettings`. If no authorization policy is configured in `appsettings`, all requests are rejected with `403 - Forbidden`.


dashboard/src/Piipan.Dashboard/Startup.cs
```diff
@@ -56,6 +59,22 @@ public void ConfigureServices(IServiceCollection services)
             services.AddDistributedMemoryCache();
             services.AddSession();
 
+            services.AddAuthorization(options => {
+                var authzPolicyOptions = Configuration
+                    .GetSection(AuthorizationPolicyOptions.SectionName)
+                    .Get<AuthorizationPolicyOptions>();
+
+                if (!(authzPolicyOptions is null))
```
ryanhofdotgov, line 67: I'm not familiar with this API yet, but it looks like it may fail open (i.e., grant access) if someone forgets to explicitly set the authorization policy in the app configuration. If so, would you consider re-writing so [it is more defensive and fails closed](https://blogs.keysight.com/blogs/tech/nwvs.entry.html/2020/05/20/fail_closed_failop-ZYAt.html)?

timoballard: Great catch - reworked this a bit per your advice

dashboard/src/Piipan.Dashboard/Startup.cs
```diff
@@ -28,9 +30,10 @@ public Startup(IConfiguration configuration, IWebHostEnvironment env)
         // This method gets called by the runtime. Use this method to add services to the container.
         public void ConfigureServices(IServiceCollection services)
         {
-            services.Configure<ClaimsOptions>(Configuration.GetSection(ClaimsOptions.Claims));
-
-            services.AddRazorPages();
+            services.Configure<ClaimsOptions>(Configuration.GetSection(ClaimsOptions.SectionName));
+            services.AddRazorPages(options => {
+                options.Conventions.AuthorizeFolder("/");
```
ryanhofdotgov, line 35: Any chance you know how this interacts with the [`excludedPaths` directive in the ARM template](https://github.com/18F/piipan/blob/09eff1b13fc37391b3e60dad866a0252cf169431/iac/arm-templates/dashboard-app.json#L172) we added for health checks?

dashboard/src/Piipan.Dashboard/appsettings.tts.json
```diff
@@ -1,5 +1,8 @@
 {
   "Claims": {
       "Email": "extension_EmailAddress"
+  },
+  "AuthorizationPolicy": {
+    "RequiredClaims": []
```
ryanhofdotgov, line 6: Thought: what will happen if a user [incorrectly] does not have an email claim?

dashboard/src/Piipan.Dashboard/Startup.cs
timoballard: Yes- the `/health-probe.html` path is a static page (not a razor page), so this authorization doesn't apply to it - verified that the health page is still accessible for an otherwise unauthorized user here: https://tts-app-dashboard-tdb.azurewebsites.net/health-probe.html

ryanhofdotgov:
[Linking to Slack thread we used while GitHub wasn't behaving...](https://gsa-tts.slack.com/archives/C018P1ZJPEC/p1628610472010600?thread_ts=1628610156.010300&cid=C018P1ZJPEC):
* As part of this PR: add email as required claim, test `BuildAuthPolicy`
* Deferring for discussion: [should we simplify access model and route all static files through ASP.NET](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/static-files?view=aspnetcore-5.0#static-file-authorization)

ryanhofdotgov (approved):
Nice work! And quick turn around from the feedback too 👍

PR #1357, opened 2021-08-11
Explicitly disable token store and document

ryanhofdotgov:
Normally we would not explicitly disable a feature that was already turned off by default, but as it was very useful for debugging OIDC claim issues, it seems appropriate to leave a "bread crumb".

[Enabling the token store enables the /.auth/me URL that prints the current user's OIDC claims.](https://docs.microsoft.com/en-us/azure/app-service/configure-authentication-user-identities#access-user-claims-using-the-api)

### DIFF
--- a/iac/arm-templates/dashboard-app.json
+++ b/iac/arm-templates/dashboard-app.json
@@ -194,6 +194,13 @@
                             }
                         }
                     }
+                },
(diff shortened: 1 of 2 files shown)


timoballard (approved):
👍

PR #1374, opened 2021-08-12
Remove exceptions field from API's

### LINKED ISSUE #1341: Remove "exceptions" field from bulk uploads
The bulk uploads currently have an "exception" field.  Its purpose has been taken on by the "protect_location" field.

We want these extraneous fields to be removed so that states have clear guidance on what to provide and we don't accidentally encourage incorrect use.

echappen:
Closes #1341  

Affects ddl

- [ ] Run iac on tts/dev prior to merging

### Removes
- Removes `exception` column from Bulk Upload csv
- Removes `exception` field from Duplicate Participation API responses


ryanhofdotgov (approved):
LGTM – thanks!

PR #1377, opened 2021-08-13
Update B2C docs for creating, updating user claims

### LINKED ISSUE #1336: Update b2c docs with info on how to add attributes to users
- How do you set up new attribute
- Is there a way to just set it in the Azure portal, or do you have to enter in the profile editing user flow for each user?

timoballard:
Closes #1336

Updates the Azure B2C IDP documentation to include instructions for creating new user attribute types (claim types) as well as updating attributes for specific users (claim values).


ryanhofdotgov (approved):
Thanks!

Issue #1029, opened 2021-06-09
Reconfigure Front Door on App Services, but with Easy Auth


ryanhofdotgov:
In PR #1028, we enabled Easy Auth on the query tool dashboard and open up direct access to the App Service instances, going around front door.

Now, with some nudging in the right direction from Fast Track, [incorporate these settings](https://gsa-tts.slack.com/archives/C018P1ZJPEC/p1627074364005500), and restrict access to the query tool/dashboard to Front Door, reversing the relevant portions of PR #1028.

~This is required due to the nature of how Front Door functions – in short, using Easy Auth and Front Door together requires a custom domain/custom certificate. This is described in at least a couple of different places:~
* ~https://edi.wang/post/2020/6/29/solving-azure-ad-sign-in-failure-with-azure-front-door~
* ~https://vincentlauzon.com/2019/07/17/azure-app-service-authentication-with-azure-front-door/~

~After we obtain custom domains/custom certificates, restrict access to the query tool/dashboard to Front Door, reversing the relevant portions of PR #1028.~

ryanhofdotgov:
Revisit this now that implementation has moved from AAD to OIDC, follow up with FastTrack to see if they were able to come up with a working example of this working without custom domains.

timoballard:
Moving from AAD to OIDC does not appear to resolve this issue. Following up with FastTrack for next steps

ryanhofdotgov:
Moving out to backlog, pending insight from FastTrack.

ryanhofdotgov:
Closed by #1380 and #1381. New issue cropped up though: #1383

PR #1380, opened 2021-08-13
Configure Easy Auth and Front Door for query tool

ryanhofdotgov:
Partially addresses #1029; similar config will be done on dashboard in follow-on PR.

The sign-out link however is now broken; wants to redirect to App Service URL for ../complete rather than the Front Door URL. Will address in separate PR.

**Deploy notes**
- Updated B2C app registration to allow Front Door callback URL, deleted App Service callback URL


timoballard (approved):
Tested matching changes to Dashboard in TTS sandbox 👍

PR #1381, opened 2021-08-13
Configure Easy Auth and Front Door for dashboard

ryanhofdotgov:
Partially addresses #1029; similar to work done for query tool in #1380

The sign-out link however is now broken; wants to redirect to App Service URL for ../complete rather than the Front Door URL. Will address in separate PR.

**Deploy notes**
- Updated B2C app registration to allow Front Door callback URL, deleted App Service callback URL


timoballard (approved):
Tested in TTS sandbox 👍

PR #1394, opened 2021-08-16
Fix: add hidden legend to form fieldset

echappen:
Fixes [nightly pa11y ci runs](https://app.circleci.com/pipelines/github/18F/piipan/2112/workflows/c6696e16-e062-456f-81d7-10ee4f1255ff/jobs/5489) on querytool

Pa11y flags form fieldsets without a legend. Because legend text would be redundant with the h2, the legend is hidden using a [technique](https://pauljadam.com/demos/fieldsetlegend.html) that will hide it visually but keep it findable by assistive technologies.


query-tool/src/Piipan.QueryTool/Pages/Index.cshtml
```diff
@@ -8,6 +8,8 @@
     <h2>Search for SNAP participants</h2>
     <form asp-page="Index" class="usa-form">
         <fieldset class="usa-fieldset">
+            @* must have a legend in a fieldset for a11y purposes *@
+            <legend style="position:absolute; clip:rect(0 0 0 0);">Search for SNAP participants</legend>
```
kingcomma, line 12: *Thought (non-blocking):* I can't find great documentation, but it looks like [USWDS has a `usa-sr-only` utility class](https://designsystem.digital.gov/components/search/#accessibility-search) that I think offers the same functionality and is maybe a touch more self-documenting.

kingcomma:
```suggestion
            <legend style="position:absolute; clip:rect(0 0 0 0);">Participant information</legend>

```
*Thought (non-blocking):* Something like "Participant information" may be more aligned with the idea that the `<legend>` [describes the form controls which are being grouped](https://www.w3.org/WAI/tutorials/forms/grouping/#associating-related-controls-with-fieldset) and would eliminate the redundant "Search for SNAP participants" read out in some screen readers.

kingcomma (approved):
Thanks for digging into this! Just a couple non-blocking thoughts/suggestions.

query-tool/src/Piipan.QueryTool/Pages/Index.cshtml
echappen: Great suggestions, thanks!

PR #1405, opened 2021-08-19
Make sign out links work with Azure Front Door

### LINKED ISSUE #1383: Front Door causes sign-out button to break
#1380 and #1381 adds Front Door reverse proxy to our web apps now that we have figured out the obscure settings to make it work.

However, `/.auth/logout` will not use the Front Door URL when Easy Auth for the App Service redirects to `/.auth/logout/complete`. Instead it uses the App Service hostname. Which results in a 403 since we've limited access to app through Front Door.

As a quick test, I tried something like:
```
diff --git a/query-tool/src/Piipan.QueryTool/Pages/Shared/_Layout.cshtml b/query-tool/src/Piipan.QueryTool/Pages/Shared/_Layout.cshtml
index 4446f58..ec4891b 100644
--- a/query-tool/src/Piipan.QueryTool/Pages/Shared/_Layout.cshtml
+++ b/query-tool/src/Piipan.QueryTool/Pages/Shared/_Layout.cshtml
@@ -35,7 +35,10 @@
                                     </button>
                                     <ul id="nav-user" class="usa-nav__submenu" hidden>
                                         <li class="usa-nav__submenu-item">
-                                            <a href="/.auth/logout">Sign out</a>
+                                            @{
+                                                var baseUrl = $"{this.Context.Request.Scheme}://{this.Context.Request.Host}";
+                                                <a href="/.auth/logout?post_logout_redirect_uri=@(baseUrl)/.auth/logout/complete">Sign out</a>
+                                            }
                                         </li>
                                     </ul>
                                 </li>
```

This doesn't do what I want either – it's not just the Easy Auth layer, the App Service doesn't know it's being proxied.

I tried playing around with Backend host header in Front Door – setting to blank did not seem to change App Service.

`X-Forwarded-Host` might be part of the solution. See [[1](https://github.com/MicrosoftDocs/azure-docs/issues/16518)] and [[2](https://docs.microsoft.com/en-us/azure/frontdoor/front-door-http-headers-protocol#2-front-door-to-backend)] and [[3](https://github.com/dotnet/AspNetCore.Docs/issues/2384#issuecomment-297980490)]. Need to figure out how to to tell App Service/Easy Auth what FQHN should be using when it creates URLs.

timoballard:
Closes #1383 

- Configures [forwarded headers middleware](https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/proxy-load-balancer?view=aspnetcore-3.1) so that the `post_logout_redirect_uri` points back to Azure FD rather than the App Service. 
- Pulls properties that are references in shared Razor pages (e.g. `_Layout.cshtml`) up into `BasePageModel` rather than having them duplicated across `PageModels`

### DIFF
--- a/dashboard/src/Piipan.Dashboard/Pages/Shared/_Layout.cshtml
+++ b/dashboard/src/Piipan.Dashboard/Pages/Shared/_Layout.cshtml
@@ -29,7 +29,7 @@
                         </button>
                         <ul id="nav-user" class="usa-nav__submenu" hidden>
                             <li class="usa-nav__submenu-item">
-                                <a href="/.auth/logout">Sign out</a>
+                                <a href="/.auth/logout?post_logout_redirect_uri=@(Model.BaseUrl)/.auth/logout/complete">Sign out</a>
                             </li>
                         </ul>
                     </li>
--- a/dashboard/src/Piipan.Dashboard/Pages/BasePageModel.cs
+++ b/dashboard/src/Piipan.Dashboard/Pages/BasePageModel.cs
@@ -1,0 +1,24 @@
+using Microsoft.AspNetCore.Mvc.RazorPages;
+using Piipan.Shared.Claims;
+
+namespace Piipan.Dashboard.Pages
+{
+    public class BasePageModel : PageModel
+    {
+        private readonly IClaimsProvider _claimsProvider;
+
+        public BasePageModel(IClaimsProvider claimsProvider)
+        {
+            _claimsProvider = claimsProvider;
+        }
+
+        public string Email 
+        { 
+            get { return _claimsProvider.GetEmail(User); }
+        }
+        public string BaseUrl
+        {
+            get { return $"{Request.Scheme}://{Request.Host}"; }
+        }
+    }
+} 
--- a/dashboard/src/Piipan.Dashboard/Pages/Index.cshtml.cs
+++ b/dashboard/src/Piipan.Dashboard/Pages/Index.cshtml.cs
@@ -4,22 +4,19 @@ using Piipan.Shared.Claims;
 
 namespace Piipan.Dashboard.Pages
 {
-    public class IndexModel : PageModel
+    public class IndexModel : BasePageModel
     {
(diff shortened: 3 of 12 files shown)


ryanhofdotgov (approved):
LGTM – thanks!

Is there room to share even more code between the apps, or is it a better tradeoff to repeat?

dashboard/src/Piipan.Dashboard/Pages/Shared/_Layout.cshtml
```diff
@@ -29,7 +29,7 @@
                         </button>
                         <ul id="nav-user" class="usa-nav__submenu" hidden>
                             <li class="usa-nav__submenu-item">
-                                <a href="/.auth/logout">Sign out</a>
+                                <a href="/.auth/logout?post_logout_redirect_uri=@(Model.BaseUrl)/.auth/logout/complete">Sign out</a>
```
ryanhofdotgov, line 32: nitpick (non-blocking): just occurred to me – now that the forwarding middleware is in place, do you still need to explicitly reference the FQHN of Front Door here?

timoballard: Great thought- intuitively I would have expected that to _just work_ with the forwarding middleware, but by default it does try to redirect back to the app service url if we're not explicit here. Poking around a bit to see if there's a framework- or configuration-level approach here that would work as well

timoballard:
@ryanhofdotgov - RE: potential to share more code between apps - this was on my mind as well during implementation here. Some scattered thoughts: 

I think there are a few straightforward opportunities for more sharing - `BasePageModel` for example is exactly the same between the Dashboard & Query Tool. We could also pull out some of the configuration steps in `Startup.cs` into shared extension methods or something similar.

Going a bit further, we could see about pulling up much of what currently lives in the `/Pages/Shared` directories in the two apps into a shared app services library (thinking specifically about things like _Layout, UsaBanner, BasePageModel, etc). That way the specific app service projects are more focused and not full of boilerplate

Issue #1339, opened 2021-08-10
Per-state API removal


cjhskippy:
Upon completion of #1229, implement strategy decided upon

kingcomma:
Copying over rough plan from #1229:

- [ ] Create role for orchestrator's system-assigned identity and grant read-only access for each per-state database (possibly using same approach as in [`create-core-databases.bash`](https://github.com/18F/piipan/blob/fa0a46428e06180eb7cd8474e14409f67e5d8b6d/iac/create-core-databases.bash#L247-L250)). May require changing when certain resources are created by the IaC.
- [ ] Bind each per-state database to the orchestrator function app using an environment variable.
- [ ] Update orchestrator application to connect to each per-state database using a system-assigned identity and use SQL query to find any matches for each person in a household request.
- [ ] Update documentation and system diagram to reflect new operating state.
- [ ] Remove any dangling infrastructure, configuration, docs, files (e.g., per-state function apps, environment variables, service principals, API specs).

PR #1406, opened 2021-08-19
Add read-only role to participants DB cluster

kingcomma:
- Create `NOLOGIN` role with permission to connect to each state database
- Grant role `SELECT` permission to necessary tables

Intent is for this role to be granted to the forthcoming orchestrator managed identity role (and potentially a future bulk match identity role).

Creating this role/group may be overkill if it's used by only one or two identities, but this approach is intended to avoid making `create-databases.bash` dependent on other resources.

Partially addresses #1339.

### DIFF
--- a/ddl/per-state.sql
+++ b/ddl/per-state.sql
@@ -3,9 +3,9 @@ BEGIN;
 -- Creates participant records tables and their access controls.
 -- Assumes 4 database roles, to be set via the psql -v option:
 --  * cluster `superuser`
---  * cluster `reader`, which gets read-only access
 --  * database `owner`, which owns the tables, sequences
 --  * database `admin`, which gets read/write access
+--  * database `reader`, which gets read-only access
 
 SET search_path=piipan,public;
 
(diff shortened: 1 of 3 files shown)


ddl/per-state.sql
```diff
@@ -1,8 +1,9 @@
 BEGIN;
 
 -- Creates participant records tables and their access controls.
--- Assumes 3 database roles, to be set via the psql -v option:
+-- Assumes 4 database roles, to be set via the psql -v option:
 --  * cluster `superuser`
+--  * cluster `reader`, which gets read-only access
```
ryanhofdotgov, line 6: super nitpick (non-blocking): It's not really a cluster-level account, as it's not a superuser. For instance, it's not going to be able to connect to `postgres` database.

ryanhofdotgov (approved):
LGTM

ddl/per-state.sql
kingcomma: Ah, excellent point. I was trying to make it clear that this role is intended to be used across all per-state databases, but instead sort of unhelpfully restated the fact that roles are global entities.

PR #1429, opened 2021-08-20
With Easy Auth at web app, drop FD firewall rules

ryanhofdotgov:
Now that Easy Auth is enabled at App Service Level, we can drop the Front Door firewall rules that restricted access to the GSA network.

- [X] Applied to tts/dev on the dashboard and query tool Front Doors

### DIFF
--- a/iac/arm-templates/front-door-app-service.json
+++ b/iac/arm-templates/front-door-app-service.json
@@ -80,47 +80,7 @@
                     "enabledState": "Enabled"
                 },
                 "customRules": {
-                    "rules": [
-                        /* Allow GSA IP's */
-                        {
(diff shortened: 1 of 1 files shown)


echappen:
@ryanhofdotgov [question] Do we want to drop the `add-front-door-to-app.bash` scripts from iac altogether, or do we anticipate having other custom firewall rules in the future?

ryanhofdotgov:
@echappen I think it is good to keep it more modular, as a separate script. We may also add back state-level, partner-level firewall rules at some point or may want to tweak the web application firewall rules beyond the default set.

echappen (approved):

PR #1430, opened 2021-08-20
Add database role for orchestrator identity

kingcomma:
The bulk of the diff for `iac/create-resources.bash` is from re-ordering existing code so the current CLI user remains a member of the Azure AD admin group until after the orchestrator's role is added to the `participant-records`.

I'm [still deferring](https://github.com/18F/piipan/pull/1100#discussion_r658080454) unifying our db commands into a set of common utilities — until then `iac/config-managed-role.bash` duplicates some code. Happy to address this now if others think it's the right time.

Partially addresses #1339

### DIFF
--- a/iac/config-managed-role.bash
+++ b/iac/config-managed-role.bash
@@ -1,0 +1,75 @@
+#!/bin/bash
+
+# shellcheck source=./tools/common.bash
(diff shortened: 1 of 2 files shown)


ryanhofdotgov (approved):
LGTM

Issue #779, opened 2021-05-04
Tidy up dashboard and only show the latest upload for each state


ryanhofdotgov:
[ ] Replace "Hello world" with "Available metrics" (or similar)
[ ] Only show the latest upload for each state

juliaklindpaintner:
Would definitely add the following edits: 
- [ ] Use sentence case for "Participant uploads" on both pages (rather than title case as it is now)
- [ ] Use sentence case for table headers (State, Uploaded at)
- [ ] Add header above table with "Most recent uploads from each state" (or similar)

Consider the following edits:
- [ ] Consider changing "participant uploads" to "State participant data uploads" (or similar; something to make it clear where the uploads are coming from — could be read now as uploads that participants are making)
- [ ] Consider changing page title to "National Accuracy Clearinghouse Metrics Dashboard" — formatted similarly to the Query and Lookup tools so there's consistency across the three (though I understand the users are different)
- [ ] Format "Available metrics" as a header

Other thoughts, lower priority:
- [ ] Add CUI banners?
- [ ] Add pagination indication (i.e. clickable page 1, 2, etc) below the table when multiple pages exist

ryanhofdotgov:
@juliaklindpaintner  Re: CUI banners on dashboard, there's no privacy information shown there and don't believe any metrics it would report would fall under any of the [CUI categories](https://www.archives.gov/cui/registry/category-list).

juliaklindpaintner:
Captured all these recommendations and a few others in the [doc of initial UX/UI suggestions](https://docs.google.com/document/d/1vdjSyvKMzKIqyZ0JM3QAlv_3X3NfLSPjGwVIi2D00y8/edit#heading=h.4j504h5gfhbh). 

Also did some rough mockups to inform these updates.
![image](https://user-images.githubusercontent.com/42982184/120845234-854d1e00-c53e-11eb-86b0-0c96a383384e.png)

juliaklindpaintner:
@echappen Skippy and I reviewed this today and it is ready to be picked up if you have extra time this sprint (it's a stretch goal).

echappen:
> Only show the latest upload for each state

@ryanhofdotgov @juliaklindpaintner If this is something we want longterm, it might be easier for me to change the endpoint of the API to accommodate this (or add a new endpoint). Mind if I break this out into another card?

juliaklindpaintner:
[...]

PR #1433, opened 2021-08-23
various metrics dashboard tweaks

echappen:
Partially addresses #779 

Everything in #779 should be addressed here, except any changes related to showing last upload for each state.

I will make another PR to add another Metrics API endpoint for last upload from each state, then come back to the dashboard and finish up those visual changes. 

### Dev notes
I added a helper method to `Piipan.Shared` to display relative time given two DateTimes. It assumes the timezones are the same when the dates are passed to this method. When I attempted to standardize dates in the method using [DateTime.ToUniversalTime](https://docs.microsoft.com/en-us/dotnet/api/system.datetime.touniversaltime?view=net-5.0), the relative date ended up being off by 4 hours. I'm wondering if it's possible for the helper method to be given two DateTimes (regardless of whether they're standardized beforehand) and standardize them in the method.


echappen:
Screenshots:

Homepage:

localhost_5001_ (3)

----

Landing for participant uploads:

localhost_5001_ParticipantUploads_

----

Search for one state:

Screen Shot 2021-08-23 at 10 10 44 AM

----

Timestamp Detail:

Screen Shot 2021-08-23 at 10 47 41 AM

juliaklindpaintner (approved):
LGTM Thank you :)

kingcomma:
@echappen I was curious about the timezone point you raised and dug into it a bit. Here's what I found in case it's helpful for any future work:

- The `uploaded_at` column in `participant_uploads` uses the `timestamp` type which [does not store timezone info](https://www.postgresql.org/docs/9.1/datatype-datetime.html#DATATYPE-DATETIME)
- The value that is inserted into that column [comes from `EventGridEvent.EventTime`](https://github.com/18F/piipan/blob/dev/metrics/src/Piipan.Metrics/Piipan.Metrics.Collect/BulkUploadMetrics.cs#L38) which appears to return the time in UTC (suggested by this [Java reference doc](https://docs.microsoft.com/en-us/java/api/com.microsoft.azure.eventgrid.models.eventgridevent.eventtime?view=azure-java-stable#com_microsoft_azure_eventgrid_models_EventGridEvent_eventTime__))
- When the dashboard deserializes the API response to `ParticipantUpload`, `ParticipantUpload.UploadedAt` is set to a `DateTime` object with its `Kind` property set to the [default value](https://docs.microsoft.com/en-us/dotnet/api/system.datetime.kind?view=net-5.0#property-value) `Unspecified` (I suspect it might be set to `UTC` if the timestamp had the timezone?)
- Finally, `DateTime.ToUniversalTime` [treats `Unspecified` as if it were `Local`](https://docs.microsoft.com/en-us/dotnet/api/system.datetime.touniversaltime?view=net-5.0#:~:text=converted%20to%20UTC.-,Unspecified,-The%20current%20DateTime) which results in the UTC time that came out of the database being treated as if it was local to the run-time environment (which would be your local time on your local machine and [UTC on Azure](https://docs.microsoft.com/en-us/answers/questions/35375/azure-local-time-is-incorrect-for-all-but-web-serv.html#:~:text=By%20default%2C%20on%20Azure%20App%20Service%2C%20the%20time%20zone%20is%20always%20UTC))

echappen:
@kingcomma Thank you for looking into this! It’s clear that this whole flow could be made more consistent and explicit in the code. An offline convo may be better, but some initial thoughts:

The most critical change here seems to be to avoid using an `Unspecified` Kind anywhere. This would mean setting the `Kind` property on `ParticipantUpload.UploadedAt` to UTC when it’s created, so that downstream conversions are unambiguous. 

A less critical but still helpful change would be to explicitly check that times saved in Postgres are in UTC. Even though they’re supposed to be UTC from `EventGridEvent.EventTime` it'd be good to double check before saving the record. 

Third change which I’m unsure of—is it worth saving timezone at the database level? 

As for the `DateFormatters.RelativeTime` helper method—is it ok to presume that DateTimes passed to this method should have a `Kind` property set? That way it could use a method like `ToUniversalTime` consistently. 

I also feel an ADR brewing about what our approaches should be on handling Dates and Times. My general approach is to keep times in UTC for as long as possible until the point where localization is needed, which in most cases happens on the frontend when displaying to a user.

Do you think any of these are worth putting into a new issue?

PR #1435, opened 2021-08-23
Add metrics API endpoint for retrieving last upload from each state

echappen:
Partially addresses #779 

This endpoint will be used as the landing page—the original endpoint will still be used for searches. 

- [ ] run iac on tts/dev after approval and prior to merge


ryanhofdotgov:
Haven't looked at this too closely, but I believe there should be one metrics API endpoint that just has different API calls. As we add new calls, we shouldn't have to alter the intrastructure or register new endpoints.

echappen:
@ryanhofdotgov I added a new function to the Metrics API function app. This created a new and separate URI, which I was calling an "endpoint". Do we instead want a single function per function app?

ryanhofdotgov:
Sorry – I wasn't writing precisely at all. Definitely **not** suggesting we that we build a new Function App for every new bit of functionality added to the Metrics API 😅 

Using the OpenAPI spec vocab, I am proposing that the Server Object url would be the value passed by the infrastructure-as-code and set in the App Service environment variable. But the one or more relative paths to individual endpoints of the server would be application-level knowledge.

This way, as more functionality get added to the Metrics API, the infrastructure shouldn't ever have to change. To continue with our analogy, we should be able to add more flesh to the bones (in app space), without changing the underlying (infra) skeleton.

If the original metrics API URL being configured in the IaC was a fully qualified URL specifying the individual path of an API call, then that should be changed to the Server Object url only.

I am hoping that once we transition the product, the Infrastructure as Code will need to be updated and re-run very infrequently, only for major new features.

echappen:
@ryanhofdotgov Thanks for explaining!

> If the original metrics API URL being configured in the IaC was a fully qualified URL specifying the individual path of an API call, then that should be changed to the Server Object url only.

This is the case—I'll make that change in this PR.

PR #1436, opened 2021-08-23
Initial PPRL write-up

### LINKED ISSUE #1184: Normalization strategy and proof of concept
As a state agency worker, I want an easy way to format the fields that will be shared to the NAC so that I know that the same name will match with another state, even if their system stores the information slightly differently.

AC:
- [ ] Document the ways in which each of the encrypted fields need to be normalized.
- [ ]

ryanhofdotgov:
Closes #1184

### DIFF
--- a/docs/pprl.md
+++ b/docs/pprl.md
@@ -1,0 +1,155 @@
+# A Privacy-Preserving Record Linkage (PPRL) approach
+
+## Overview
+
+Piipan incorporates a secure hash encoding technique to de-identify Personally Identifiable Information (PII) of SNAP participants using a centralized [Privacy-Preserving Record Linkage (PPRL)](https://link.springer.com/referenceworkentry/10.1007%2F978-3-319-63962-8_17-1) model.
+
+In brief:
+- States compute a secure hash for every SNAP participant using the participant's PII as input
+- The secure hash is submitted in the state's daily bulk upload instead of the participant's source PII
+- Piipan searches for exact matches across the secure hashes of the participating states
+- When state eligibility systems query the duplicate participation API during (re)certification, the secure hash for the SNAP applicant is provided as input, instead of their PII
+
+Only one secure hash is currently defined for Piipan and is derived from a participant's *last name*, *Date of Birth (DoB)*, and *Social Security Number (SSN)* and the application of the [SHA-512 cryptographic hash function](https://en.wikipedia.org/wiki/SHA-2). Additional secure hashes may be defined in the future to support additional matching algorithms beyond an exact match against last name, DoB, and SSN.
+
+Finally, in order for exact matching against de-identified PII to be effective, states must consistently validate and normalize the participant PII values, concatenate them a specified manner, and then apply the SHA-512 hash function. These steps are described in detail below.
+
+## 1. Validation and Normalization
+
+### Last name
+
+Participant's Last name should be normalized and validated in accordance with these **ordered** rules. These transformations assume ASCII-encoded input.
+
+1. Convert to lower case
+1. Remove any suffixes (e.g.; `junior`, `jnr`, `jr`, `jr.`, `iii`, etc.)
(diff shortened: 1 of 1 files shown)


ryanhofdotgov:
Tagging everyone as this will be referenced from our updated API specs.

echappen:
What do you think of a table for the name rules? (a la the Plain Language [guide](https://content-guide.18f.gov/our-approach/structure-the-content/))

| Rule | Incorrect | Correct |
| --- | --- | --- | 
| Convert to lowercase | Thatcher | thatcher |
| Remove any character that is not an ASCII space (0x20) or in the range [a-z] (0x61-0x70) | García | garcia | 
| Remove any suffixes (e.g.; junior, jnr, jr, jr., iii, etc.) | Jones III | jones |
| Replace hyphens with a space | Barrable-Tishauer | barrable tishauer |
| Replace multiple spaces with one space | Heathcote  Drummond | heathcote drummond |
| Trim any spaces at the start and end of the last name | &nbsp;&nbsp;&nbsp;Amit | amit |
| Remove apostrophes | O'Sullivan | osullivan |
| Validate that name has at least one ASCII character | 이 | i |

docs/pprl.md
```diff
@@ -0,0 +1,155 @@
...
+
+1. Convert to lower case
+1. Remove any suffixes (e.g.; `junior`, `jnr`, `jr`, `jr.`, `iii`, etc.)
+1. Replace hyphens with a space
+1. Remove any character that is not an ASCII space (`0x20`) or in the range `[a-z]` (`0x61`-`0x70`)
+1. Replace multiple spaces with one space
+1. Trim any spaces at the start and end of the last name
+1. Validate that the resulting value is at least one ASCII character in length
+
+If your source data set includes non-ASCII characters using the ISO-8859-1 (Latin-1) or Unicode encoding formats, [an ASCII normalization process to remove diacritics and derive base characters](https://ahinea.com/en/tech/accented-translate.html) should be applied before these rules.
```
kingcomma, line 31: *Question:* I haven't had to do much ASCII normalization work — will we eventually need a clear set of criteria for transformations? E.g., `í` to `i`, `ä` to `ae`, etc.

ryanhofdotgov:
@echappen Good suggestion to incorporate tables. But there's not a 1:1 correspondence of rules with the incorrect examples I think are useful to show. I think I just have one that violates that (`heathcote drummond-willoughby`), but there may be others.

How about I add 2 tables for each of the 3 fields, for correct and incorrect, but leave the ordered rules (and other validation steps) outside a table?

docs/pprl.md
ryanhofdotgov: I left that a little less defined (i.e., go do what this web page suggests) as I am skeptical that any state/federal system that does this type of work supports names with diacritics. Check out [this article for California from a few years ago](https://www.theguardian.com/us-news/2015/apr/11/california-birth-certifcates-accents-marks) and [this one for Texas](https://www.kut.org/texas/2017-03-22/this-lawmaker-says-its-time-to-allow-accent-marks-on-state-documents).

I figure if this actually becomes a problem, a future development team will have the pleasure of formalizing all the steps on that referenced page 🙂

kingcomma: 😇

juliaklindpaintner:
@ryanhofdotgov Looking at the rendered view, I had a question: Are we intentionally not specifying whether they should normalize the first or middle name as well? I realize that those don't feed into the matching algorithm, but just curious what the expectation would be?

ryanhofdotgov:
@juliaklindpaintner We won't be collecting first name or middle name (as plain text PII or in de-identified form), so we get to skip specifying those.

echappen:
> How about I add 2 tables for each of the 3 fields, for correct and incorrect, but leave the ordered rules (and other validation steps) outside a table?

Sounds good!

PR #1441, opened 2021-08-24
Swap SQL for per-state APIs in Orchestrator

kingcomma:
- Add IaC for binding orchestrator to database
- Update application to use SQL when lookig for matches
- Small refactoring to group together error/exception exception handling
- Update unit tests and add integration tests
- Update documentation

Apologies for the large commit — new functionality is primarily in commit ccad31c.

Follow-on work will drop the per-state codebase and IaC. I didn't want to complicate this PR even further.

Partially addresses #1339.


kingcomma:
@echappen I figure you're best suited for the code-level changes since you were the last one to touch this sub-system.

@ryanhofdotgov tagging you as well in case you have any general (or other) input as the one who created this ticket/plan.

match/src/Piipan.Match.Orchestrator/Api.cs
```diff
@@ -45,110 +53,84 @@ public Api(IAuthorizedApiClient apiClient)
...
+            {
+                return DeserializationErrorResponse(ex);
+            }
+            catch (System.FormatException ex)
+            {
+                return DeserializationErrorResponse(ex);
+            }
+            catch (Exception ex)
+            {
+                return InternalServerErrorResponse(ex);
```
echappen, line 117: Nice! This exception handling is a lot cleaner.

match/src/Piipan.Match.Orchestrator/Api.cs
```diff
@@ -159,56 +141,140 @@ private OrchMatchRequest Parse(string requestBody, ILogger log)
...
+            {
+                personResult.Matches.AddRange(stateResult.Result);
+            }
 
-            return matchResponse;
+            return personResult;
         }
 
-        private async Task<List<PiiRecord>> PersonMatch(PersonMatchRequest request, ILogger log)
+        private async Task<string> ConnectionString(string database)
```
echappen, line 198: [non-blocking] It took me a minute to confirm the name of the database is the same as the state abbreviation. I wonder how this can be more explicit here? Someone familiar with the db schema would remember this, but it seems implied here.

match/tests/Piipan.Match.Orchestrator.IntegrationTests/DbFixture.cs
```diff
@@ -0,0 +1,199 @@
...
+
+                using (var cmd = Factory.CreateCommand())
+                {
+                    cmd.Connection = conn;
+
+                    // cmd.CommandText = "DROP TABLE IF EXISTS participants;";
+                    // cmd.ExecuteNonQuery();
+
+                    // cmd.CommandText = "DROP TABLE IF EXISTS uploads;";
+                    // cmd.ExecuteNonQuery();
```
echappen, line 75: [question] Did you still want these commented out?

match/docs/orchestrator-match.md
```diff
@@ -14,40 +14,47 @@ The orchestrator matching API is implemented in the `Piipan.Match.Orchestrator`
...
+| `DatabaseConnectionString` | [details](../../docs/iac.md#\:\~\:text=DatabaseConnectionString) — Additionally, `Database` is set to the placeholder value `{database}`. The relevant per-state database name is inserted at run-time as needed. |
+| `States` | [details](../../docs/iac.md#\:\~\:text=States) |
 
-## Binding to state APIs
+## Local development
 
-The orchestrator treats per-state APIs as backing services. When running the [IaC](../../docs/iac.md):
-- Per-state URIs are compiled into a JSON list and saved as an environment variable
-- The orchestrator's system-assigned identity is given an authorized application role (which will be checked by the state API upon receiving requests)
+Local development is currently limited as a result of using a managed identity to connect to the state databases. The Instance Metadata Service used by managed identities to retrieve authentication tokens is not available locally. There are [potential solutions](https://docs.microsoft.com/en-us/dotnet/api/overview/azure/service-to-service-authentication#local-development-authentication) using the `Microsoft.Azure.Services.AppAuthentication` library. None have been implemented at this time.
```
echappen, line 30: [non-blocking question] Do you think we'd be able to implement this solution before vendors come on?

match/tests/docker-compose.yml
```diff
@@ -10,6 +10,7 @@ services:
     environment:
       DatabaseConnectionString: "Server=statedb;Database=ea;Port=5432;User Id=postgres;Password=securepass;"
       StateAbbr: ea
+      States: ea
```
echappen, line 13: [non-blocking question] Probably best done in another PR, but do you think it's worth using multiple state databases for our integration tests?

echappen (approved):
This was a lot of work, Great job 🥇

match/tests/Piipan.Match.Orchestrator.IntegrationTests/DbFixture.cs
kingcomma: Good catch — I did not! I was debugging the integration tests and wanted to leave some breadcrumbs.

match/docs/orchestrator-match.md
kingcomma: Yeah, I think it's necessary that we're able to get this working before then one way or another. The groundwork is mostly there — Docker, PostgreSQL database — and the main remaining task is to move the process for creating the database(s) and applying our schema out of `DbFixture.cs` and into scripts that can be run in any environment.

match/tests/docker-compose.yml
kingcomma: Absolutely — I'll open a ticket for this and the local development work mentioned above, which go hand in hand.

match/src/Piipan.Match.Orchestrator/Api.cs
kingcomma: Good point that I hadn't thought of! My thinking was that keeping this method agnostic of the implementation specifics of the database(s) made it more flexible to future changes (e.g., the unlikely scenario that databases are renamed to something other than just the two-letter state abbreviation).

Do you think this current implementation is confusing enough to merit an edit? Or, maybe this confusion goes away if/when we address #192?

echappen: Yeah, it may not be worth addressing until #192 is done. I'd say it's ok for now.

PR #1448, opened 2021-08-24
Initial release instructions

ryanhofdotgov:
This is the process I've been following so far, but feel free to correct/adjust/suggest. Thanks!

### DIFF
--- a/docs/releases.md
+++ b/docs/releases.md
@@ -15,7 +15,7 @@ Note: before tagging the release, it is easiest to update the CHANGELOG first.
 Tagging process:
 
 ```
-git checkout develop
+git checkout dev
 git checkout -b release/<version-number>    # e.g., release/0.5
 git tag v<version-number>                   # e.g., v0.5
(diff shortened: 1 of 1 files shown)


docs/releases.md
```diff
@@ -0,0 +1,39 @@
...
+The convention for each release tag is `v<version-number>`; e.g., `v0.1`, `v0.2`.
+
+## Tagging the release
+
+Note: before tagging the release, it is easiest to update the CHANGELOG first.
+
+Tagging process:
+
+```
+git checkout develop
```
kingcomma, line 18:
```suggestion
git checkout dev
```

kingcomma (approved):
Just one little tweak, but LGTM 👍 

Thanks for taking the time to type this up.

docs/releases.md
ryanhofdotgov: Good catch – stole most of this from one of my previous engagements

PR #1480, opened 2021-08-27
Removed per-state API, 1...N from bulk match

ryanhofdotgov:
[Corresponding Google Drawing](https://docs.google.com/drawings/d/1iSC5M6pxcnfV_LoqTnSGgZ9U4TZTbfPf4CpHCv001ug). To reproduce, export drawing to PDF, open in Preview, export as PNG at 150dpi.


kingcomma (approved):
👍 Thanks for handling!

PR #1489, opened 2021-08-30
Add integration test for saving CSV data on bulk upload

### LINKED ISSUE #1487: Add initial database integration tests for Etl subsystem
Non-blocking, but it might be a good time to incorporate a basic database integration test like @kingcomma did a while back for the match subsystem: [Piipan.Match.Orchestrator.IntegrationTests](https://github.com/18F/piipan/tree/dev/match/tests/Piipan.Match.Orchestrator.IntegrationTests)

_Originally posted by @ryanhofdotgov in https://github.com/18F/piipan/pull/1486#issuecomment-908531012_

echappen:
Closes #1487 

Also adds the test suite to CI pipeline and sln. 

Wondering if it'd be good to stream `example.csv` in this test.


etl/tests/Piipan.Etl.IntegrationTests/DbFixture.cs
```diff
@@ -0,0 +1,182 @@
...
+
+                    cmd.CommandText = "DROP TABLE IF EXISTS uploads;";
+                    cmd.ExecuteNonQuery();
+                }
+
+                conn.Close();
+            }
+        }
+
+        // xxx Apply schema directly from ddl/per-state.sql
```
ryanhofdotgov, line 83: The test fixture really needs to use the DDL we use to create the database in production. Part of the motivation for prioritizing integration testing now, is that it would have automatically flagged the DDL errors in the recent ETL PR.

When I'm making DDL changes, I will have a local (e.g., Dockerized) instance of PostgreSQL to interactively test my changes. We've sort of deferred formalizing this (along with schema migration), but basic steps look like this:
```
$ docker pull postgres:11-alpine
$ docker start <container-id>
$ docker cp ddl/per-state.sql <container-id>:/tmp
$ docker exec -it <container-id> /bin/bash
# su - postgres 
$ createdb foobar
$ createuser fooowner
$ createuser fooadmin
$ createuser fooreader
$ createuser foosuper
$ psql -d foobar -v owner=fooowner -v admin=fooadmin -v reader=fooreader -v superuser=foosuper -f /tmp/per-state.sql
```

We need to externalize the metrics schema the same way that the ETL schema is externalized in a file. And I see now that the match subsystem integration test also took this shortcut, and we will need to revisit that next sprint.

etl/tests/Piipan.Etl.IntegrationTests/DbFixture.cs
```diff
@@ -0,0 +1,164 @@
...
```
ryanhofdotgov, line 83:
```suggestion
```

ryanhofdotgov (approved):
Ready for merge 👍 

I've opened up some related tickets for follow-up at some point in the future:
- Use Dapper with ETL Function App and tests (will make both functional and test code cleaner)
- Create shared library functions for integration test harness
- Share PostgreSQL docker file across subsystems for integration tests

PR #1491, opened 2021-08-31
Update API specs to incorporate PPRL approach

### LINKED ISSUE #1452: Update OpenAPI spec to accept de-identified PII
Update spec should reflect:
- [ ] Internal systems are still able to query by name, DOB, SSN (may require wrangling to existing doc generation process)
- [ ] API accepts de-identified hash in place of name, DOB, SSN

Split off from original issue #1215.

kingcomma:
- Update both internal-facing and state-facing specs, keeping internal-facing PII-based endpoint
- Bump all endpoints from `v1` to `v2`
- Rename endpoints from `/query` to `/find_participants`
- Remove PII fields (name, dob, ssn) from response

As usual, it may be easiest to review the state-facing changes by looking at `docs/openapi/generated/duplicate-participation-api/openapi.md`. Internal-facing changes might be easiest to review by pulling down this branch, running `match/tools/generate-specs.bash` and reviewing at the output in `match/docs/openapi/orchestrator/generated`.

@cjhskippy @ryanhofdotgov Should the internal-facing PII endpoint still require match requests include first name?

@ryanhofdotgov @echappen @timoballard tagging all of you since I'm using this change as an opportunity to rename the endpoint to be less ambiguous. I'm opening the bidding with `find_participants` (in the spirit of [Google's RPC naming conventions](https://cloud.google.com/apis/design/naming_convention#method_names)), but open to all suggestions.

Closes #1452

### DIFF
--- a/match/docs/openapi/schemas/match-query.yaml
+++ b/match/docs/openapi/schemas/match-query.yaml
@@ -1,4 +1,3 @@
-#/components/requestBodies/MatchRequest
 MatchRequest:
   required: true
   content:
@@ -15,8 +14,35 @@ MatchRequest:
             items:
               $ref: '#/Person'
       examples:
-        $ref: '#/MatchRequestExamples'
+        Single person:
+          $ref: '#/MatchRequestExamples/Single'
+        Multiple persons:
+          $ref: '#/MatchRequestExamples/Multiple'
+MatchRequestWithPii:
+  required: true
+  content:
+    application/json:
+      schema:
+        type: object
+        required:
+            - data
+        properties:
+          data:
+            type: array
+            minItems: 1
+            maxItems: 50
+            items:
+              $ref: '#/PersonWithPii'
 Person:
+  type: object
+  required:
+    - lds_hash
+  properties:
+    lds_hash:
+      type: string
+      description: "SHA-512 digest of participant's last name, DoB, and SSN. See docs/pprl.md for details"
+      pattern: "^[0-9a-f]{128}$"
+PersonWithPii:
   type: object
   required:
     - first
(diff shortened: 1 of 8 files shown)


match/docs/openapi/schemas/match-query.yaml
```diff
@@ -26,54 +52,37 @@ Person:
   properties:
     first:
       type: string
-      description: "PII record's first name"
+      description: "Person's first name"
```
echappen, line 55: [question] What's now the purpose of passing `first` and `middle` name in the request?

match/docs/openapi/schemas/match-query.yaml
```diff
@@ -26,54 +52,37 @@ Person:
...
     middle:
       type: string
-      description: "PII record's middle name"
+      description: "Person's middle name"
     last:
       type: string
-      description: "PII record's last name"
+      description: "Person's last name"
```
echappen, line 61: [question] Do we need to specify here that `last`, `dob`, and `ssn` are required in the request?

echappen: Scratch that—I always forget required params specified at top-level. My next question though would be why `first` is required?

kingcomma: Agreed this is an open question (I directed it to @ryanhofdotgov / @cjhskippy in the first comment on the PR). At this point the requirement is a holdover from the previous approach to querying for matches where the product owner wanted the data included in queries even if it wasn't used as part of the current matching algorithm.

The question at this point is: Does this requirement still apply in the context of internal tools (i.e., the Query Tool) which are still able to make PII-based queries?

kingcomma: Perhaps a more straightforward way to ask the question: The query tool still requires first name and accepts middle name as an optional field — do we want the API that backs the query tool to enforce the same schema?

ryanhofdotgov:
> Should the internal-facing PII endpoint still require match requests include first name?

@kingcomma A "soft" yes, as it will allow us to leave that API unchanged if the system ever adds a secondary hash that incorporates first name (e.g., to address maiden name -> married name false negative).

But it's an internal API so it's easily changed with limited coordination.

kingcomma:
Renamed endpoints to `/find_matches`. Relevant [slack conversation](https://gsa-tts.slack.com/archives/C018P1ZJPEC/p1630429401052900).

ryanhofdotgov (approved):
Ready for merge 👍

PR #1486, opened 2021-08-30
Accept and store de-identified hashes in Bulk Upload process

echappen:
Closes #1214 

- [ ] run iac on `tts/dev` prior to merge

### Removes
- Removes `last`, `first`, `dob`, `ssn` columns from bulk upload csv

### Adds
- Add  `lds_hash` column to bulk upload csv

### Changes
- Makes `participant_id` column required on bulk upload csv

Issue #1215, opened 2021-07-27
Use deidentified PII in query API


cjhskippy:
As a state agency developer, I want to be able to query for participants using the deidentified hash instead of submitting PII directly so that I can protect the privacy of SNAP participants in my state.

AC:
Query API...
- [ ] ~~No longer allows states to provide name, DOB, SSN~~
- [ ] Allows internal methods to continue to query by name, DOB, SSN (does not need to work for the purposes of this story) (see #1540)
- [x] Accepts deidentified hash
- [x] Requires hash to be formatted properly
- [x] Has updated documentation (see #1452)
- [ ] Includes updated test data sets for users to query against and verify successful use (see #1213 )

PR #1500, opened 2021-09-01
Refactor match API in prep for PPRL approach

kingcomma:
- Pull matching logic out to a method (`FindMatches`) that will be shared between the de-identified data endpoint and the PII endpoint
- Rename `PiiRecord` --> `ParticipantRecord`

Just a small refactor setting the stage for #1215. Note: since the database was changed with #1486 the remote match API is temporarily broken and will not return match results until it is updated to reflect the new schema for participant records.


ryanhofdotgov (approved):
Ready to merge 👍

As follow-up, should make a similar class name change in the ETL? from PiiRecord to ParticipantRecord? Or otherwise refactor to reuse code?

kingcomma:
I think just a name change (to reflect our new approach) makes sense for now. It would be great to figure out reuse, but right now our various `PiiRecord`s/`ParticipantRecord`s are a mix of DTOs, serializers, mappers, etc. A clean solution to reuse may take some work.

PR #1503, opened 2021-09-01
Bump version for bulk upload api to v2

### LINKED ISSUE #1488: Bump version number of bulk upload
For the API revisions where we are dropping plain text PII and replacing with hashes, we should bump the major version of the publicly exposed API, including the bulk upload API, as it is a breaking change.

(We probably should have started with v0 to reflect that everything was still in flux, but since we started with v1, seems worth a bump)

echappen:
Closes #1488 

@kingcomma Wanted to make sure I'm covering all that's needed for APIM to update. Would I just need to re-run IaC with the new version number?


kingcomma:
This should be verified, but my understanding of the change you have in `apim.json` is it will update the existing "v1" API resource in place. I.e., each of the existing per-state `Microsoft.ApiManagement/service/apis` resources will be modified, changing the "Version identifier" from "v1" to "v2" and thus the URL from `/bulk/{state}/v1` to `/bulk/{state}/v2`. No new resources will be created.

The critical factor is whether or not the existing subscriptions (API keys) will still be scoped to the updated API resources. If they aren't, we'll have to go through the whole process of creating new subscriptions and delivering them to states. If what I described above is true, I think the subscriptions will still be scoped correctly. To verify this, I'd recommend:
1. Create a temporary subscription on the `tts/dev` APIM resource that is scoped to one of the per-state bulk upload "v1" APIs
2. Run the IaC
3. Confirm the temporary subscription is scoped to the updated API (now "v2")

Happy to chat one on one about this too.

ryanhofdotgov (approved):
Ready for merge 👍

PR #1511, opened 2021-09-01
Update match API to use de-identified data

kingcomma:
- Replace PII properties with hash property
- Update SQL query/matching algorithm
- Align integration tests database fixture with recent work in `Piipan.Etl`
- Use Dapper in database fixture

Partially addresses #1215.

Note: the PII-based endpoint `/find_matches_by_pii` is stubbed in and currently returns just a 204 response.

### DIFF
--- a/match/tests/Piipan.Match.Orchestrator.IntegrationTests/ApiIntegrationTests.cs
+++ b/match/tests/Piipan.Match.Orchestrator.IntegrationTests/ApiIntegrationTests.cs
@@ -19,6 +19,7 @@ namespace Piipan.Match.Orchestrator.IntegrationTests
         {
             return new ParticipantRecord
             {
+                // farrington,1931-10-13,000-12-3456
                 LdsHash = "eaa834c957213fbf958a5965c46fa50939299165803cd8043e7b1b0ec07882dbd5921bce7a5fb45510670b46c1bf8591bf2f3d28d329e9207b7b6d6abaca5458",
                 CaseId = "CaseIdExample",
                 ParticipantId = "ParticipantIdExample",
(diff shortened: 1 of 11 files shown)


echappen (approved):

match/tests/Piipan.Match.Orchestrator.IntegrationTests/ApiIntegrationTests.cs
```diff
@@ -19,11 +19,7 @@ static ParticipantRecord FullRecord()
         {
             return new ParticipantRecord
             {
-                First = "First",
-                Middle = "Middle",
-                Last = "Last",
-                Dob = new DateTime(1970, 1, 1),
-                Ssn = "000-00-0000",
+                LdsHash = "eaa834c957213fbf958a5965c46fa50939299165803cd8043e7b1b0ec07882dbd5921bce7a5fb45510670b46c1bf8591bf2f3d28d329e9207b7b6d6abaca5458",
```
ryanhofdotgov, line 22: Non-blocking, but I think it would be a good practice to start including the input to the hash in an adjacent comment whenever we initialize a hash variable like this.

match/tests/Piipan.Match.Orchestrator.IntegrationTests/DbFixture.cs
```diff
@@ -83,46 +72,17 @@ public void Dispose()
         // xxx Apply schema directly from ddl/per-state.sql
         private void ApplySchema()
         {
+            string sqltext = System.IO.File.ReadAllText("per-state.sql", System.Text.Encoding.UTF8);
```
ryanhofdotgov, line 75: Non-blocking, but there's an obsolete xxx comment a few lines above this

ryanhofdotgov (approved):

PR #1538, opened 2021-09-02
Require participant_id in match responses in docs

echappen:
@kingcomma Just noticed this in the openapi generated docs—I think this field will always be returned since it's required now.


kingcomma (approved):
Great catch! That invalid example snuck right past me.

Issue #1553, opened 2021-09-07
Orchestrator API knows user's State


cjhskippy:
The ...

kingcomma:
Split off the tasks specific to the Query Tool into #1695.

PR #1569, opened 2021-09-08
Derive state name from APIM subscription and send to DupPart API

kingcomma:
~WIP~

### DIFF
--- a/iac/apim-duppart-policy.xml
+++ b/iac/apim-duppart-policy.xml
@@ -13,31 +13,28 @@
         <!--
             Derive the state abbreviation from the subscription name and forward to
             the internal endpoint. Subscriptions names are expected to be in the
-            format: {state-abbr}-DupPart
+            format: {state-abbr}-DupPart.
+
+            Default to "ea".
         -->
-        <set-variable name="InitiatingState" value="@{
-            string pattern = @"^(?<state>[a-z]{2})-";
-            var matches = Regex.Match(context.Subscription.Name.ToLower(), pattern);
-            return matches.Groups["state"].Value;
-        }" />
+        <choose>
+            <when condition="@(Regex.IsMatch(context.Subscription.Name, "^[A-Z]{2}-DupPart"))">
+                <set-variable name="InitiatingState" value="@{
+                    string pattern = @"^(?<state>[A-Z]{2})-DupPart";
+                    var matches = Regex.Match(context.Subscription.Name, pattern);
+                    return matches.Groups["state"].Value.ToLower();
+                }" />
+            </when>
+            <otherwise>
+                <set-variable name="IntitiatingState" value="ea" />
+            </otherwise>
+        </choose>
         <!--
-            Inject derived state abbreviation into request body's `meta` object,
-            creating `meta` if it does not exist.
+            Pass derived state abbreviation to backend as X-Initiating-State header
         -->
-        <set-body>@{
-            const string metaKey = "meta";
-            JObject inBody = context.Request.Body.As<JObject>();
-
-            if (!inBody.ContainsKey(metaKey))
-            {
-                inBody[metaKey] = new JObject();
-            }
-
-            JToken meta = inBody.GetValue(metaKey);
-            meta["initiating_state"] = context.Variables.GetValueOrDefault<string>("InitiatingState");
-
-            return inBody.ToString();
-        }</set-body>
+        <set-header name="X-Initiating-State" exists-action="override">
+            <value>@(context.Variables.GetValueOrDefault<string>("InitiatingState", ""))</value>
+        </set-header>
     </inbound>
     <backend>
         <base />
--- a/iac/apim-duppart-policy.xml
+++ b/iac/apim-duppart-policy.xml
@@ -10,6 +10,34 @@
         <set-header name="Ocp-Apim-Subscription-Name" exists-action="override">
             <value>@(context.Subscription.Name)</value>
         </set-header>
+        <!--
+            Derive the state abbreviation from the subscription name and forward to
+            the internal endpoint. Subscriptions names are expected to be in the
+            format: {state-abbr}-DupPart
+        -->
+        <set-variable name="InitiatingState" value="@{
+            string pattern = @"^(?<state>[a-z]{2})-";
+            var matches = Regex.Match(context.Subscription.Name.ToLower(), pattern);
+            return matches.Groups["state"].Value;
+        }" />
+        <!--
+            Inject derived state abbreviation into request body's `meta` object,
+            creating `meta` if it does not exist.
+        -->
+        <set-body>@{
+            const string metaKey = "meta";
+            JObject inBody = context.Request.Body.As<JObject>();
+
+            if (!inBody.ContainsKey(metaKey))
+            {
+                inBody[metaKey] = new JObject();
+            }
+
+            JToken meta = inBody.GetValue(metaKey);
+            meta["initiating_state"] = context.Variables.GetValueOrDefault<string>("InitiatingState");
+
+            return inBody.ToString();
+        }</set-body>
     </inbound>
     <backend>
         <base />
--- a/iac/apim-duppart-policy.xml
+++ b/iac/apim-duppart-policy.xml
@@ -30,15 +30,9 @@
             </when>
             <otherwise>
                 <!-- Subscription name does not match the prescribed format -->
+                <!-- TODO: send log to event hub detailing the error -->
                 <return-response>
-                    <set-status code="401" reason="Unauthorized"/>
-                    <set-header name="Content-Type" exists-action="override">
-                        <value>application/json</value>
-                    </set-header>
-                    <set-body>{
-                         "statusCode": 401,
-                         "message": "A state role could not be determined for this subscription key."
-                    }</set-body>
+                    <set-status code="500" />
                 </return-response>
             </otherwise>
         </choose>
--- a/iac/apim-duppart-policy.xml
+++ b/iac/apim-duppart-policy.xml
@@ -11,11 +11,11 @@
             <value>@(context.Subscription.Name)</value>
         </set-header>
         <!--
-            Derive the state abbreviation from the subscription name and forward to
-            the internal endpoint. Subscriptions names are expected to be in the
-            format: {state-abbr}-DupPart.
+            Derive the state abbreviation from the subscription name and pass it on
+            to the backend as a header.
 
-            Default to "ea".
+            Fail if subscription name does not match the expected format of:
+            {uppercase-state-abbr}-DupPart (e.g., EA-DupPart).
         -->
         <choose>
             <when condition="@(Regex.IsMatch(context.Subscription.Name, "^[A-Z]{2}-DupPart"))">
@@ -24,17 +24,24 @@
                     var matches = Regex.Match(context.Subscription.Name, pattern);
                     return matches.Groups["state"].Value.ToLower();
                 }" />
+                <set-header name="X-Initiating-State" exists-action="override">
+                    <value>@((string)context.Variables["InitiatingState"])</value>
+                </set-header>
             </when>
             <otherwise>
-                <set-variable name="IntitiatingState" value="ea" />
+                <!-- Subscription name does not match the prescribed format -->
+                <return-response>
+                    <set-status code="401" reason="Unauthorized"/>
+                    <set-header name="Content-Type" exists-action="override">
+                        <value>application/json</value>
+                    </set-header>
+                    <set-body>{
+                         "statusCode": 401,
+                         "message": "A state role could not be determined for this subscription key."
+                    }</set-body>
+                </return-response>
             </otherwise>
         </choose>
-        <!--
-            Pass derived state abbreviation to backend as X-Initiating-State header
-        -->
-        <set-header name="X-Initiating-State" exists-action="override">
-            <value>@(context.Variables.GetValueOrDefault<string>("InitiatingState", ""))</value>
-        </set-header>
     </inbound>
     <backend>
         <base />
--- a/match/docs/duplicate-participation-api.md
+++ b/match/docs/duplicate-participation-api.md
@@ -44,6 +44,23 @@ The connection between an operation and a backend server is made by specifying t
 
 ## Managing API keys
 
-API keys are managed as [subscriptions](https://docs.microsoft.com/en-us/azure/api-management/api-management-subscriptions) in APIM. Each subscription is granted a primary and secondary API key to support credential rotation without downtime. Subscriptions can be scoped to a single API, a group of APIs (called a [product](https://docs.microsoft.com/en-us/azure/api-management/api-management-key-concepts#--products)), or all APIs.
+API keys are managed as [subscriptions](https://docs.microsoft.com/en-us/azure/api-management/api-management-subscriptions) in APIM. Currently, keys are created and managed ad-hoc by system developers via the Portal (Portal > {APIM instance} > Subscriptions).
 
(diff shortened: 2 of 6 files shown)


kingcomma:
@ryanhofdotgov Wanted to get a gut check on this proof-of-concept for addressing the APIM side of #1553.

The approach consists of an APIM policy that pulls the state abbreviation out of the subscription name and forwards the value on to the orchestrator as a property of the request body's `meta` object. The orchestrator would have to handle cases where the value is not set — presumably not creating a match record. The Query Tool would be updated to send this value as part of the request (which goes directly to the orchestrator, not through APIM).

I'm anticipating the `meta` property is where other values that have an effect on match record creation would go. E.g., a `inquiry_type` property that could accommodate the different reasons why a match request might be made (`investigative`, `certification`, etc).

iac/apim-duppart-policy.xml
```diff
@@ -10,6 +10,34 @@
...
+            const string metaKey = "meta";
+            JObject inBody = context.Request.Body.As<JObject>();
+
+            if (!inBody.ContainsKey(metaKey))
+            {
+                inBody[metaKey] = new JObject();
+            }
+
+            JToken meta = inBody.GetValue(metaKey);
+            meta["initiating_state"] = context.Variables.GetValueOrDefault<string>("InitiatingState");
```
ryanhofdotgov, line 37: Is this missing a default value? Just quickly glanced at [function signature in this doc](https://docs.microsoft.com/en-us/azure/api-management/api-management-policy-expressions)

iac/apim-duppart-policy.xml
```diff
@@ -10,6 +10,34 @@
...
         </set-header>
+        <!--
+            Derive the state abbreviation from the subscription name and forward to
+            the internal endpoint. Subscriptions names are expected to be in the
+            format: {state-abbr}-DupPart
+        -->
+        <set-variable name="InitiatingState" value="@{
+            string pattern = @"^(?<state>[a-z]{2})-";
+            var matches = Regex.Match(context.Subscription.Name.ToLower(), pattern);
+            return matches.Groups["state"].Value;
```
ryanhofdotgov, line 21: What happens if there is no matching group (e.g., the subscription name wasn't set correctly)? Drilling down the docs, I think it will throw a [KeyNotFoundException](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.ireadonlydictionary-2.item?view=net-5.0#System_Collections_Generic_IReadOnlyDictionary_2_Item__0_).

How does the error manifest itself (if at all) to the caller of the API? Have you seen how the error gets logged by APIM?

The approach is a bit brittle (though alternatives aren't great either) and just trying to walk through how a dev would find out how this went wrong.

ryanhofdotgov:
@kingcomma I think this is a good tradeoff and the `meta` element @echappen added to our API conventions seems like a great fit for this purpose. I'd want to make sure that if the policy can't determine an initiating state, then the call fails at APIM level. We will have to consider how we name our own developer subscriptions keys that we use for testing.

I wished we could annotate a subscription key with a custom key/value pair to use here. I looked around for that; closest thing is a free-form comment field that's really for a different purpose (`State comment`). I also wish there was an easy way to unit test the policy code.

However, if we don't take this convention over configuration approach, then we are stuck maintaining some type of mapping table somewhere that APIM can access quickly (e.g., compiled into policy) and that's not great.

I think as a follow-up, we would want to have a admin tool that creates these subscription keys to automate the naming convention part.

iac/apim-duppart-policy.xml
```diff
@@ -10,6 +10,34 @@
...
+        }" />
+        <!--
+            Inject derived state abbreviation into request body's `meta` object,
+            creating `meta` if it does not exist.
+        -->
+        <set-body>@{
+            const string metaKey = "meta";
+            JObject inBody = context.Request.Body.As<JObject>();
+
+            if (!inBody.ContainsKey(metaKey))
```
ryanhofdotgov, line 31: If the request body already has meta.initiating_state, we need to delete it first.

iac/apim-duppart-policy.xml
```diff
@@ -10,6 +10,34 @@
         <set-header name="Ocp-Apim-Subscription-Name" exists-action="override">
             <value>@(context.Subscription.Name)</value>
         </set-header>
+        <!--
+            Derive the state abbreviation from the subscription name and forward to
+            the internal endpoint. Subscriptions names are expected to be in the
+            format: {state-abbr}-DupPart
+        -->
+        <set-variable name="InitiatingState" value="@{
+            string pattern = @"^(?<state>[a-z]{2})-";
```
ryanhofdotgov, line 19: Thinking of ways to make this approach less brittle...

A small thing, but I'd tighten this regex even further to look for `BulkUpload` or `DupPart`. That way there is an explicit breadcrumb in the policy code for someone to find, if they alter the naming convention in the subscription creation tool.

Also would want to double-check what `Name` is. In the Azure portal UI they use `Display Name` (e.g., `Michael King`) and `Name` (e.g., `admin-king`) but in the PowerShell API, it is `Name` (e.g., `Michael King`) and  `SubscriptionId` (e.g., `admin-king`).

When we set up the first round states in APIM, both the display name and name in Azure Portal are identical. Capital letters for state abbreviation as well. For example: `MT-DupPart` and `MT-BulkUpload`.

match/docs/duplicate-participation-api.md
```diff
@@ -44,6 +44,23 @@ The connection between an operation and a backend server is made by specifying t
 
 ## Managing API keys
 
-API keys are managed as [subscriptions](https://docs.microsoft.com/en-us/azure/api-management/api-management-subscriptions) in APIM. Each subscription is granted a primary and secondary API key to support credential rotation without downtime. Subscriptions can be scoped to a single API, a group of APIs (called a [product](https://docs.microsoft.com/en-us/azure/api-management/api-management-key-concepts#--products)), or all APIs.
-
-Currently, keys are created and managed ad-hoc by system developers via the Portal (Portal > {APIM instance} > Subscriptions).
+API keys are managed as [subscriptions](https://docs.microsoft.com/en-us/azure/api-management/api-management-subscriptions) in APIM. Currently, keys are created and managed ad-hoc by system developers via the Portal (Portal > {APIM instance} > Subscriptions).
+
+When creating subscriptions during the onboarding process for states use the following process:
```
ryanhofdotgov, line 49: Also need to think about how we document this internal adaptation of the public API; perhaps a table on orchestrator-match.md?

ryanhofdotgov:
@kingcomma Giving this PR quite a bit more thought than most PRs as I think the value of initiating_state might result in access control decisions later and the low testability of APIM policy.

I also think the Orchestrator should be modified to reject any request that does not have match.initiating_state set or if it is set to a value that is not expected. And the query tool will need to be modified to map users to states; perhaps a basic domain-based approach will work for a long while.

ryanhofdotgov:
> However, if we don't take this convention over configuration approach, then we are stuck maintaining some type of mapping table somewhere that APIM can access quickly (e.g., compiled into policy) and that's not great.

BTW, I just discovered the named values feature of APIM. I briefly explored doing something like this:
Screen Shot 2021-09-10 at 4 55 43 PM

...but then discovered [you can't reference the named value in a variable](https://stackoverflow.com/a/58590560):

kingcomma:
Really appreciate the thoroughness, @ryanhofdotgov! I should have provided more context on my thinking/research rather than just dropping this in your lap.

Overall, I agree this feels like a brittle approach, especially absent any means of unit testing the policy. I'll respond to the line-by-line comments, but have a few additional thoughts here.

> I'd want to make sure that if the policy can't determine an initiating state, then the call fails at APIM level. We will have to consider how we name our own developer subscriptions keys that we use for testing.

> I also think the Orchestrator should be modified to reject any request that does not have match.initiating_state set or if it is set to a value that is not expected.

I was expecting we may have other legitimate API callers that aren't state-affiliated: developers (as you note), federal users (investigators? quality control?), ISO, etc. Is that the case? If so, in that case I was anticipating the orchestrator would expect `initiating_state` to be an empty string.

iac/apim-duppart-policy.xml
kingcomma: 👍 To tightening up the regex and providing more breadcrumbs.

Yep, as you noted, `context.Subscription.Name` equates to the "Display Name" entered into the portal while `context.Subscription.Id` equates to the "Name" field in the portal (or subscriptionId in the PowerShell and REST APIs). I wrote the [naming instructions](https://github.com/18F/piipan/blob/f6370f12e313fee7b6f22784890ec290b32d0385/match/docs/duplicate-participation-api.md#managing-api-keys) with this in mind — just setting Display Name and Name to identical values to avoid confusion.

iac/apim-duppart-policy.xml
kingcomma: Believe it or not, if there is no matching group name this still returns a [Group object with `Value` set to `String.Empty`](https://docs.microsoft.com/en-us/dotnet/api/system.text.regularexpressions.groupcollection?view=net-5.0#remarks). I could be more explicit by first checking that the GroupCollection contains the key rather than relying on this quirk (as I see it). See also: https://stackoverflow.com/questions/16392337/explanation-for-having-regex-match-group-no-error

In general, APIM policy errors are passed to the `on-error` policy section where you can implement custom error handling. If a policy expression causes an unhandled exception, APIM returns a 500 error with an associated activity ID that can be used as a reference to lookup the related log message that is being streamed to our event hub instance. (Of course, we developers don't yet have a good way of gaining insight into those events).

iac/apim-duppart-policy.xml
kingcomma: Yep, this is an oversight.

Per your comment above about deleting any incoming `meta.initiating_state` property, this is where I intend for that to happen. The value is either set to the value derived from the subscription name—which could be an empty string if the subscription name does match the prescribed pattern—or the default (which will also be set to an empty string).

match/docs/duplicate-participation-api.md
kingcomma: It's a bit more work, but I was anticipating this being done in the OpenAPI schema for the internal API endpoint. I.e., the internal-facing doc would reference a required `meta.initiating_state` value but that value would be absent from the external-facing doc.

kingcomma:
@ryanhofdotgov Just requested your review on this.

After our conversation, I proceeded with the following:
- Pass the initiating state value as a header rather than modifying the request body
- Continue to derive the state abbreviation from the subscription name for now, being as defensive as possible (could later be improved using APIM products or some sort of map stored as a string in a named value)
- If a caller's state abbreviation can not be derived, default to "ea"
- Update the OpenAPI schema to document the divergence between the external and internal APIs

Partially addresses #1553

iac/apim-duppart-policy.xml
```diff
@@ -10,6 +10,31 @@
...
+        -->
+        <choose>
+            <when condition="@(Regex.IsMatch(context.Subscription.Name, "^[A-Z]{2}-DupPart"))">
+                <set-variable name="InitiatingState" value="@{
+                    string pattern = @"^(?<state>[A-Z]{2})-DupPart";
+                    var matches = Regex.Match(context.Subscription.Name, pattern);
+                    return matches.Groups["state"].Value.ToLower();
+                }" />
+            </when>
+            <otherwise>
```
ryanhofdotgov, line 28: I'd recommend against setting a default value here. Instead, force devs to use API subscription keys called EA-DupPart, etc. and derive state affiliation of query tool users in the query tool.

When I was talking about defaults, I was referring to the derivation in the query tool, but on second thought the state affiliation might be able to be set in the partner agency's user meta store, at account provisioning, rather than inferring it from email domain.

iac/apim-duppart-policy.xml
```diff
@@ -10,6 +10,31 @@
...
+                <set-variable name="IntitiatingState" value="ea" />
+            </otherwise>
+        </choose>
+        <!--
+            Pass derived state abbreviation to backend as X-Initiating-State header
+        -->
+        <set-header name="X-Initiating-State" exists-action="override">
+            <value>@(context.Variables.GetValueOrDefault<string>("InitiatingState", ""))</value>
```
ryanhofdotgov, line 36: Is there a way to only set the header if there is a value? Shades of [null pointer references: billion dollar mistake](https://www.lucidchart.com/techblog/2015/08/31/the-worst-mistake-of-computer-science/)

iac/apim-duppart-policy.xml
ryanhofdotgov: To clarify, from a security principle, if the request has some unexpected property, we should fail the request as soon as we know that and not try to make it still go.

kingcomma: I've changed this to remove the use of a default and return a 401 error if the subscription name does not match the prescribed format.

iac/apim-duppart-policy.xml
kingcomma: This is now the case — the header is set immediately after the variable. I've kept the step of setting the variable (rather than just setting the header directly) as I expect it will make sense to move that step higher up the policy chain at some point.

iac/apim-duppart-policy.xml
```diff
@@ -10,6 +10,38 @@
...
+            </when>
+            <otherwise>
+                <!-- Subscription name does not match the prescribed format -->
+                <return-response>
+                    <set-status code="401" reason="Unauthorized"/>
+                    <set-header name="Content-Type" exists-action="override">
+                        <value>application/json</value>
+                    </set-header>
+                    <set-body>{
+                         "statusCode": 401,
```
ryanhofdotgov, line 39: Almost there – thanks for bearing with me as I scrutinize this.

Can you say a little bit as to the choice of this JSON response structure? I was expecting something like the examples in the spec.

What do you think about using 500 Internal Server Error instead? If a client manages to trigger this error, there is nothing they can do to correct it. Their authn credentials are valid if they got this far. Instead this error means someone at the agency set up their credentials wrong – basically an `// This code will never execute` 😅  

If you were filtering on the HTTP request logs in the SIEM, you'd probably want to look for/alert on 500s more carefully than for 401 which can be caused by user error, so that's another reason too.

kingcomma: I'm good using a `500` — the client not having any course of action and the SIEM tool point are plenty compelling 😅 

To answer your question though, the current structure mimics the pattern for responses returned by APIM when a subscription key does not have access to the endpoint it is calling:

```
401 Access Denied
{
    "statusCode": 401,
    "message": "Access denied due to invalid subscription key. Make sure to provide a valid key for an active subscription."
}
```

Since the this "initiating state" value is partially an authorization step, `401` felt sensible at the time. It also could serve to differentiate between an error we "expect" (or at least can anticipate) and a wholly unanticipated error which would cause APIM to return its standard `500` response.

ryanhofdotgov: 👍 In general, we would not send back a detailed message to the client for an internal server error. That would usually reveal an internal implementation detail. If the code is open source that argument goes out the door a little bit.

Is there a way to send a 500 without a body, but also log an error to APIM the log?

kingcomma: Yep, it's easy to return a 500 without a body.

Logging appears to be possible as well, but [more involved](https://docs.microsoft.com/en-us/azure/api-management/api-management-howto-log-event-hubs) — create a custom logger using the REST API, link it to event hub, send the custom event. In the interest of time, are you ok with splitting the logging component off into another issue that we address in a subsequent sprint?

ryanhofdotgov: Yeah, let's skip the addition log entry. The 500 response can be sufficient for now.

ryanhofdotgov (approved):
Thanks!

PR #1589, opened 2021-09-13
Refactor Metrics subsystem

### LINKED ISSUE #1559: Idiomatic .NET coding style
Review Piipan implementation for common .NET patterns.  This should make it easier for experienced .NET programmers to onboard and begin work on Piipan.

AC:
- [ ] Make recommendations for refactoring
- [ ] Implement some of the recommendations - keep work timeboxed within the bounds of this sprint (likely to be a single example within a subsystem)

timoballard:
Closes #1559

Refactors the `metrics` subsystem to be in line with [standard subsystem architecture](https://github.com/18F/piipan/blob/dev/docs/adr/0018-standardize-subsystem-software-architecture.md)

## Notes 
- Create `Piipan.Metrics.Core` project
- Create `Piipan.Metrics.Core.Tests` project
- Rename `Piipan.Metrics.Api` project to `Piipan.Metrics.Func.Api`
- Rename `Piipan.Metrics.Collect` project to `Piipan.Metrics.Func.Collect`
- Create `Piipan.Metrics.Func.Api.Tests` project
- Create `Piipan.Metrics.Func.Collect.Tests` project
- Rename `Piipan.Metrics.Models` project to `Piipan.Metrics.Api`


echappen:
_suggestion_: Still reviewing this, but the filename`IParticipantUploadsApi.cs` might need to be updated to `IParticipantUploadApi.cs`

echappen:
_question_: As I’m fairly new to Dotnet idioms, I gathered a few assumptions while reviewing this. Can you confirm whether these are correct? (happy to do this by video if that’s faster)

- There should be an intermediary Service between Data Access Objects and Api’s
- `IEnumerable` is preferable to `Array` or `List` when defining class properties (I think you’ve touched on this before)
    - If you need list-like functionality down the road, use `someEnumerable.ToList()`
-  Extensions are the preferable way of adding lower-level “helper” functions

timoballard:
@echappen - I'll take pass here, happy to jump on a call to chat more if you'd prefer-

- the `Controller -> Service -> DataAccess` setup is pretty common in my experience, yeah (the Function is playing the role of controller here). In this case it looks a little weird because the `ParticipantUploadService` doesn't really _do_ anything, it's just a pass through to the DAO. The thinking there is that there shouldn't be any application logic in the data access object. We don't happen to have any application logic at this point -- we just query the data and return it as is -- but if in the future we needed to do some kind of additional transformation or filtering or whatever on that data after querying it, that would live in the Service layer.
- Yeah, exactly - it seems to be a best practice to use the most abstract / least specific [collection interface](https://docs.microsoft.com/en-us/dotnet/api/system.collections?view=netcore-3.1#interfaces) you can, given your use case
- I'm not sure what the best practices are around extensions to be honest. I tend to use them in cases where, like you say, you want to add some helper functionality to a type, but where that functionality doesn't depend on anything else. Extensions are just syntactic sugar for static functions, so you can't use injected dependencies or anything. The other consideration is that because they're statics under the hood, they can't be mocked, so they can complicate unit testing in certain scenarios

Hope that helps!

echappen:
That helps, thanks!

echappen:
This refactor looks great to me—kudos on the 9% spike in code coverage. But I can give more time for @kingcomma and @ryanhofdotgov to weigh in before I hit approve, if you all need it.

ryanhofdotgov:
@echappen @timoballard No need to wait for me for review 🚀

kingcomma:
@echappen @timoballard Same with me, no need to wait for my review.

echappen (approved):
🚀 🚀 🚀

PR #1592, opened 2021-09-14
Assert correct number of records when testing Read

echappen:
Used these to quickly verify that records were being returned from Read. Otherwise the Assertions in the iterations have the potential to not be run, and the tests return a false positive.


kingcomma (approved):
👍

Wonder if it's worth adding something similar to `Piipan.Etl.IntegrationTests` at some point? I guess since the ETL app is currently all or nothing, the existing integration test implicitly confirms the row count (if there's one row, you assume all the rows made it in). So it's probably not necessary until if / when we start loading participants row by row.

PR #1593, opened 2021-09-14
Create de-identification library for query tool

### LINKED ISSUE #1540: Allow internal resources to find matches using PII
Split off from #1215 

Implementation details:
- [x] An agreed upon approach for accepting/normalizing non-ASCII characters via the Query Tool and/or API (see #1544)
  - this is a library, likely in Piipan shared
- [ ] A shared tool for normalization and hash generation
~- [ ] Implementation of the [PII-based API endpoint]~(https://github.com/18F/piipan/blob/dev/match/src/Piipan.Match.Orchestrator/Api.cs#L109)

echappen:
Closes #1540 (?)

This creates a Class Library to de-identify PII for Bulk Upload and Duplicate Participation API ingestion. Initially it'll be used by the Query Tool to prepare user input for match querying. 

### Notes
- In order to fix the Query Tool within this sprint, I handled the easiest normalization elements and deferred the harder ones (like suffix removal) until after the Query Tool is fixed. 
- The unit tests describe all normalization elements, skipping the tests for the unimplemented ones. I hope this makes it clear to others which elements are pending.
- I'm sure the class/methods could be made more idiomatic of Dotnet.


shared/src/Piipan.Shared/Deidentification.cs
```diff
@@ -0,0 +1,131 @@
...
+using System.Text;
+using System.Text.RegularExpressions;
+namespace Piipan.Shared.Deidentification
+{
+    /// <summary>
+    /// Tool for deidentifying Personally Identifiable Information (PII)
+    /// in the acceptable format for the Bulk Upload and Matching systems
+    /// as described in PPRL approach (/docs/pprl.md)
+    /// </summary>
+    public class Deidentifier
```
timoballard, line 13: _comment_: I love the approach of having the component methods (`NormalizeName`, `NormalizeSsn`, etc) available, but having the top-level `Run` method that runs the whole batch

_suggestion_: I think rather than statics, we should have proper objects here which inherit from interfaces. That should allow us to inject this deidentification component where we need it without leaking implementation details. What do you think about building the `Deidentifier` up from object composition, such as:

```
Deidentifier : IDeidentifier
    INameNormalizer
    ISsnNormalizer
    IDobNormalizer
    IHasher
```

There are tradeoffs here and I don't want to overcomplicate this, but I think something like above would set us up nicely for any future normalization/hashing/matching algorithm changes. Thoughts?

timoballard:
Kudos on the thorough and clear comments in this code - so helpful and will make for a great reference library!

shared/src/Piipan.Shared/Deidentification.cs
echappen: I'm ok with trying this! My thought was that since all the tasks are purely functional/stateless, it wasn't dangerous to expose it. But I agree this would be a better setup for future changes. 

To make sure I'm understanding how this would prevent leaking of implementation details—most callers could just need to implement `IDeidentifier` to run the whole process, and `IDeidentifier` would only expose the `Run` method. But in the case where a caller wanted to run a sub-process, it could implement something like `INameNormalizer` which would only expose the `NormalizeName` method. Is that correct?

shared/src/Piipan.Shared/Deidentification.cs
```diff
@@ -0,0 +1,131 @@
...
+                throw new ArgumentException("name must contain only ascii characters");
+            }
+            // Convert to lower case
+            string result = lname.ToLower();
+            // TODO: Remove any suffixes(e.g.; junior, jnr, jr, jr., iii, etc.)
+            // Replace hyphens with a space
+            result = result.Replace("-", " ");
+            // TODO: Remove any character that is not an ASCII space(0x20) or in the range[a - z] (0x61 - 0x70)
+            // Replace multiple spaces with one space
+            result = result.Replace("  ", " ");
```
kingcomma, line 54: *Question:* will this catch instances of more than two consecutive spaces? E.g., does `Foo    Bar` (four spaces) get transformed to `Foo Bar` (one space) or `Foo  Bar` (two spaces)?

shared/src/Piipan.Shared/Deidentification.cs
timoballard: Yep! For components that need to do deidentification, they really just need some interface that does deidentifcation, and the less they have to "know" beyond that the better (e.g. normalization is an implementation detail of this particular deidentifier and is irrelevant to the consumer in the common case). Going a step further down, our initial set of name normalization rules are an irrelevant detail to the deidentifier -- it just needs an interface that does normalization, whatever that happens to mean.

These patterns are nicely summarized by [SOLID](https://en.wikipedia.org/wiki/SOLID), most specifically the `S` and `D` in this case

kingcomma: At the risk of thinking too far ahead, but also on the topic of future changes — I wonder if it's worth anticipating additional hashes and setting the precedent of having a `Deidentification` namespace with individual classes named after each hash/de-identification implementation (e.g., `LdsHash`, `FdsHash`, etc)? Not trying to prescribe naming, just thinking about organization.

shared/src/Piipan.Shared/Deidentification.cs
echappen: Yep, good catch!

shared/src/Piipan.Shared/Deidentifier.cs
```diff
@@ -0,0 +1,39 @@
...
+        private readonly ISsnNormalizer _ssnNormalizer;
+        private readonly IDobNormalizer _dobNormalizer;
+        private readonly IHasher _hasher;
+
+        public Deidentifier(){
+            _nameNormalizer = new NameNormalizer();
+            _dobNormalizer = new DobNormalizer();
+            _ssnNormalizer = new SsnNormalizer();
+            _hasher = new Hasher();
+        }
```
timoballard, line 22: This looks great - and a quick turnaround!

_suggestion_: We should have these 4 dependencies injected via the constructor -- that way the `Deidentifier` never has to know which particular implementation of those interfaces it's dealing with. The object construction will be handled by the [dependency injection framework](https://docs.microsoft.com/en-us/dotnet/core/extensions/dependency-injection), so only one place in the code (the "composition root", i.e. `Startup.cs` in .NET core apps) has to know which implementation we want to actually use for each interface type. There's a general guideline with apps that use dependency injection where you really shouldn't see the word `new` anywhere in the code, except when dealing with plain old data objects. Service-like classes should pretty much always be `newed` up by the DI framework and injected as dependencies

_nitpick_: piggy-backing on @kingcomma's comment from earlier, it might make sense to name this class and interface something like `ILdsDeidentifier & LdsDeidentifier`, since the interface is specifically for the case where we're operating on last name, ssn, and dob as inputs. If sometime later we started taking in first name, this interface would have to change (or more likely, we'd introduce a new one, something like `IFldsDeidentifier`)

shared/src/Piipan.Shared/Deidentification.cs
echappen: I updated this to use an Interface approach. I was unsure of how much to expose in the IHasher interface, since the underlying class does three things instead of one thing (which was the norm with the rest of the interfaces). Let me know if this is on the right track.

shared/tests/Piipan.Shared.Tests/NameNormalizerTests.cs
```diff
@@ -0,0 +1,109 @@
...
+
+            // Remove any character that is not an ASCII space (0x20) or in the range [a-z] (0x61-0x70)
+            [Theory (Skip = "not yet implemented")]
+            [InlineData("foo.")]
+            public void removeAsciiNotInRange(string name)
+            {
+                string result = _nameNormalizer.Run(name);
+                Regex rgx = new Regex(@"[^a-z|\s]");
+                Assert.DoesNotMatch(rgx, result);
+            }
```
kingcomma, line 100: *Suggestion (non-blocking):* I know this is forthcoming, so just noting it would be good to tighten this regex up a bit by only accepting `\x20` rather than all white space characters. And perhaps testing other common scenarios like `f'bar` (i.e., `o'connor`).

shared/tests/Piipan.Shared.Tests/LdsDeidentifierTests.cs
```diff
@@ -0,0 +1,42 @@
...
+            }
+
+            [Theory]
+            [InlineData("Foobar", "2000-12-29", "111-11-1111")]
+            [InlineData("von Neuman", "2020-01-01", "111-11-1111")]
+            public void returnsValidDigest(string lname, string dob, string ssn)
+            {
+                string result = _ldsDeidentifier.Run(lname, dob, ssn);
+                Assert.Matches("^[0-9a-f]{128}$", result);
+                Assert.Equal(128, result.Length);
```
timoballard, line 31: _question_: in addition to asserting on the shape of the result, do you think we should assert on the actual hash value being what we would expect?

timoballard (approved):
👏

PR #1638, opened 2021-09-15
Add ADR for continuing with .NET Core 3.1

### LINKED ISSUE #1231: Spike: Evaluate work load of .NET 3.1 to .NET 5 migration
- Confirm support exists for App Service, Function Apps in US Government Cloud
- Test one web app, test one function app

timoballard:
Closes #1231 

Adds an ADR for our decision to continue with .NET Core 3.1


ryanhofdotgov:
@timoballard I'm not sure if it is accurate to say that Azure Government is the reason we aren't migrating to .NET 5 right now. That's was just a hypothesis of FastTrack. Isn't the main reason is that the basic `dotnet` commands to publish to Function App seems to be broken in the last available release – and this is affects both the public cloud and AzureUSGovernment.

timoballard:
@ryanhofdotgov Yeah I'll add the function publishing bug as a reason. When I mentioned a lack of support in Azure gov, I was (not so clearly) referring to the link that FastTrack sent: https://docs.microsoft.com/EN-US/azure/azure-government/compare-azure-government-global-azure#azure-functions

ryanhofdotgov:
@timoballard Unfortunately that web page is notoriously out of date. We've had to get them to bug someone to update it once or twice this past year.

timoballard:
@ryanhofdotgov - good to know, thanks! I'll update this to stick to what we know for sure

ryanhofdotgov (approved):
Thanks!

PR #1660, opened 2021-09-16
fix query tool by using de-identification library for match request

### LINKED ISSUE #1330: Update query website after APIs use deidentification
Issue #1215 ~~is likely to break~~ broke the query website.  Now that bulk uploads accept deidentified PII, we need to update the query website to allow eligibility workers to check for NAC matches by entering PII into the NAC website.

- [ ] Website-based match queries using raw PII (name, DOB, SSN) work when users enter PII that doesn't need normalization
- [ ] (open question of whether this should be in-scope) Inputs are normalized (related issues: #1217, second AC of #1540)

~~Blocked until #1215 is complete.~~

May depend on #1540

echappen:
Closes #1330 

### Dev Notes
- ~Still working through 2 failing tests, but the code is reviewable at this point.~ Fixed
- We were using `PiiRecord` for both form input and API response data and that was getting confusing, so I finally split up all the request/response models into their own files. `PiiRecord` now only represents the form input. 
- According to the [new subsystem architecture](https://github.com/18F/piipan/blob/dev/docs/adr/0018-standardize-subsystem-software-architecture.md), would it be appropriate to put these models into a project like `Piipan.QueryTool/Piipan.QueryTool.Api`?
- I chose `AddSingleton` to add the `ILdsDeidentifier` interface in `Startup.cs` so I could instantiate it the way I needed to. Lmk if there's a better way to do this. 

### Screens
querytool_deid_results


query-tool/src/Piipan.QueryTool/Startup.cs
```diff
@@ -63,6 +64,16 @@ public void ConfigureServices(IServiceCollection services)
...
 
+            services.AddSingleton<ILdsDeidentifier>((s) =>
+            {
+                return new LdsDeidentifier(
+                    new NameNormalizer(),
+                    new DobNormalizer(),
+                    new SsnNormalizer(),
+                    new LdsHasher()
+                );
+            });
```
timoballard, line 75: I think your choice of `AddSingleton` is appropriate here since the deidentifier is purely functional, no external connections or anything that you might want to reestablish for each incoming request.

An equivalent, but I think more idiomatic way to do this sequence of service registrations would be:

```
services.AddSingleton<INameNormalizer, NameNormalizer>();
services.AddSingleton<IDobNormalizer, DobNormalizer>();
services.AddSingleton<ISsnNormalizer, SsnNormalizer>();
services.AddSingleton<ILdsHasher, LdsHasher>();
services.AddSingleton<ILdsDeidentifier, LdsIdentifier>();
```

Main advantage of ☝️  being that the service provider then knows how to build the normalizers and hasher independently of the `LdsDeidentifier`

timoballard:
RE: your question about where to put the request/response models -- I would leave them in the `QueryTool` project for the time being. I suspect that once we're done refactoring the `metrics/etl/match` subsystems, we'll end up opting for a slightly different project structure for the web apps

query-tool/src/Piipan.QueryTool/Pages/Index.cshtml.cs
```diff
@@ -39,8 +43,13 @@ public async Task<IActionResult> OnPostAsync()
                 {
                     _logger.LogInformation("Query form submitted");
 
-                    MatchResponse result = await _apiRequest.Match(Query);
-
+                    string digest = _ldsDeidentifier.Run(
+                        Query.LastName,
+                        Query.DateOfBirth.GetValueOrDefault().ToString("yyyy-MM-dd"),
```
kingcomma, line 48: *Question:* Even though `DateOfBirth` should be set to a value if `ModelState.IsValid`, is it still possible this will return the default value (`"0001-01-01"`)? Would it be more defensive to use `Query.DateOfBirth.Value` and cause an exception in the unexpected event the property is not set?

query-tool/src/Piipan.QueryTool/Pages/Index.cshtml.cs
```diff
@@ -39,8 +43,13 @@ public async Task<IActionResult> OnPostAsync()
                 {
                     _logger.LogInformation("Query form submitted");
 
-                    MatchResponse result = await _apiRequest.Match(Query);
-
+                    string digest = _ldsDeidentifier.Run(
+                        Query.LastName,
+                        Query.DateOfBirth.Value.ToString("yyyy-MM-dd"),
+                        Query.SocialSecurityNum
+                    );
```
kingcomma, line 50: *Thought (non-blocking):* I've been wondering if it would make sense to add `LdsHash` (and any future hashes) to the underlying model as read-only properties. Haven't really thought through the implications of that decision though, and certainly not something that needs to be explored now.

kingcomma (approved):
I defer to @timoballard on the service registration approach, but from a "fix the query tool" standpoint this looks good to me! 👍

PR #1661, opened 2021-09-16
Adding manual test script for APIM DupPart

ryanhofdotgov:
In parallel with the existing test script that uses the dev's Azure identity to bind to the orchestrator API and submit a match query, add a test script that makes the call through APIM. Pull the EA subscription key and use that for the query.


match/tools/test-orch-match-api.bash
```diff
@@ -8,12 +8,14 @@
...
 #
-# usage: test-match-api.bash <azure-env>
+# usage: test-orch-match-api.bash <azure-env>
 
 # shellcheck source=./tools/common.bash
 source "$(dirname "$0")"/../../tools/common.bash || exit
 
 MATCH_API_FUNC_NAME="find_matches"
+
+# Hash digest for farrington,10/13/31,000-12-3456
```
kingcomma, line 18: Thanks for adding this.

kingcomma (approved):
👍 Looks good to me — thanks for this!

PR #1662, opened 2021-09-16
Adding manual test script for APIM bulk upload

ryanhofdotgov:
In parallel with the existing test script that uses the dev's Azure identity to directly upload to a state storage account, add a test script that makes the bulk upload call through APIM. Pull the EA subscription key and use that for the query.


etl/docs/etl.md
```diff
@@ -43,16 +43,16 @@ func azure functionapp publish <function-app-name> --dotnet
 
 ## Ad-hoc testing
 
-In a development environment, the `upload.bash` tool can be used to upload test CSV files to a storage account.
+In a development environment, the `test-storage-upload.bash` tool can be used to upload test CSV files to a storage account.
 
 For example, if you are targeting the `tts/dev` environment and the `eastate5or4l3nqzevf4` storage account:
 ```
-./tools/upload.bash tts/dev ./docs/csv/example.csv eastate5or4l3nqzevf4
+./tools/test-storage-upload.bash tts/dev ./docs/csv/example.csv eastate5or4l3nqzevf4
```
kingcomma, line 50: I'll create a follow-on PR to update this example storage account name to match our current naming structure (`ttssteauploaddev`).

kingcomma (approved):
👍 👍

etl/docs/etl.md
ryanhofdotgov: Good idea – thank you!

PR #1663, opened 2021-09-16
Add ADR for storing and converting time

### LINKED ISSUE #1474: Add ADR about how we handle Dates and Times throughout system
Based on [this thread](https://github.com/18F/piipan/pull/1433#issuecomment-905533989), we'd like to describe our general approach to storing and working with Dates and Times in the Dotnet ecosystem. 

Decisions:
- Store datetimes as `timestamp` in postgresql without offset. All dates in DB are assumed UTC.
- Specify `Kind` when working with DateTimes in Dotnet to avoid `Unspecified` DateTimes and ambiguity in downstream timezone conversions.

echappen:
Closes #1474 

I drew a blank on the Consequences section.


edwintorres (approved):

docs/adr/0020-storing-and-converting-time.md
```diff
@@ -0,0 +1,33 @@
...
+
+Date: 2021-09-16
+
+## Status
+
+Accepted
+
+## Context
+
+We use Postgresql for our databases. Postgres has two ways of storing times and timestamps: `without time zone` (default) and `with time zone`. There's no right one to use, but they [convert timezones differently](https://www.postgresql.org/docs/11/datatype-datetime.html) to and from the database, and the fact that `without time zome` is the default suggests this is Postgres' general preference.
```
ryanhofdotgov, line 11:
```suggestion
We use Postgresql for our databases. Postgres has two ways of storing times and timestamps: `without time zone` (default) and `with time zone`. There's no right one to use, but they [convert timezones differently](https://www.postgresql.org/docs/11/datatype-datetime.html) to and from the database, and the fact that `without time zone` is the default suggests this is Postgres' general preference.
```

ryanhofdotgov: It doesn't need to be called out here, but one of the reasons why you might not have timezones in your column is to implement a defensive approach – if you don't need to store future dates that might be affected by DST, etc. and you've agreed to always store UTC, then you don't want your column to have the possibility of storing something different.

ryanhofdotgov (approved):

docs/adr/0020-storing-and-converting-time.md
kingcomma: I was curious about Postgres' preference and found a couple interesting, new-to-me bits of context. But I don't think this changes anything on our end since we've already decided to always store in UTC.

- [`timestampz` is marked as `typeispreferred`](https://stackoverflow.com/questions/9571392/ignoring-time-zones-altogether-in-rails-and-postgresql/9576170#9576170)
- [PostgreSQL Wiki: Don't use timestamp (without time zone)](https://wiki.postgresql.org/wiki/Don't_Do_This#Don't_use_timestamp_(without_time_zone))

ryanhofdotgov: That second link has an interesting argument, and I suppose you could accomplish the goal of preventing other timezones being inserted by having a constraint check on the column. It's a self-documenting approach.

I have to think about whether or not the psql/libpq behavior in the first link would create other problems/confusion. And how that works when your PostgreSQL hosting provider does not use UTC for their system time (less common nowadays with cloud PaaS).

echappen: @kingcomma This is new to me as well. I think I should modify the ADR (breaking my own rule that they should be immutable) in order to:
1. change the bit about `timestamp [without time zone]` being implicitly preferred and say that this is actually not preferred by the Postgres community
2. add [these items](https://wiki.postgresql.org/wiki/Don't_Do_This#Don.27t_use_timestamp_.28without_time_zone.29_to_store_UTC_times) to the Consequences section:
> There is no way for the database to know that UTC is the intended timezone for the column values.
> This complicates many otherwise useful time calculations [with examples]

Issue #1659, opened 2021-09-16
Refactor etl subsystem


timoballard:
Refactor the `etl` subsystem to align with the [standard subsystem architecture](https://github.com/18F/piipan/blob/dev/docs/adr/0018-standardize-subsystem-software-architecture.md).

PR #1672, opened 2021-09-17
Introduce participants subsystem

### LINKED ISSUE #1669: Introduce participants subsystem
Add a new top-level `participants` subsystem that provides a API for interacting with the participants database. Project organization for this subsystem will align with the [standard subsystem architecture](https://github.com/18F/piipan/blob/dev/docs/adr/0018-standardize-subsystem-software-architecture.md), as seen [here](https://docs.google.com/drawings/d/1vz22qBCP3ivH819hB8trDiG9OGV9Ikj5-07DJ0cwrFE). Once available the `etl` and `match` subsystems should be refactored to use the `participants` service(s) instead of accessing the database directly.

Relates to #1659 
Relates to #1589

timoballard:
Closes #1669 

Adds a new top-level `participants` library for managing interactions with the state databases.

### Notes
- As of this PR, this library is not being consumed by any subsystems. #1658 and #1659 detail the work to refactor the `match` and `etl` subsystems to make use of it.


participants/tests/Piipan.Participants.Core.IntegrationTests/ParticipantDaoTests.cs
```diff
@@ -0,0 +1,123 @@
...
+using Piipan.Participants.Core.Models;
+using Piipan.Participants.Core.Services;
+using Dapper;
+using Npgsql;
+using Xunit;
+using Newtonsoft.Json;
+
+namespace Piipan.Participants.Core.IntegrationTests
+{
+    [Collection("Core.IntegrationTests")]
```
timoballard, line 18: This was new to me so sharing in case it's useful to others: you can prevent xUnit from parallelizing tests spread across multiple files/classes by adding them to the same `Collection` using this attribute. This turned out to be necessary for the `participant` & `upload` integration tests, as they both try to apply the database schema during test setup

participants/src/Piipan.Participants/Piipan.Participants.Core/Services/ParticipantService.cs
```diff
@@ -0,0 +1,57 @@
...
+        public async Task AddParticipants(IEnumerable<IParticipant> participants)
+        {
+            if (!participants.Any())
+            {
+                throw new ArgumentException("participants cannot be empty!");
+            }
+
+            var upload = await _uploadDao.AddUpload();
+
+            var participantDbos = participants.Select((p) =>
```
ryanhofdotgov, line 40: This looks like it will create an object for each CSV record in memory. I believe the existing implementation avoids slurping in the entire data, though it all kind of depends on [cvs.GetRecords](https://github.com/18F/piipan/blob/3518cb97cdfa33e22c982a92b0e84cf375db11e5/etl/src/Piipan.Etl/BulkUpload.cs#L80) doing what we'd expect or something smarter, which admittedly I haven't verified.

Could you check this math, and my understanding of what the memory footprint will look like with this code?
* lds hash: ~18 + (2 * `128`) = 274 bytes
* case id: ~18 + (2 * ~`12`) = 42 bytes
* participant id: ~18 + (2 * ~`12`) = 42 bytes
* benefits end date: ~8 + `8` bytes = 16 bytes
* recent benefit months: ~8 + `24` bytes = 32 bytes
* upload id = ~8 + `4` bytes = 12 bytes

Let's say ~400 bytes per record. [Around ~4,000,000 records is current upper bound of CSV file in 2021](https://www.zippia.com/advice/10-states-people-food-stamps/). That's ~1.5GB.

If this number is right, could we restructure so as to lower the working memory requirements and work in a more streaming/iterator type fashion? 

References: https://www.red-gate.com/simple-talk/blogs/how-big-is-a-string-in-net/,

timoballard: Thanks for calling this out - I didn't fully realize how big the CSV files can get! 

That said, I _think_ we're still ok here, if my memory serves (😏 ). The execution of LINQ transformations like `Select` is deferred until the `IEnumerable` is actually enumerated or projected into something like a `List`, which in this case happens [here](https://github.com/18F/piipan/blob/0542d16de2f8cb733ec595872101739390c6eee3/participants/src/Piipan.Participants/Piipan.Participants.Core/DataAccessObjects/ParticipantDao.cs#L66-L69) in the `ParticipantDao`. As you mentioned, `csv.GetRecords` yields the rows one at a time, so that should carry all the way through and prevent everything getting loaded at once. Definitely something to test and verify once `etl` is wired up.

ryanhofdotgov: Nice 👍

ryanhofdotgov (approved):

PR #1676, opened 2021-09-17
Use current working directory in bulk upload test script

kingcomma:
Missed this when reviewing #1662.


ryanhofdotgov (approved):
Nice catch – thanks!

Issue #1546, opened 2021-09-04
Create Persistent Match Records


cjhskippy:
When a match is found, create a record that can be used to reference the match again later.  For this story, we can create a new record every time a query results in a match, but that behavior will be refined in [story link]

The match record should include:
- [ ] A unique match ID (should follow the same formatting guidance as we used for lookup IDs)
- [ ] The hash representing the deidentified PII (name, DOB, SSN) - this will be used to determine whether or not to create a new match.
- [ ] The date + time when the match was found
- [ ] The states involved in the match. (At this stage, we can assume only two states per match, but this will likely change eventually)
- [ ] If the match is invalid (this should default to false)
- [ ] A status for the match.  This should default to "Open".

kingcomma:
@cjhskippy @ryanhofdotgov I'm trying to get a handle on the scope of this work in the context of the current sprint. To satisfy this ticket, I'm planning on:
- A single table (e.g, `matches`) where match records are stored
- The minimum amount of columns on that table necessary to capture the data above
- No relationships (e.g., a `states` table that is referenced by `matches`)
- No attempt to define the possible values for match status, match disposition, or participant eligibility.

Is that sufficient?

cjhskippy:
@kingcomma from my perspective, as long as you're storing the data, I've got no preference on how that gets done.  Totally okay with me if we don't try to enforce a limited set of values for match status, validity, or disposition right now.

cjhskippy:
Something that was left out of the original AC (and can stay out, but will need its own issue if so) - we should verify that multi-person queries can produce multiple match records after this shift.

PR #1685, opened 2021-09-20
Refactor IaC database tools to common script

### LINKED ISSUE #1493: Metrics db schema needs to be externalized into DDL file
The schema for the Metrics subsystem is embedded in the IaC for the core database cluster:
https://github.com/18F/piipan/blob/eaef1ef3654002079fd8afe9098ded22457dac45/iac/create-core-databases.bash#L37

This should be externalized into a DDL file, much like ddl/per-state.sql, to allow for easier schema migration and basic db integration testing.

kingcomma:
Moves database-related functionality out of `create-core-databases.bash` and into a share common script. Namespaces/prefixes all functions with `db_`.

Prep work for creating the match records database required by #1546.

Closes #1493, incidentally.

There's some follow on work to update `create-databases.bash` to use the common functionality where possible.

### DIFF
--- a/metrics/ddl/metrics.sql
+++ b/metrics/ddl/metrics.sql
@@ -1,0 +1,11 @@
+BEGIN;
+
+CREATE TABLE IF NOT EXISTS participant_uploads(
+    id serial PRIMARY KEY,
+    state VARCHAR(50) NOT NULL,
+    uploaded_at timestamp NOT NULL
+);
+
+COMMENT ON TABLE participant_uploads IS 'Participant bulk upload event record';
+
+COMMIT;
--- a/iac/create-core-databases.bash
+++ b/iac/create-core-databases.bash
@@ -13,139 +13,10 @@ set_constants () {
   VAULT_NAME=$PREFIX-kv-core-$ENV
   PG_SECRET_NAME=core-pg-admin
 
-  PSQL_OPTS=(-v ON_ERROR_STOP=1 -X -q)
-  TEMPLATE_DB=template1
-
   # Name of Azure Active Directory admin for PostgreSQL server
   PG_AAD_ADMIN=piipan-admins
 
   PRIVATE_DNS_ZONE=$(private_dns_zone)
-}
-
-init_db () {
-  db=$1
-  owner=$db
-
-  create_role "$owner"
-  config_role "$owner"
-
-  create_db "$db"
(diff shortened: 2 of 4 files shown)


metrics/ddl/metrics.sql
```diff
@@ -0,0 +1,11 @@
+BEGIN;
+
```
echappen, line 2: A few weeks ago when I split up the per-state SQL from the per-state controls into their own transactions, I had to add a `SET search_path=piipan,public;` here in order for this transaction to work.

kingcomma: In this case, I don't think that should be necessary. The purpose of the custom `piipan` schema in the participants cluster is to provide a level of isolation given the "multi-tenancy" nature of that cluster (@ryanhofdotgov please correct me if I'm wrong).

So far, we have not applied that practice to other sub-systems/clusters, and [have stuck with the `public` schema](https://github.com/18F/piipan/issues/409#issuecomment-865350359).

echappen (approved):

ryanhofdotgov:
@kingcomma I would not say it is even a medium priority, but in general, nesting tables, etc. under an application schema has been good practice in PostgreSQL because:
* [accounts by default get create access in the public schema](https://blog.crunchydata.com/blog/postgresql-defaults-and-impact-on-security-part-2)
* PostgreSQL extensions can create resources in there, and then dumping the schema gets a mix of application specific stuff plus all the stuff that gets reinstalled when you install an extension

The main trade off is psql (or other code) not having it in its default search path, but you can also configure that if you choose to:
https://www.postgresql.org/docs/12/ddl-schemas.html#DDL-SCHEMAS-PATTERNS

Here's a couple of relevant links:
* https://wiki.postgresql.org/wiki/Database_Schema_Recommendations_for_an_Application
* https://kukuruku.co/post/postgresql-set-of-practices/

kingcomma:
Thanks @ryanhofdotgov, super helpful context.

kingcomma:
@echappen — sorry, I'm re-requesting review on this. I realized I had unintentionally created a dependency chain between `iac-common` and `db-common` that would fail under certain circumstances. I've made a small edit so `iac-common` and `db-common` are independent of each other. Just want a quick set of eyes on the change before merging.

echappen (approved):

Issue #1549, opened 2021-09-04
Queries return existing, open matches instead of creating new ones


cjhskippy:
As a state agency worker, I do not want to get a new match - and be required to take duplicate actions - if an existing match already exists.

This issue assumes that the mechanisms to flag issues as valid / invalid have not yet been implemented.

Match queries through the API and FNS website are updated to align with this decision tree:
<img width="879" alt="Screen Shot 2021-09-04 at 2 39 10 PM" src="https://user-images.githubusercontent.com/1299543/132108708-d1add27e-e964-4f1d-8271-d623eb7f94eb.png">

AC:
- [ ] If an existing match exists for the same "person" (deidentified PII hash), and the match is still open, return the information for the existing match record instead of creating a new match.
- [ ] If no existing match exists for the same person, or all existing matches that exist for the same person have a status of closed, then create a new match record.  If the match involves multiple states, it would be acceptable at this stage for the new match record to get created for any two of the matching states.
  - [ ] At the engineering team's discretion, multi-state matches that meet the above criteria may instead create a new match for each pair of states.

cjhskippy:
@kingcomma Something that I left ambiguous on the AC - what exactly does it mean to return a match?  Our API does something similar today, but the structure differs a little from what we consider a match in #1546 

Might be good for us to connect to make sure we sync on how that API changes.  I'm also assuming we'll need to at some point transmit a match URL, and realizing I don't have an issue for that yet.

PR #1686, opened 2021-09-20
Initial database schema for match records

kingcomma:
Partially address #1546 — starting with the database schema before getting into implementation.

Trying to keep this as simple as possible for meeting the needs of the ticket, knowing that requirements are still coming into focus and the underlying implementation is likely to change.

### DIFF
--- a/match/ddl/match-record.sql
+++ b/match/ddl/match-record.sql
@@ -21,7 +21,7 @@ COMMENT ON TABLE matches IS 'Match records';
 COMMENT ON COLUMN matches.match_id IS 'Match record''s human-readable unique identifier.';
 COMMENT ON COLUMN matches.created_at IS 'Match record''s creation date/time.';
 COMMENT ON COLUMN matches.initator IS 'Match record''s initiating entity.';
-COMMENT ON COLUMN matches.states IS 'States/territories involved in match.';
+COMMENT ON COLUMN matches.states IS 'State/territory pair involved in match.';
 COMMENT ON COLUMN matches.hash IS 'Value of hash used to identify match.';
 COMMENT ON COLUMN matches.hash_type IS 'Type of hash used to identify match.';
 COMMENT ON COLUMN matches.input IS 'Incoming data from real-time match request.';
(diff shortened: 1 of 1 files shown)


echappen (commented):
_question_: Would we want an `updated_at` in addition to a `created_at`? Or maybe a `closed_at`?

_question_: Would `initiator` be a state name, or a something like “bulk match”? 

_question_: How does a match record the “other” state being matched? (Based on my assumption that `initiator` would be the initiating state)

kingcomma:
@echappen In general, I'm trying to toe the line between what's required for this specific ticket and what we know we will need in the future. Specifically:

> _question_: Would we want an `updated_at` in addition to a `created_at`? Or maybe a `closed_at`?

Agree that some form of both `updated_at` and `created_at` will be necessary in the near future, but left them out absent a current use case.

> _question_: Would `initiator` be a state name, or something like “bulk match”?

This perhaps goes against my thinking above, but I left this loosely defined intentionally thinking it would currently capture the initiating state name, and may be adapted to capture the method used to identify the match (e.g., `bulk` or `api`) if/when the states involved in a match are recorded as relationships.

> _question_: How does a match record the “other” state being matched? (Based on my assumption that `initiator` would be the initiating state)

The `data` column would contain the API response which in turn contains the other state(s) in the match. I was envisioning this initial implementation as more like the old lookup records for now (recording a match event and all associated data), refining over time as the user stories dictate things like [pair-wise matches](https://gsa-tts.slack.com/archives/C018P1ZJPEC/p1632182331045300?thread_ts=1632172097.042200&cid=C018P1ZJPEC) (for example).

echappen:
_suggestion_: Thinking through how this data will be queried down the road... which states the matches pertain to is probably something that will be queried pretty often, like when checking if an open match already exists. Keeping the matching state data in a jsonb field makes a sql query slightly more complicated and more brittle if that json structure should change. It may make the queries more straightforward to break out that data into its own field like `matching_state` or something, that stores the state abbreviation as text or enum.

kingcomma:
@echappen Based on our discussion, I added a `states` column that accepts an array of two text values. Like we said, this should make it easier to perform the work necessary for #1549. E.g.,:

```
SELECT *
FROM matches
WHERE hash=@hash
  AND hash_type=@hash_type
  AND states @> @states 
  AND states <@ @states
```

I went with `states` as the name of the column to match the `state` property in the API responses.

echappen (approved):

Issue #1558, opened 2021-09-07
Spike: Testing strategy


cjhskippy:
Research strategies to implement tests that will bring the Piipan codebase closer in line with the 90% code coverage that we want as a baseline before handing off development.

- [x] service registration testing (see comment below)

ryanhofdotgov:
As part of this work, be sure to [include service registration testing as outlined here](https://github.com/18F/piipan/issues/1445#issuecomment-915351590), making sure that we do not exclude code from test coverage [like this example](https://github.com/18F/piipan/blob/369900c8da0ccd73dcd741cdee6f9500431f3c63/query-tool/src/Piipan.QueryTool/Startup.cs#L18)

ryanhofdotgov:
@timoballard Are there any conclusions that can be written up from the spike – even just a comment, if not worthy of an ADR?

timoballard:
@ryanhofdotgov - yes, absolutely. Apologies, was cleaning up tasks as the sprint was winding down and forgot to come back to this. I think an ADR for the testing approach with respect to the standard project setup makes sense - I'll add a card

PR #1708, opened 2021-09-22
Refactor ETL subsystem

### LINKED ISSUE #1659: Refactor etl subsystem

timoballard:
Closes #1659 

Refactors the `etl` subsystem to make use of the `participants` library for bulk upload execution. While not strictly necessary, the `Piipan.Etl` project was moved and re-namespaced to `Piipan.Etl.Func.BulkUpload` so as to be consistent with other subsystems.  

## Notes
- As part of this work, a [frequently duplicated piece of code](https://github.com/18F/piipan/blob/de42c339f6a93dd38a2e99fdd6f31dc1ace85c4e/shared/src/Piipan.Shared/AzurePgConnectionFactory.cs#L13-L61) for establishing database connections was moved to the `Piipan.Shared` library so that it can be reused and covered by unit tests (partially addressing #1558). This change produced transitive dependency updates across the repository, which made this PR larger than it otherwise would have been.


shared/src/Piipan.Shared/AzurePgConnectionFactory.cs
```diff
@@ -0,0 +1,63 @@
...
+using Npgsql;
+
+namespace Piipan.Shared
+{
+    public class AzurePgConnectionFactory : IDbConnectionFactory
+    {
+        // Environment variables (and placeholder) established
+        // during initial function app provisioning in IaC
+        public const string CloudName = "CloudName";
+        public const string DatabaseConnectionString = "DatabaseConnectionString";
```
kingcomma, line 16: *Question (non-blocking):* Thinking ahead to #1658, the `match` function apps are potentially bound to many databases — _N_ per-state databases and one match record database. Just curious if/how this class would be adapted to support that scenario?

timoballard: Yes, great question- my thinking was to borrow [this snippet](https://github.com/18F/piipan/blob/c6c5ca67640cd0e4aa89cdb0f99945bfcc387048/match/src/Piipan.Match.Orchestrator/Api.cs#L263-L266) from the `match` API and have the `Build` method here take in a (optional?) database parameter. I was thinking about including this functionality as part of this PR (note the `DatabasePlaceholder` property, but decided to hold off until it was actually needed (i.e. until `match` gets refactored to use this stuff). 

From there... either [IParticipantDao.GetParticipants](https://github.com/18F/piipan/blob/c6c5ca67640cd0e4aa89cdb0f99945bfcc387048/participants/src/Piipan.Participants/Piipan.Participants.Core/DataAccessObjects/IParticipantDao.cs#L10) will need to be updated to take in a `state` argument, or the [ParticipantService](https://github.com/18F/piipan/blob/c6c5ca67640cd0e4aa89cdb0f99945bfcc387048/participants/src/Piipan.Participants/Piipan.Participants.Core/Services/ParticipantService.cs#L14) will need to have _N_ participant DAOs, one for each state. I would lean toward the former, but I can see arguments for either approach

kingcomma: 👍 Makes sense — appreciate the explanation.

kingcomma (approved):
👍 Looks good to me, though also happy to wait for review from someone with more exposure to the ETL sub-system than me.

ryanhofdotgov (approved):
Nice! LGTM – but no need to wait to merge once you got your first approval 👍

PR #1721, opened 2021-09-23
Create match records database in IaC

kingcomma:
- Update `db_create_managed_role` to allow other resource groups when creating managed identities
- Make DDL re-runnable by handling case where ENUMs already exist
- Fix typo in schema

Partially addresses #1546

### DIFF
--- a/iac/create-core-databases.bash
+++ b/iac/create-core-databases.bash
@@ -9,7 +9,7 @@ set_constants () {
   SUPERUSER=$DB_ADMIN_NAME
 
   METRICS_DB_NAME=metrics
-  MATCH_RECORDS_DB_NAME=matches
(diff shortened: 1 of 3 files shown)


ryanhofdotgov (approved):
Ready to merge, but with a naming nitpick: I might have called this new database `collab`, `collaboration`, or something reflective of a larger scope since we've had that as the placeholder on the high-level diagram and it will likely quickly grow to include things other that match records/issues.

For example, points of contact info for states, possibly some per-state access control configuration, etc.

kingcomma:
@ryanhofdotgov I think that's worth addressing before merging. I just pushed a commit that uses `collaboration`.

ryanhofdotgov (approved):

PR #1723, opened 2021-09-24
Scaffold refactored match sub-system

kingcomma:
Partially addresses #1658.


.github/dependabot.yml
```diff
@@ -44,6 +44,21 @@ updates:
...
+- package-ecosystem: nuget
+  directory: "/match/src/Piipan.Match.Api"
+  schedule:
+    interval: daily
+    time: "09:00"
+- package-ecosystem: nuget
+  directory: "/match/src/Piipan.Match.Func.Api"
+  schedule:
+    interval: daily
+    time: "09:00"
```
timoballard, line 61: _issue_: I think these are missing the `Piipan.Match` directory in the middle:

`/match/src/Piipan.Match/Piipan.Match.*`

timoballard (approved):

.github/dependabot.yml
kingcomma: 😅 good catch — thanks!

PR #1737, opened 2021-09-24
Refactor Match subsystem

### LINKED ISSUE #1637: Remove references to plain text matching endpoint
Now that we've made the decision to perform de-identification within the Query Tool, references to the plain text matching endpoint can be removed.

- [x] Remove PII-based endpoint from the [OpenAPI schema](https://github.com/18F/piipan/blob/dev/match/docs/openapi/orchestrator/index.yaml#L14-L15)
- [x] Remove stubbed-in PII-based resources ([example](https://github.com/18F/piipan/blob/dev/match/src/Piipan.Match.Orchestrator/Api.cs#L100-L116)) from `Piipan.Match.Orchestrator`.

### LINKED ISSUE #1658: Refactor match subsystem
Refactor the `match` subsystem to align with the [standard subsystem architecture](https://github.com/18F/piipan/blob/dev/docs/adr/0018-standardize-subsystem-software-architecture.md).

timoballard:
Closes #1658 
Closes #1637 

Refactors the Match subsystem to align with our [standard subsystem architecture](https://github.com/18F/piipan/blob/dev/docs/adr/0018-standardize-subsystem-software-architecture.md).


match/src/Piipan.Match/Piipan.Match.Func.Api/Startup.cs
```diff
@@ -0,0 +1,55 @@
...
+            builder.Services.AddTransient<IMatchApi, MatchResolver>();
+
+            builder.Services.AddSingleton<ITokenProvider>((s) =>
+            {
+                if (configuration?["DEVELOPMENT"] == "true")
+                {
+                    return new CliTokenProvider();
+                }
+                return new EasyAuthTokenProvider();
+            });
```
kingcomma, line 41: *Suggestion (non-blocking):* This _was_ [used by the shared API client](https://github.com/18F/piipan/blob/v0.6/match/src/Piipan.Match.Orchestrator/Startup.cs#L27-L30) to generate OAuth tokens for accessing the per-state APIs when they still existed. Once those were ripped out, this token provider was used to [generate the token (password) for connecting to the database](https://github.com/18F/piipan/blob/v0.8/match/src/Piipan.Match.Orchestrator/Api.cs#L257-L261) — not the intended use, but it worked well enough as a service that could be injected and abstract away the `Azure.Identity` details. Now, I think it's not used at all and could be dropped from this startup.

That said, we could probably update the `*PgConnectionFactory` implementations to use and finally cut ties with `Microsoft.Azure.Services.AppAuthentication` ([which MS no longer recommends](https://docs.microsoft.com/en-us/dotnet/api/overview/azure/service-to-service-authentication) — which we learned after we started using it / saw it recommended in other docs 🙃 ).

match/tests/Piipan.Match.Core.Tests/Parsers/OrchMatchRequestParserTests.cs
```diff
@@ -0,0 +1,113 @@
...
+            var parser = new OrchMatchRequestParser(validator, logger);
+
+            // Act / Assert
+            await Assert.ThrowsAsync<StreamParserException>(() => parser.Parse(BuildStream(s)));
+        }
+
+        [Theory]
+        [InlineData(@"{ data: [{ lds_hash: 'abc' }]}", 1)] // invalid hash, but valid request
+        [InlineData(@"{ data: [{ lds_hash: '' }]}", 1)] // empty hash, but valid request
+        [InlineData(@"{'data':[
```
kingcomma, line 51:
```suggestion
        // farrington,1931-10-13,000-12-3456
        [InlineData(@"{'data':[

```
*Nitpick*: A small thing, but just trying to add comments where we use actual hashes.

kingcomma (approved):
🏅 🏅 🏅 

This took a lot of work, and it paid off nicely! Thanks for being the one who took it on.

Just a couple notes, but nothing that should hold up merging.

match/src/Piipan.Match/Piipan.Match.Func.Api/Startup.cs
timoballard: awesome! I'll add a separate issue for updating the shared `*PgConnectionFactorys`

PR #1767, opened 2021-09-27
Refine De-Identification Library

### LINKED ISSUE #1217: Normalize inputs to NAC query website
As a State agency worker using the NAC website, I want the NAC to be able to apply normalization rules to the PII that I enter into NAC queries so that I do not have to memorize those complicated rules when performing data entry.

This would also be useful for providing states with sample code that implements the normalization rules.

AC:
- [ ] When entering queries through the NAC website, each rule from #1184 is applied before hashing the participant record for comparison.

echappen:
Closes #1217  

### Notes
- Every class throws ArgumentException. I was thinking of having each normalizer throw its own exception that inherits from ArgumentException. Would that be more desirable?
- Per our Slack discussion, I added a list of SSN's suitable for testing, with a small doc describing where they come from and what they should be used for.


shared/src/Piipan.Shared/DobNormalizer.cs
```diff
@@ -12,13 +13,28 @@ public class DobNormalizer : IDobNormalizer
         /// Public entrypoint for class.
         /// </summary>
         /// <param name="dob">date of birth of individual</param>
+
```
ryanhofdotgov, line 16: It's not in this diff, but the method documentation should specify what the expected format for the input param is, particularly since it is a string type and note a Date.

For these public method docblocks, a good guideline is for the caller to be able to use the method without looking at the source code.

shared/src/Piipan.Shared/DobNormalizer.cs
```diff
@@ -12,13 +13,28 @@ public class DobNormalizer : IDobNormalizer
...
         {
             Regex iso8601rgx = new Regex(@"^\d{4}-\d{2}-\d{2}$");
             if (!iso8601rgx.IsMatch(dob))
             {
                 throw new ArgumentException("dates must be in ISO 8601 format using a 4-digit year, a zero-padded month, and zero-padded day");
             }
+            try
+            {
+                // Dates are gregorian by default
+                DateTime.Parse(dob, new CultureInfo("en-US"));
```
ryanhofdotgov, line 28: nitpick: Since the dates are ISO 8601, they shouldn't be culture (US English) specific. I think [this should work](https://stackoverflow.com/a/5106814).

shared/src/Piipan.Shared/NameNormalizer.cs
```diff
@@ -14,23 +14,24 @@ public class NameNormalizer : INameNormalizer
         /// <param name="lname">last name of individual</param>
         public string Run(string lname)
         {
-            // loud failure of any non-ascii characters
```
ryanhofdotgov, line 17: Similar to the other method comment – should document any restrictions on the input; i.e., expecting only ASCII characters.

I think this is a good check to have at the top and wouldn't remove it – it reinforces that this method only operates on ASCII strings. It is consistent on not operating on input that has errors, exiting immediately when you aren't operating from a known state.

shared/src/Piipan.Shared/NameNormalizer.cs
```diff
@@ -14,23 +14,24 @@ public class NameNormalizer : INameNormalizer
...
             // Replace multiple spaces with one space
             result = Regex.Replace(result, @"\s{2,}", " ");
             // Trim any spaces at the start and end of the last name
             char[] charsToTrim = { ' ' };
             result = result.Trim(charsToTrim);
+            // Remove suffixes: roman numerals i-ix, variations of junior/senior
+            result = Regex.Replace(result, @"(\s(?:ix|iv|v?i{0,3}|junior|jr\.|jr|jnr|senior|sr\.|sr|snr)$)", "");
+            // Throw exception for any character that is not an
+            // ASCII space(0x20) or not in range[a - z] (0x61 - 0x7a)
+            Match m = Regex.Match(result, @"[^\x61-\x7a|\x20]", RegexOptions.IgnoreCase);
```
ryanhofdotgov, line 30: Can you encode the regular expression here without using hexadecimals?

I wouldn't throw an exception since we are normalizing, not just validating. We should be removing anything that isn't an space or a-z. Like "O'Grady" is acceptable input and output should be "ogrady", not a thrown exception.

shared/src/Piipan.Shared/SsnNormalizer.cs
```diff
@@ -17,6 +17,28 @@ public string Run(string ssn)
             {
                 throw new ArgumentException("social security number must have a 3-digit area number, a 2-digit group number, and a 4-digit serial number, in this order, all separated by a hyphen");
             }
+            string[] numberGroups = ssn.Split("-");
+            // Area numbers 000, 666, and 900-999 are invalid
+            string area = numberGroups[0];
+            int areaInt = Int32.Parse(area);
+            if (String.Equals(area,"000") |
+                String.Equals(area,"666") |
+                (areaInt >= 900 && areaInt <= 999))
```
ryanhofdotgov, line 26: nitpick: since you've already confirmed it is a 3 digit integer, you can skip the integer parsing and just check if String.StartsWith("9")

shared/tests/Piipan.Shared.Tests/NameNormalizerTests.cs
```diff
@@ -7,106 +7,89 @@ namespace Piipan.Shared.Tests
...
-            public void removeAsciiNotInRange(string name)
-            {
-                string result = _nameNormalizer.Run(name);
-                Regex rgx = new Regex(@"[^a-z|\x20]");
-                Assert.DoesNotMatch(rgx, result);
-            }
+        // Throw exception for any character that is not an
+        // ASCII space (0x20) or not in range [a-z] (0x61-0x7a)
+        [Theory ]
+        [InlineData("foobar.")]
```
ryanhofdotgov, line 75: These first three cases should be normalized rather than cause an exception

shared/src/Piipan.Shared/DobNormalizer.cs
echappen: The "s" format would crash without including time, offset etc, so I made a custom format to fit our Year-Month-Day standard.

shared/src/Piipan.Shared/NameNormalizer.cs
echappen: That makes sense to me for things like apostrophes, but less so for things like accents over letters—we'd allow "garcía" and it would become "garca"?

echappen: Oh I suppose garcía would be caught at the very beginning once we put the non-ascii check back in.

echappen: As for the regex, @kingcomma [pointed out to me](https://github.com/18F/piipan/pull/1593#discussion_r708710973) that we might want to distinguish between ascii and non-ascii whitespaces by using the hexadecimal. What about something like `[^a-z|\x20]`?

echappen: ^ That's caught in the prior check too. Ignore me!

ryanhofdotgov: Non-ascii whitespaces should be now be covered by the input check at the top/method contract that lays out expectation of ASCII only input?

timoballard:
This looks great @echappen!

RE: your question about custom exceptions: I think that would make sense if it's reasonable to expect that consumers of this library would need/want to take different actions depending on the exception type (e.g. `NameNormalizationException` vs. `SsnNormalizationException`). If, on the other hand, the consumer is likely to treat them all the same and just log/return the exception message to the user, then I think having everything throw `ArgumentException` with a detailed message makes sense.

This isn't specific to this PR, but I wonder if this is a good opportunity to organize the `Piipan.Shared` library a bit so that classes and interfaces are in directories that correspond to their namespace. The library has grown quite a bit over the last couple of months and it's a little hard to visually parse with no directory structure, and when adding a new class it's not obvious what namespace it should fall under without being able to see what already exists. Thoughts?

timoballard (approved):
🚀

echappen:
@timoballard RE: custom exceptions, that was my thinking too. That choice will probably become clear once consumers start using the API in earnest, so I'm thinking hold off until then.

PR #1771, opened 2021-09-28
Persistent match record implementation

### LINKED ISSUE #1546: Create Persistent Match Records

kingcomma:
- Service for adding records
- DAO for accessing database
- Builder for creating a `MatchRecordDbo` from match event data
- Service for generating match IDs

Partially addresses #1546. Final work will be implemented once this PR is reconciled with #1737. Rough plan at this point is for the orchestrator to loop through all matches and call `MatchRecordService.AddRecord()` for each.

Also, this work only addresses the need to create records for now. Retrieving/reconciling records will be addressed with #1549.


match/src/Piipan.Match/Piipan.Match.Core/Services/MatchRecordService.cs
```diff
@@ -0,0 +1,75 @@
...
+            // Match IDs are generated randomly from a large pool, but collision
+            // is still possible. Retry failed inserts a limited number of times.
+            var retries = 0;
+            while (retries < InsertRetries)
+            {
+                try
+                {
+                    return await _matchRecordDao.AddRecord(matchRecordDbo);
+                }
+                catch (PostgresException ex)
```
timoballard, line 52: _nitpick_:  can we have the DAO catch the `PostgresException` and rethrow it either as something more generic (e.g. `InvalidOperationException`) or as something custom (e.g. `DuplicateMatchRecordException`)? That way the DAO isn't leaking implementation details (in this case, that we're using PSQL) into the service layer

match/src/Piipan.Match/Piipan.Match.Core/Models/ActiveMatchRecordBuilder.cs
```diff
@@ -0,0 +1,48 @@
+using System.Text.Json;
+using Piipan.Match.Orchestrator;
+
+namespace Piipan.Match.Core.Models
```
timoballard, line 4: _suggestion (non-blocking)_: I think a separate `Builders` namespace/directory would make sense here

match/src/Piipan.Match/Piipan.Match.Core/Models/IActiveMatchRecordBuilder.cs
```diff
@@ -0,0 +1,13 @@
...
+using Piipan.Match.Orchestrator;
+
+namespace Piipan.Match.Core.Models
+{
+    public interface IActiveMatchRecordBuilder
+    {
+        IActiveMatchRecordBuilder SetMatch(RequestPerson input, ParticipantRecord match);
+        IActiveMatchRecordBuilder SetStates(string stateA, string stateB);
+
+        MatchRecordDbo GetRecord();
```
timoballard, line 11: _question_: can this return an `IMatchRecord`?

match/tests/Piipan.Match.Core.Tests/Models/MatchRecordDboTests.cs
```diff
@@ -0,0 +1,365 @@
...
+            };
+
+
+            // Act / Assert
+            Assert.True(record.Equals(recordMatch));
+            Assert.Equal(record.GetHashCode(), recordMatch.GetHashCode());
+        }
+
+        [Fact]
+        public void Equals_Match()
```
timoballard, line 336: _nitpick_: I think if this test also showed that equality doesn't depend on the `createdAt` property, the test coverage for this PR might get to 100% (from 99 😁)

timoballard (commented):
Took a quick first pass - looks great! Left a few comments

kingcomma:
@timoballard Really appreciate the review and suggestions — I think they made for some solid improvements! I pushed a commit that should address everything, and I also added in some documentation on the public classes/methods that I had intended to include before.

timoballard (approved):
:amaze:

PR #1794, opened 2021-10-04
Make directories for every Shared project namespace

### LINKED ISSUE #1774: Organize Shared project into namespace directories
I wonder if this is a good opportunity to organize the `Piipan.Shared` library a bit so that classes and interfaces are in directories that correspond to their namespace. The library has grown quite a bit over the last couple of months and it's a little hard to visually parse with no directory structure, and when adding a new class it's not obvious what namespace it should fall under without being able to see what already exists. Thoughts?

_Originally posted by @timoballard in https://github.com/18F/piipan/issues/1767#issuecomment-931387300_

echappen:
Closes #1774


timoballard:
@echappen - yeah, maybe just `Piipan.Shared.Database`?

timoballard (approved):

Issue #1828, opened 2021-10-05
Match API should use serialized `Participant` models for returning matches 


kingcomma:
The [match service](https://github.com/18F/piipan/blob/dev/match/src/Piipan.Match/Piipan.Match.Core/Services/MatchService.cs#L68-L69) currently returns matches as `IEnumerable<IParticipant>` rather than `IEnumerable<Piipan.Match.Core.Models.Participant>`. This results in issues like JSON properties in CamelCase instead of snake_case and date formatting not being applied.

To resolve this issue, project each `IParticipant` instance to a `Piipan.Match.Core.Models.Participant` instance before returning matches from service. E.g.,

```
var matches = (await states
    .SelectManyAsync(state => _participantApi.GetParticipants(state, person.LdsHash)))
    .Select(p => new Participant { ... });
```

PR #1834, opened 2021-10-06
Fix: convert IParticipant to Match.Core.Models.Participant

echappen:
Json serialization for the Match API is specified on the Match Core models, not through the Participant API

For example, without this serialization, the property `BenefitsEndDate` is serialized as `"benefitsEndDate"` instead of `"benefits_end_month"` as is expected from the Match API.

### Notes
- Converting to a `Participant` this way marries the Match Service to the schema of `Participant`. The only ways I found to succinctly convert one object to a similar object without relying on schema was either through [Json serialization/deserialization](https://stackoverflow.com/a/38519023) or through [casting](https://stackoverflow.com/a/3672766) a subclass object to a superclass object.  
- Testing end-to-end serialization of the API data felt most appropriate to do in the Match API Integration Tests, since other unit tests mocked participants as the intended model. I wonder if it would be worth adding more tests to assert that what's returned from the API matches the OpenApi schema more generally.


kingcomma:
Closes #1828

echappen:
@timoballard Thanks for explaining—I was thinking more about the likelihood of forgetting to add properties to this mapping in `MatchService` as the Participant Model or Interface is changed. Maybe that's just something we live with in order to get the benefits of flexibility.

match/src/Piipan.Match/Piipan.Match.Core/Models/Participant.cs
```diff
@@ -32,6 +33,18 @@ public class Participant : IParticipant
...
 
+        public Participant() {}
+
+        public Participant(IParticipant p) {
+            LdsHash = p.LdsHash;
+            State = p.State;
+            CaseId = p.CaseId;
+            ParticipantId = p.ParticipantId;
+            BenefitsEndDate = p.BenefitsEndDate;
+            RecentBenefitMonths = p.RecentBenefitMonths.ToList();
```
kingcomma, line 44: *Suggestion:* At the risk of this PR spiraling out scope, this line made me realize that while  the `[JsonConverter]` attribute has been placed on a property with type `IEnumerable<DateTime>` it will throw an exception unless it is operating on `List<DateTime>`.

Do you think this is worth addressing now or something that can be handled later in a follow-on ticket?  I'm thinking the work entails updating the [converter](https://github.com/18F/piipan/blob/dev/match/src/Piipan.Match/Piipan.Match.Core/Serializers/JsonConverters.cs#L132) so it is designed to operate on an `IEnumerable<>` instead of `List<>`. Something like:

```
var dateList = ((IEnumerable<DateTime>)value).ToList();
```

Apologies, I could have caught this in an early review on a different PR.

echappen: Thanks for pointing this out. I'll see if I can resolve it in a short time box and if not, I'll defer it to another card in the interest of getting a fix out to the Duplication Participation API.

echappen: I have a fix but I'm not sure if it's the right fix. I'll merge this and create another PR to review.

Issue #1796, opened 2021-10-04
Update IaC documentation/scripts to easy onboarding 


edwintorres:
IaC documentation/scripts need to be updated to remove know friction on onboarding people.

- [x] Add to the documentation the version of the tools used and are known to work, given that most recent versions are not working.
- [ ] Update create-resource script, so it is more robust with errors publishing func apps.
- [x] Add to the additional documentation step needed after create-resource ran but before testing
- [x] Add to the documentation suggested test to verify infrastructure is running after create-resource script ran

PR #1835, opened 2021-10-06
Eth/fix iac create resource2

edwintorres:
Updated documentation and IaC scripts after findings on onboard according  to task #1796


ryanhofdotgov:
@edwintorres It looks like this PR might have more changes than you intended? Just expecting diffs for the associated Markdown and IaC bash scripts.

edwintorres:
Yes @ryanhofdotgov, I did have some problems with the rebase, ended doing a merge. I'm doing the change again.

PR #1836, opened 2021-10-06
Display input-specific errors in query tool

### LINKED ISSUE #1544: Query Tool should accept only printable ASCII characters as input
As part of conforming with the [PPRL normalization process](https://github.com/18F/piipan/blob/dev/docs/pprl.md#1-validation-and-normalization), the Query Tool should accept only printable ASCII characters as input (e.g., `[ -~]`).

Related to #1540.

AC:
- [ ] Web NAC search form only allows ASCII input
- [ ] Form displays errors to the user when inputs are rejected because they are outside the ASCII range.

echappen:
Closes #1544 

### Screenshots

Old error:

Screen Shot 2021-10-06 at 11 23 20 AM

New error:

Screen Shot 2021-10-06 at 11 22 42 AM


juliaklindpaintner:
@echappen A couple thoughts/questions: 

- I think our existing error messages use punctuation at the end. Let's add periods to these two as well.  
- I'm assuming that with the generic error, there isn't anymore specificity you can provide? What do you think about being able to provide the user some idea about what to do next? e.g., "Please try again." or "Please refresh the page and try again." ? 
- For the ASCII one, I think we should do something more plain language. What about "Name fields may only contain ASCII characters: Letters A-Z, numbers 0-9, and the special characters XYZ." ? Or if we think that it's mostly going to be about accents, we could even add: "Please make sure the name fields don't contain any special characters such as letters with accent marks."

query-tool/src/Piipan.QueryTool/Pages/Index.cshtml.cs
```diff
@@ -55,6 +55,18 @@ public async Task<IActionResult> OnPostAsync()
...
+                    _logger.LogError(ex, ex.Message);
+                    if (ex.Message.ToLower().Contains("only ascii"))
+                    {
+                        RequestError = $"{ex.Message}.";
+                    }
+                    else
+                    {
+                        RequestError = $"{ex.Message}. Please try again.";
+                    }
+                }
```
kingcomma, line 69: *Suggestion:* In addition to catching the exception thrown by the normalizer, I think there may be an advantage to implementing this constraint via data annotations on the `PiiRecord` model (like we do with SSN and DoB). That would enforce the same constrain on the First and Middle fields which wouldn't be caught by the current name normalizer.

echappen: I couldn't find any docs on how to get the matching character into the error message through Data Annotations. I'm syncing with @juliaklindpaintner shortly to figure out how specific we'd like the messages to be, and will explain benefits/tradeoffs of each approach.

kingcomma: If we go the data annotations route, perhaps a custom validator along [these lines](https://stackoverflow.com/a/3413875) would work.

juliaklindpaintner: Just to expand upon what Eleni said, even though it would provide a more consistent user experience in the event that someone had special characters in multiple name fields, it would create unnecessary burden/potential confusion for users who have entered a special character only in the first or middle name. Those would then be "fake" errors that we don't actually need the user to fix. So part of our logic for leaving those error messages out is also to avoid this scenario.

juliaklindpaintner (approved):
LGTM! Thanks so much :)

PR #1841, opened 2021-10-06
Add code to retry func publishing on create resources

edwintorres:
Changes to the create-resources scripts


iac/azfunc-publish.bash
```diff
@@ -0,0 +1,49 @@
...
+# availble yet, or other kind of error. The wrapper function will try 
+# to publish the functionapp up to an MAX_TRIES of times. 
+#
+# app_name - name of the function application
+# directory - path to the application project
+#
+# usage:   source ./azfunc-publish.bash
+#          azfunc_publish <app_name> <directory>
+# example: azfunc_publish "${func_app}" "../etl/src/Piipan.Etl/Piipan.Etl.Func.BulkUpload"
+#
```
kingcomma, line 15: *Suggestion:* Rather than creating a new bash script, this function could go in the existing `iac-common` script which is already sourced by our various bash scripts. That way developers don't have to worry about whether or not this script has already been sourced before using `azfunc_publish`.

iac/azfunc-publish.bash
```diff
@@ -0,0 +1,49 @@
...
+  ERR=0 # or some non zero error number you want
+  MAX_TRIES=6
+
+  echo "Publishing ${app_name} function app"
+  pushd "$directory" || exit
+    for (( i=1; i<=MAX_TRIES; i++ ))
+      do
+        ERR=0
+        echo "Waiting to publish function app"
+        sleep $(( i * 30 ))
```
kingcomma, line 33: *Question:* Because this script will retry after an initial failure, would it be more efficient to try the first attempt immediately rather than waiting 30 seconds?

I'm thinking of the cases when IaC is run against existing resources and waiting isn't necessary. In the case of the per-state ETL function apps, this wait time would compound noticeably.

iac/azfunc-publish.bash
```diff
@@ -0,0 +1,49 @@
...
+
+        echo "func azure functionapp publish ${app_name} --dotnet" 
+        func azure functionapp publish "$app_name" --dotnet || ERR=1
```
kingcomma, line 36: *Suggestion (non-blocking):* A future improvement could be to make this retry function agnostic of the command that is actually being run. That would allow for any command to be retried, for example:

```
retry func azure functionapp publish "$app_name" --dotnet
retry func azure functionapp publish "$app_name" --dotnet --force
retry func azure functionapp publish "$app_name" --dotnet --slot "dev"
retry other_custom_function
```

iac/create-metrics-resources.bash
```diff
@@ -132,10 +132,8 @@ main () {
     --output none
 
   # publish the function app
-  echo "Publishing function app $METRICS_COLLECT_APP_NAME"
-  pushd ../metrics/src/Piipan.Metrics/$COLLECT_APP_FILEPATH
-    func azure functionapp publish "$METRICS_COLLECT_APP_NAME" --dotnet
-  popd
+  source ./azfunc-publish.bash
+  azfunc_publish "${METRICS_COLLECT_APP_NAME}" "../metrics/src/Piipan.Metrics/$COLLECT_APP_FILEPATH"
```
kingcomma, line 136: *Nitpick:* Looks like there is a `sleep` command above this on line 122 that could be removed.

iac/azfunc-publish.bash
```diff
@@ -0,0 +1,49 @@
...
+        fi
+
+      done
+    if [ $ERR -eq 1 ];then
+      echo "Too many non-sucessful tries"
+      exit $ERR
+    fi
+  popd || exit
+
+}
```
kingcomma, line 49: Looks like this file can be deleted now.

iac/iac-common.bash
```diff
@@ -202,4 +202,49 @@ private_dns_zone () {
...
+#
+# mycommand - command to be run
+# directory - path where tje mycommand should be run
+# max_tries - max number of try, default value 6
+#
+# usage:   try_command <mycommand> <directory> <max_tries>
+#
+try_run () {
+  mycommand=$1
+  directory="${2:-${"./"}}"
```
kingcomma, line 220: *Issue:* Looks like this will throw a `bad substitution` error in bash if `$2` is not set. Instead, I think `directory="${2:-./}"` might be the solution.

That said, I'm wondering if it makes sense to remove the `pushd/popd` functionality from this script entirely and instead have it focus on just one thing: retrying failed commands. Then, the relevant lines in `create-resources` would be something like:

```
pushd ../match/src/Piipan.Match/Piipan.Match.Func.Api
  try_run "func azure functionapp publish ${ORCHESTRATOR_FUNC_APP_NAME} --dotnet"
  # or:
  try_run "func azure functionapp publish ${ORCHESTRATOR_FUNC_APP_NAME} --dotnet" 5
popd
```

What do you think?

edwintorres: I changed the parameters order and fix the problem with the directory default value

https://github.com/18F/piipan/blob/e5ec657fcff9c8fa2be5cb8a7fd1cbb6d9eca7af/iac/iac-common.bash#L219-L221


This is how it looks like the call now.

**Try to run** this code, **X times**, and **try to run it on** this directory.

https://github.com/18F/piipan/blob/e5ec657fcff9c8fa2be5cb8a7fd1cbb6d9eca7af/iac/create-resources.bash#L322-L323

iac/iac-common.bash
```diff
@@ -202,4 +202,50 @@ private_dns_zone () {
...
+#
+# The function help with the robusness of the IaC code. 
+# In ocassions the original when run a command it can fail, because any kind of error. 
+# The wrapper function will try run the command to a max_tries of times. 
+#
+# mycommand - command to be run
+# max_tries - max number of try, default value 3
+# directory - path where tje mycommand should be run
+#
+# usage:   try_command <mycommand> <max_tries> <directory>
```
kingcomma, line 216:
```suggestion
# usage:   try_run <mycommand> <max_tries> <directory> 
```

iac/iac-common.bash
```diff
@@ -202,4 +202,50 @@ private_dns_zone () {
 
   echo $base
 }
+
+# try_command()
```
kingcomma, line 206:
```suggestion
# try_run()
```

kingcomma (approved):
👍

Just a couple copy edit suggestions, but this looks good to me!

Issue #1799, opened 2021-10-04
Refactor Dashboard


timoballard:
Refactor the Dashboard application to make use of the metrics subsystem API and data models.

- [ ] Create `Piipan.Metrics.Client` library in the metrics subsystem, which will provide [a standard HTTP client implementation](https://github.com/18F/piipan/blob/dev/docs/adr/0018-standardize-subsystem-software-architecture.md#optional-namespacesubsystemclient) of `IParticipantUploadApi`.
- [ ] Update the Dashboard application to use the newly created ParticipantUpload API client for its metrics requests
- [ ] Update the Dashbaord to reference the Metrics request and response models in `Piipan.Metrics.Api`
- [ ] Remove Metrics request and response models in `dashboard` subsystem

Issue #1798, opened 2021-10-04
Refactor Query Tool 


timoballard:
Refactor the Query Tool application to make use of the match subsystem API and data models.

- [ ] Create `Piipan.Match.Client` library in the match subsystem, which will provide [a standard HTTP client implementation](https://github.com/18F/piipan/blob/dev/docs/adr/0018-standardize-subsystem-software-architecture.md#optional-namespacesubsystemclient) of `IMatchApi`.
- [ ] Update the Query Tool application to use the newly created Match API client for its match requests
- [ ] Update the Query Tool to reference the Match request and response models in `Piipan.Match.Api`
- [ ] Remove Match request and response models in `query-tool` subsystem

PR #1921, opened 2021-10-15
Match Client library

timoballard:
Relates to #1798

- Creates a new library (`Piipan.Match.Client`) to provide an [HTTP client-side implementation](https://github.com/18F/piipan/blob/dev/docs/adr/0018-standardize-subsystem-software-architecture.md#optional-namespacesubsystemclient) of [`IMatchApi`](https://github.com/18F/piipan/blob/dev/match/src/Piipan.Match/Piipan.Match.Api/IMatchApi.cs). 
- Refactors `IAuthorizedApiClient` (and `AuthorizedJsonApiClient`) so that the `PostAsync` and `GetAsync` methods are generic and object serialization/deserialization is internalized.

## Notes
- The [original `ITokenProvider` and `IAuthorizedApiClient` interfaces and implementations](https://github.com/18F/piipan/blob/dev/shared/src/Piipan.Shared/Authentication/Authentication.cs) remain in place for the time being, but will be removed when all usages are refactored to use the [new versions](https://github.com/18F/piipan/tree/dev/shared/src/Piipan.Shared/Http) (i.e., when #1798 and #1799 are complete)


match/src/Piipan.Match/Piipan.Match.Client/MatchClient.cs
```diff
@@ -0,0 +1,30 @@
+﻿using System.Collections.Generic;
+using System.Threading.Tasks;
+using Piipan.Match.Api;
+using Piipan.Match.Api.Models;
+using Piipan.Shared.Http;
+
+namespace Piipan.Match.Client
+{
+    public class MatchClient : IMatchApi
```
echappen, line 9: _Request_: Could you add annotations for public classes

PR #2038, opened 2021-11-03
docker iac documentation  Closes #1797

### LINKED ISSUE #1797: Create Azure VM with needed software
Given that not all tools (azure-cli, azure core tools) works with the IaC, there is the opportunity to create a optional script to deploy a VM with the needed software is case is needed.

edwintorres:
Added docker and docker-compose to run IAC code with the version of the pre-requisite software that we know works well.

### DIFF
--- a/iac/Dockerfile
+++ b/iac/Dockerfile
@@ -1,0 +1,20 @@
+FROM mcr.microsoft.com/dotnet/sdk:3.1-focal AS build-env
+#Copy IAC scripts to prepare the deployment enviroment
+COPY ./iac/docker/* ./iac/
+COPY ./iac/install-extensions.bash ./iac/
+COPY ./tools/common.bash ./tools/
+#Set to an working Azure CLI version
+ENV AZURE_CLI_VERSION=2.25.0-1
+#Run IAC scripts to prepare the deployment enviroment
+RUN ./iac/install-azure-cli.bash
+RUN ./iac/install-azure-core-tool.bash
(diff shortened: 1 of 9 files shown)


echappen (approved):
🙇 🚀

iac/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+RUN ./iac/install-psql.bash
+RUN ./iac/install-extensions.bash
+# Configure az cli
+# RUN az config set extension.use_dynamic_install=yes_without_prompt
+RUN az cloud set --name AzureCloud
+#Copy piipan and set the permision to execute the code
+COPY . ./piipan
+RUN chmod -R ug+rwx /piipan
+#Start the deployment with the set enviroment "tts/dev"
+CMD ["/piipan/iac/docker/start.bash", "tts/eth"]
```
kingcomma, line 20: Looks like this is referencing a local env file instead of `tts/dev`.

iac/docker/start.bash
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
```
kingcomma, line 1:
```suggestion
#!/usr/bin/env bash
```

iac/Dockerfile
edwintorres: good catch

edwintorres:
Added docker to run iac code with the software prerequisite working version

PR #2251, opened 2021-11-09
Update minor version to dotnet dependencies

edwintorres:
Update minor version to dotnet dependencies on ASPNetCore and Extensions to all projects


kingcomma (approved):
👍 Thanks for pulling these all into one PR!

PR #2253, opened 2021-11-09
Specify which suffixes to remove in PPRL docs

### LINKED ISSUE #2285: Specify which suffixes to remove in PPRL docs
To improve consistency between state [PPRL](https://github.com/18F/piipan/blob/dev/docs/pprl.md#1-validation-and-normalization) implementations, we should be more specific about which Suffixes should be removed in last names. We can use the [set](https://github.com/18F/piipan/blob/dev/shared/src/Piipan.Shared/Deidentification/NameNormalizer.cs#L34) used in our own Normalization library.

echappen:
Closes #2285


docs/pprl.md
```diff
@@ -23,7 +23,10 @@ Finally, in order for exact matching against de-identified PII to be effective,
 Participant's Last name must be normalized and validated in accordance with these **ordered** rules. These transformations assume ASCII-encoded input.
 
 1. Convert to lower case
-1. Remove any suffixes (e.g.; `junior`, `jnr`, `jr`, `jr.`, `iii`, etc.)
+1. Remove suffixes:
+    1. Roman numerals I-IX
```
kingcomma, line 27:
```suggestion
    1. Roman numerals i-ix
```
It's a little unexpected, but since the name will be lowercase at this point I'm wondering if it makes sense to use the lowercase numerals here?

kingcomma (approved):
👍 Thanks for doing this so quickly!

One more chore/request: could you create an issue, associated it with this PR, and put it in this sprint's project?

echappen:
Done!

Issue #2023, opened 2021-11-01
Placeholder for security issues


cjhskippy:
- [x] Cookie Not Marked as Secure
- [x] Missing X-Frame-Options Header - Dashboard
- [ ] ~~Weak Ciphers Enabled~~ (now issue #2317)
- [x] [Possible] Phishing by Navigating Browser Tabs

PR #2255, opened 2021-11-09
Force antiforgery cookies to be https-only

echappen:
Addresses "Cookie Not Marked as Secure" item in #2023 

Analysis from ISO stated "The .AspNetCore.Antiforgery.9fXoN5jHCXs cookie has been marked as httponly with no secure flag" on the dashboard and query tool apps.

To solve this, we force Antiforgery to always use secure cookies at application startup.


dashboard/src/Piipan.Dashboard/Startup.cs
```diff
@@ -64,6 +65,10 @@ public void ConfigureServices(IServiceCollection services)
                 var mockFile = $"{_env.ContentRootPath}/mock_user.json";
                 services.UseJsonFileToMockEasyAuth(mockFile);
             }
+
+            services.AddAntiforgery(options => {
+                options.Cookie.SecurePolicy = CookieSecurePolicy.Always;
+            });
```
edwintorres, line 71: I'm reviewing [this documentation](https://docs.microsoft.com/en-us/aspnet/core/security/anti-request-forgery?view=aspnetcore-3.1) to see if we need more options.

edwintorres (approved):

PR #2260, opened 2021-11-09
Prevent phishing by Navigating Browser Tabs

echappen:
Addresses "Phishing by Navigating Browser Tabs" in #2023 

From ISO analysis:
"Open windows with normal hrefs with the tag target="_blank" can modify window.opener.location and replace the parent webpage with something else, even on a different origin. While this vulnerability doesn't allow script execution, it does allow phishing attacks that silently replace the parent tab. If the links lack rel="noopener noreferrer" attribute, a third party site can change the URL of the source tab using window.opener.location.assign and trick the users into thinking that they’re still in a trusted page and lead them to enter their sensitive data on the malicious website... Add rel=noopener to the links to prevent pages from abusing window.opener. This ensures that the page cannot access the window.opener property in Chrome and Opera browsers."


edwintorres:
I'm assuming this is a test with only one of the URLs

edwintorres (approved):

echappen:
This was the only URL that was pinged by the security review.

PR #2284, opened 2021-11-10
Eth/update docker azcli

edwintorres:
Added option to install az-cli latest version without removing the option to install a specific version.


kingcomma (approved):
👍

PR #2311, opened 2021-11-16
Deny default network access to storage containers

### LINKED ISSUE #2258: Deny default network access for storage accounts
From CIS Policy: Restricting default network access helps to provide a new layer of security, since storage accounts accept connections from clients on any network. To limit access to selected networks, the default action must be changed.

We should allow access for:
- GSA IP range (or deprecate `test-storage-upload.bash`)
- `vnet-core-$ENV`
- `*-apim-duppartapi-$ENV` resource
- pending experimentation, possibly `Microsoft.EventGrid/topics`

timoballard:
Closes #2258 
Configures all storage accounts to disallow default network access. Per state upload storage accounts allow access from `vnet-core-$ENV/snet-apps1-$ENV` and `*-apim-duppartapi-$ENV`, while function app storage accounts only allow access from `vnet-core-$ENV/snet-apps1-$ENV`.

## Notes
- This change renders the `test-storage-upload.bash` script unusable, as storage accounts are no longer accessible from outside the vnet. So, the script and associated documentation are removed as part of this PR.

image


iac/configure-storage-firewalls.bash
```diff
@@ -0,0 +1,90 @@
...
+      --output tsv)
+
+  while IFS=, read -r abbr name ; do
+    abbr=$(echo "$abbr" | tr '[:upper:]' '[:lower:]')
+
+    echo "Configuring storage firewalls something for $name ($abbr)"
+
+    blob_st_name="${PREFIX}st${abbr}upload${ENV}"
+
+    disallow_default "$blob_st_name" "$RESOURCE_GROUP"
```
kingcomma, line 76: *Question:* What's the difference between the CLI command this ends up calling and the [similar looking configuration in the ARM template](https://github.com/18F/piipan/blob/6eef0fc0919026b23eb724c068e0e80398e614da/iac/arm-templates/blob-storage.json#L35-L37)?

iac/configure-storage-firewalls.bash
```diff
@@ -0,0 +1,90 @@
...
+    allow_snet "$blob_st_name" "$RESOURCE_GROUP" "$VNET_NAME" "$FUNC_SUBNET_NAME"
+    allow_apim "$blob_st_name" "$RESOURCE_GROUP" "$APIM_ID" "$tenant_id"
+
+    func_st_name=${PREFIX}st${abbr}etl${ENV}
+
+    disallow_default "$func_st_name" "$RESOURCE_GROUP"
+    allow_snet "$func_st_name" "$RESOURCE_GROUP" "$VNET_NAME" "$FUNC_SUBNET_NAME"
```
kingcomma, line 83: Ditto the questions above.

iac/configure-storage-firewalls.bash
```diff
@@ -0,0 +1,90 @@
...
+
+  while IFS=, read -r abbr name ; do
+    abbr=$(echo "$abbr" | tr '[:upper:]' '[:lower:]')
+
+    echo "Configuring storage firewalls something for $name ($abbr)"
+
+    blob_st_name="${PREFIX}st${abbr}upload${ENV}"
+
+    disallow_default "$blob_st_name" "$RESOURCE_GROUP"
+    allow_snet "$blob_st_name" "$RESOURCE_GROUP" "$VNET_NAME" "$FUNC_SUBNET_NAME"
```
kingcomma, line 77: *Question:* Is it possible to enable this same network rule using the existing blob storage ARM template? If so, what do you think about that approach? I lean towards it having the benefit of moving configuration as close to the resource as possible and limiting the number of places our configuration is spread around.

iac/configure-storage-firewalls.bash
```diff
@@ -0,0 +1,90 @@
...
+    allow_apim "$blob_st_name" "$RESOURCE_GROUP" "$APIM_ID" "$tenant_id"
```
kingcomma, line 78: *Thought:* If the above configuration is moved to an ARM template, this would be the only command left I think. In that case, maybe it moves to [`create-apim.bash`](https://github.com/18F/piipan/blob/dev/iac/create-apim.bash#L142-L147) so that all the authorization for APIM is performed in one place?

kingcomma (commented):
Will be nice to check this off the list! A few thoughts/questions below on where this configuration lives.

Also, are the changes to the query tool's `package-lock.json` inadvertent?

timoballard:
@kingcomma - this is great feedback, thank you! Will push updates shortly

kingcomma (approved):
👍 Can you just revert the change in `Piipan.QueryTool/package-lock.json` before merging (assuming it's unrelated)? Small thing, but keeps the commit history clean.

PR #2320, opened 2021-11-18
Noting the node task failures in circle in the docs

echappen:
This was an action item from today's retro.


echappen:
Fixed in #2321

PR #2321, opened 2021-11-19
Separate app build from app run in CI frontend testing

echappen:
This fixes a bug where the variable build processes would prevent app from spinning up in our CI job, causing frontend tests to fail.


kingcomma (approved):
👍👍 Thanks!

PR #2330, opened 2021-11-24
Deny default access to Orchestrator function storage account

### LINKED ISSUE #2324: Deny default network access to orchestrator function storage account
Relates to #2258 

We recently made changes to deny default network access to our storage accounts, but overlooked the orchestrator function storage account -- we should update our IaC scripts to disallow default network access for this storage account as well.

timoballard:
Closes #2324 
Related to #2258 

Configures the orchestrator function app storage account to disallow default network access. Access is granted only to `vnet-core-$ENV/snet-apps1-$ENV`.

This case is slightly different from the other storage accounts, in that this account is part of the `rg-match` resource group, as opposed to `rg-core`. This seems to cause issues when attempting to publish the function app while network access restrictions are in place. To get around these issues, we temporarily disable the network firewall while executing the function app publish step.

### DIFF
--- a/iac/arm-templates/function-orch-match.json
+++ b/iac/arm-templates/function-orch-match.json
@@ -34,6 +34,12 @@
         },
         "states": {
             "type": "String"
+        },
+        "vnet": {
+            "type": "string"
+        },
+        "subnet": {
+            "type": "string"
         }
     },
     "variables": {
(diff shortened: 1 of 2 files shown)


kingcomma (approved):
👍 Thanks for catching this outlier and dealing with the new issues that popped up!

Issue #2331, opened 2021-11-24
Storage account deployments fail in fresh environments


kingcomma:
IaC deployments to fresh environments are failing when deploying the per-state storage accounts:

```
{
  "code": "ResourceDeploymentFailure",
  "message": "The response for resource had empty or invalid content."
}
```

This is likely a result of our recent networking restrictions. As [described here](https://jeroen-vdb.medium.com/azure-storage-account-in-a-virtual-network-the-response-for-resource-had-empty-or-invalid-content-a5984f5d9ff), the solution is to add the `Microsoft.Storage` [service endpoint](https://docs.microsoft.com/en-us/azure/virtual-network/virtual-network-service-endpoints-overview) to our subnets.

PR #2350, opened 2021-12-01
Release/0.94 - add service endpoint

echappen:
When deploying release 0.94 to `tts/test` I got the following error when attempting to create per-state storage accounts:

> Subnets snet-apps1-test of vnet-core-test do not have ServiceEndpoints for Microsoft.Storage resources configured. Add Microsoft.Storage to subnet's ServiceEndpoints collection before trying to ACL Microsoft.Storage resources to these subnets

Adding this service endpoint in the vnet arm template worked, although I'm not sure this change is what we intend.


kingcomma:
This issue is referenced in #2331 and should be addressed by PR #2332. It seemed like the issue only presented itself when deploying to new environments, but @echappen sounds like it also popped up when deploying to an existing environment?

echappen:
Correct, was on a deploy to an existing environment.

echappen:
Closing this in deference to #2332

PR #2411, opened 2021-12-13
add diagnostic settings to event monitoring arm template

### LINKED ISSUE #2302: Configure Event Hub diagnostic settings to send all logs to Event Hub
Should use existing approach of configuring diagnostic settings in the resource's ARM template. See [dashboard app ARM template](https://github.com/18F/piipan/blob/dev/iac/arm-templates/dashboard-app.json#L134-L170) and https://github.com/18F/piipan/blob/dev/docs/log-streaming.md as references. All log categories should be configured and sent to the existing `logs` event hub namespace.

See "Azure Event Hub: Review Logging" in security issue tracker.

echappen:
closes #2302


kingcomma (approved):
👍

Did you happen to double check that all of these log categories are available for the event hub resource in AzureGovernment? If not, it might be worth a quick check before merging (just inspecting what's available in the portal). We've been bit by inconsistencies between the clouds with respect to diagnostic settings before.

echappen:
I have not! Good call.

echappen:
@kingcomma Updated to remove logs not used in Azure Gov: `RuntimeAuditLogs` and `ApplicationMetricsLogs`

Issue #2409, opened 2021-12-13
Sequencing error found on IaC: kv-core is now dependent on the event hub


edwintorres:
Sequencing error found on IaC code. Related to #2351. Specifically, kv-core is now dependent on the event hub but is deployed before it.

PR #2352, opened 2021-12-02
Add diagnostic settings to keyvault arm template

echappen:
This will enable log streaming to our event hub resource

Closes #2304

PR #2437, opened 2021-12-14
Fix: switch creation order for key vault and event hub monitoring

echappen:
Fixes a bug introduced in #2352 

Now that the key vault relies on event hub monitoring, event hub monitoring should be created before the key vault.


kingcomma:
Fixes #2409

kingcomma (approved):
👍 Thanks!

PR #2463, opened 2021-12-15
Upgrade Npgsql packages across subsystems

### LINKED ISSUE #2286: Upgrade to latest Npgsql
In order to upgrade to the latest version of Npgsql, we will need to (in order):

- [ ] drop databases
- [ ] change any `timestamp` data columns to `timestamptz`
- [ ] upgrade Npgsql
- [ ] ensure we always use DateTime with Kind=Utc when interacting with Npgsql throughout the subsystems
- [ ] re-migrate databases

 [Npgsql upgrade details](https://www.npgsql.org/doc/release-notes/6.0.html#major-changes-to-timestamp-mapping)

These changes should be done at the same time to avoid unintended behavior with Npgsql `timestamp` mapping.

echappen:
Closes #2286

This entails converting Postgres `timestamp without timezone` fields to `timestamp with timezone` (`timestamptz`).

We often use the PG `now()` function to insert timestamps, which by default is based on the database server's timezone setting. I'm assuming the default timezone setting for Azure servers is UTC, but I haven't been able to confirm this. As an extra precaution, I added a timezone setting to our instances of `now()` to be `now() at time zone 'utc'`, although this may be excessive.

Other instances of timestamp database insertion were from `EventGridEvent.EventTime` which are confirmed to be in UTC, so there was no need to update this application code for this upgrade. 

⚠️ This change will require a database drop in order to run IAC ⚠️


metrics/tests/Piipan.Metrics.Core.IntegrationTests/ParticipantUploadDaoTests.cs
```diff
@@ -188,9 +188,11 @@ public async Task AddUpload_InsertsRecord()
 
             // Act
             await dao.AddUpload("ea", uploadedAt);
+            var latest = await dao.GetLatestUploadsByState();
 
             // Assert
             Assert.True(Has("ea", uploadedAt));
+            Assert.Equal(uploadedAt, latest.ToList()[0].UploadedAt);
```
kingcomma, line 195: *Thought (non-blocking):* This test might be slightly misleading. I think it would ideally specify `DateTimeKind.Utc` when setting `uploadedAt`.

As it stands, since we set a `DateTimeKind` for `uploadedAt`, it is set to `Unspecified`. Per the [Npgsql release notes](https://www.npgsql.org/doc/release-notes/6.0.html#detailed-notes), `Unspecified` is sent to PostgreSQL as `timestamp` (i.e., without any time zone information). When that time is selected out of PostgreSQL, it comes back as `DateTimeKind.Utc`.

The theoretical issue here is that if the tests (or, worse, applications) were run on a machine with a time zone set to something other than UTC, `uploadedAt` would represent the local time but PostgreSQL would handle it (both when reading and writing) like it's UTC.

Since the "real world" use case for the metrics collector involves `EventGridEvent.EventTime` (which as you noted is `UTC`), I don't think there are any bugs we are currently missing.

kingcomma (approved):
👍 👍 

I left a comment on one of the tests mostly for posterity in case an issue pops up in the future.

PR #2470, opened 2021-12-17
Re-sequence orchestrator networking config in IaC

### LINKED ISSUE #2410: Sequencing error found on IaC: orchestrator function app deployment
Line 314 is the orchestrator function app deployment. Looks like this might be another sequencing error we introduced. The ARM template limits traffic to the storage account for the function app to the vnet, but the function app itself isn’t added to the vnet until later on in create-resources.

https://gsa-tts.slack.com/archives/C018P1ZJPEC/p1639164800006300?thread_ts=1639148341.000300&cid=C018P1ZJPEC

kingcomma:
Apply networking restrictions after base function app and backing storage account are deployed.

Loosely follows Microsoft's [guide for restricting traffic to a function app's storage account](https://docs.microsoft.com/en-us/azure/azure-functions/configure-networking-how-to#restrict-your-storage-account-to-a-virtual-network). (But does not create multiple storage accounts and lets ARM handle file share creation/configuration).

This brings the ordering of the orchestrator deployment closer to [that of our ETL apps](https://github.com/18F/piipan/blob/dev/iac/create-resources.bash#L393-L430). One key difference remains: the ETL apps are able to deploy the storage account with the [`networkAcls` in place](https://github.com/18F/piipan/blob/dev/iac/arm-templates/function-storage.json#L37-L45). I was only able to get the orchestrator working if those rules were added as a [second step](https://github.com/18F/piipan/blob/61babde0b58f497918b7710a39dcf896992e7fa1/iac/create-resources.bash#L332-L349). I wasn't able to reconcile this difference. My best guess is Azure's resource manager performs some additional steps when deploying a storage <-> function app pair from a single ARM template, and this is different than what happens when using the AZ CLI to deploy the pair in separate steps.

Tested against a fresh `tts/mrk` environment. Deployment was successful and tests passed. Subsequent deployments and publishes to the existing environment were also successful. Environment is still available for review.

Closes #2410

### DIFF
--- a/iac/create-resources.bash
+++ b/iac/create-resources.bash
@@ -306,7 +306,9 @@ main () {
   state_abbrs=${state_abbrs:1}
 
   # Create orchestrator-level Function app using ARM template and
-  # deploy project code using functions core tools.
+  # deploy project code using functions core tools. Networking
+  # restrictions for the function app and storage account are added
+  # in a separate step to avoid deployment and publishing issues.
   db_conn_str=$(pg_connection_string "$PG_SERVER_NAME" "$DATABASE_PLACEHOLDER" "$ORCHESTRATOR_FUNC_APP_NAME")
   collab_db_conn_str=$(pg_connection_string "$CORE_DB_SERVER_NAME" "$COLLAB_DB_NAME" "$ORCHESTRATOR_FUNC_APP_NAME")
   az deployment group create \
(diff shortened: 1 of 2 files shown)


iac/create-resources.bash
```diff
@@ -306,7 +306,9 @@ main () {
   state_abbrs=${state_abbrs:1}
 
   # Create orchestrator-level Function app using ARM template and
-  # deploy project code using functions core tools.
+  # deploy project code using functions core tools. Networking
+  # restrictions for the function app and storage account are added
+  # in a separate step to avoid deployment and publishing issues.
```
edwintorres, line 311: Context https://github.com/18F/piipan/issues/2410#issuecomment-995979418

edwintorres (approved):

PR #2499, opened 2021-12-21
Dispose DbConnections to allow connection pooling

kingcomma:
- Return unopened connections from DB connection factories
- Wrap queries in `using` to dispose connections
- Lets Dapper manage opening and closing of one-off queries

Observed the issue by connecting to the database, running match queries, and using `select count(*) from pg_stat_activity;` to see any affect on the number of open connections. Prior to this change the number of open connections would climb after each query. With this change in place, the number is relatively stable (as expected).

Since all the `using` blocks affect indentation, hiding whitespace changes is helpful for review.

Closes #2499


timoballard (approved):
Looks great! 

_question (non-blocking)_: it looks like the differentiating factor for whether or not `OpenAsync` is called explicitly is if there are multiple commands executed against the connection. Does Dapper require an open connection in these cases, or is this a performance optimization? Either way, do you think it's worth adding a comment explaining the difference?

Thanks!

kingcomma:
Great point/question — it's a performance optimization. Dapper does not require an open connection, but if passed a closed connection would open and close the connection before and after each call. I'll add a comment before merging.

Issue #892, opened 2021-05-19
Configure appropriate redundancy setting for storage accounts that back Function Apps


ryanhofdotgov:
- [ ] Explore https://docs.microsoft.com/en-us/azure/storage/common/storage-disaster-recovery-guidance
- [ ] It is preferred to have identical redundancy settings across dev/test/pre-prod/prod, but if it is cost prohibitive, allow redundancy setting to be environment specific.

Focus here I believe should be on availability benefits within a zone rather than geo-redundancy across regions, as Function Apps can be re-created at will.

edwintorres:
@ryanhofdotgov the zone redundancy is not yet available on Azure Gov according to [the documentation](https://docs.microsoft.com/en-us/azure/azure-government/compare-azure-government-global-azure#azure-managed-disks).

Given that Function Apps can be re-created at will. The Standard_LRS should be ok for this.

PR #2503, opened 2021-12-22
Added storage redundancy configuration. It close #891 and #892

### LINKED ISSUE #891: Configure geo-redundant options for per-state storage accounts
- [ ] Evaluate options: https://docs.microsoft.com/en-us/azure/storage/common/storage-disaster-recovery-guidance
- [ ] Allow redundancy setting to be configurable per environment; e.g., we don't need redundancy at all for dev/test environments

edwintorres:
## What’s changing?

- Azure Storage sku to add region redundancy
- Changed to the activity monitor name, given that when the resources are changed to another region, the creation of the activity monitor will fail because it is a global resource. Added LOCATION as part of the name.

## Why?

Explained on #891 and #892

## This PR has:

## Test
This was tested on `tts/eth2` deployed to westus2

### DIFF
--- a/iac/arm-templates/function-storage.json
+++ b/iac/arm-templates/function-storage.json
@@ -16,6 +16,9 @@
         },
         "subnet": {
             "type": "string"
+        },
+        "sku": {
+            "type": "string"
         }
     },
     "variables": {
@@ -29,7 +32,8 @@
             "location": "[parameters('location')]",
             "tags": "[parameters('resourceTags')]",
             "sku": {
-                "name": "Standard_ZRS"
+                "name": "[parameters('sku')]"
+                // "Standard_ZRS"
             },
             "kind": "StorageV2",
             "properties": {
--- a/iac/arm-templates/function-storage.json
+++ b/iac/arm-templates/function-storage.json
@@ -29,7 +29,7 @@
             "location": "[parameters('location')]",
             "tags": "[parameters('resourceTags')]",
             "sku": {
-                "name": "Standard_LRS"
+                "name": "Standard_ZRS"
             },
             "kind": "StorageV2",
             "properties": {
--- a/iac/arm-templates/function-storage.json
+++ b/iac/arm-templates/function-storage.json
@@ -33,7 +33,6 @@
             "tags": "[parameters('resourceTags')]",
             "sku": {
                 "name": "[parameters('sku')]"
-                // "Standard_ZRS"
             },
             "kind": "StorageV2",
             "properties": {
--- a/iac/env/tts/dev.bash
+++ b/iac/env/tts/dev.bash
@@ -4,7 +4,7 @@
 ENV=$(basename "${BASH_SOURCE%.*}")
 
 # Default location
-LOCATION=westus
+LOCATION=westus2
 
 # Default resource group
 RESOURCE_GROUP=rg-core-$ENV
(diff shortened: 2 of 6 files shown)


kingcomma:
*Question:* Can you remind me, is zone-redundant storage available in `westus` or will we need to update our `tts` environments to use a different region?

edwintorres:
@kingcomma, this issue merits a conversation. The SKU Standard_ZRS is not available on westus; that is why I deployed to westus2.

Ryan checked Azure Gov and the SKU Standard_ZRS is available there, but the storages need to be deleted first. But we also need to manage the Azure Monitor Name because I changed the name, so I was able to test it on our Azure, but that is not needed on the Azure Gov.

kingcomma:
Cool — As we [just discussed](https://gsa-tts.slack.com/archives/C018P1ZJPEC/p1640199079023200), let's update the two `tts` environment files to specify `LOCATION=westus2`.

Also, can you update the [Orchestrator ARM template](https://github.com/18F/piipan/blob/eth/change-azure-storage-configuration2/iac/arm-templates/function-orch-match.json#L43) and the CLI commands for the metrics collector and metrics API  storage accounts in [`create-metrics-resources`](https://github.com/18F/piipan/blob/eth/change-azure-storage-configuration2/iac/create-metrics-resources.bash)? I think that should cover everything intended by #892.

edwintorres:
The last update includes STORAGE_SKU environment variable to set the Azure Storage SKU. This was implemented for Function Orchestrator #892 and blob storage #891

iac/env/tts/dev.bash
```diff
@@ -33,3 +33,7 @@ QUERY_TOOL_APP_IDP_CLIENT_ID=a8e3c164-77a9-45fd-9950-cc9862aa774a
 
 # SIEM tool app registration name
 SIEM_RECEIVER=$PREFIX-siem-tool-$ENV
+
+# Azure Storage SKU
+STORAGE_SKU="Standard_ZRS" # Standard Zone Redundant Storage
+# STORAGE_SKU="Standard_ZRS" # Standard Locally Redundant Storage (Use this when ZRS is not avilable in the region)
```
kingcomma, line 39:
```suggestion
# STORAGE_SKU="Standard_LRS" # Standard Locally Redundant Storage (Use this when ZRS is not available in the region)

```

iac/env/tts/dev.bash
```diff
@@ -33,3 +33,7 @@ QUERY_TOOL_APP_IDP_CLIENT_ID=a8e3c164-77a9-45fd-9950-cc9862aa774a
 
 # SIEM tool app registration name
 SIEM_RECEIVER=$PREFIX-siem-tool-$ENV
+
+# Azure Storage SKU
```
kingcomma, line 37:
```suggestion
# Azure Storage SKU for per-state storage accounts and storage accounts backing function apps
```

iac/arm-templates/function-storage.json
```diff
@@ -29,7 +32,8 @@
             "location": "[parameters('location')]",
             "tags": "[parameters('resourceTags')]",
             "sku": {
-                "name": "Standard_LRS"
+                "name": "[parameters('sku')]"
+                // "Standard_ZRS"
```
kingcomma, line 36:
```suggestion
```

kingcomma (commented):
Thanks for these improvements — looks great!

There are a couple other places where this change needs to be applied in order to cover all the storage accounts that back our Function apps:

- The storage account that backs the Metrics Collector Function App: https://github.com/18F/piipan/blob/93983828d1612f4c4156120b1d81e115a2d4f3e2/iac/create-metrics-resources.bash#L51
- The storage account that backs the Metrics API Function App: https://github.com/18F/piipan/blob/93983828d1612f4c4156120b1d81e115a2d4f3e2/iac/create-metrics-resources.bash#L177

Do you mind updating those two SKUs as part of this PR?

edwintorres:
@kingcomma thank you for the review. I made all that changes.

kingcomma (approved):
👍 👍

PR #2511, opened 2021-12-27
Added documentation related to 429 too many request

edwintorres:
## What’s changing?

Duplicate Participation API docs updated to include 429 Too Many Requests as a possible response and document the response body format

## Why?

The API Rate Limite was implemented, but the documentation not. 

## This PR has:


match/docs/openapi/schemas/match-query.yaml
```diff
@@ -272,3 +272,5 @@ components:
               $ref: '../schemas/match-query.yaml#/MatchResponseExamples/TopLevelError'
     '400':
       description: "Bad request. Request body does not match the required format."
+    '429':
+      description: "Too Many Requests. The status code indicates the user has sent too many requests in a given amount of time 'rate limiting'."
```
kingcomma, line 276: *Issue:* Since this file just contains bits that are reused elsewhere, there are two other files that need to be edited to include references to this error message:
- [match/docs/openapi/orchestrator/find.yaml](https://github.com/18F/piipan/blob/4e9b8ad38b334fd61a15ad344eb4bf421036df7c/match/docs/openapi/orchestrator/find.yaml#L15-L16)
- [docs/openapi/duplicate-participation-api.yaml](https://github.com/18F/piipan/blob/4e9b8ad38b334fd61a15ad344eb4bf421036df7c/docs/openapi/duplicate-participation-api.yaml#L23-L24)

You should be able to follow the same approach that's used for the `400` error in each doc.

You'll also need to run `docs/openapi/tools/generate-docs.bash` so the update is included in the [API docs](https://github.com/18F/piipan/blob/dev/docs/openapi/generated/duplicate-participation-api/openapi.md) we link to from the quick start guide for states.

kingcomma (changes requested):

match/docs/openapi/schemas/match-query.yaml
edwintorres: I think is fixed.

PR #2541, opened 2022-01-03
Add diagnostic settings for state upload resources (blob)

### LINKED ISSUE #977: Stream logs from blob storage to event hub
Send logs from blob storage to event hub in a manner that is supported by `AzureCloud` as well as `AzureUSGovernment`. 

In #976 we reverted the changes done in #950, due to this error when running the template in `AzureUSGovernment`:
```
Subscription 'XXXXXXX' is not allowed in the private preview of diagnostic log settings for Azure resource type 'microsoft.storage/storageaccounts/blobservices', feature flag: 'microsoft.insights/diagnosticsettingpreview'."
```
Consulting FastTrack for guidance on how to implement this in a cross-cloud manner.

echappen:
## What’s changing?

Adding back event hub configurations for per-state upload storage resources (blob-specific)

## Why?

Closes #977

Past attempts were made at this back in June (described in #977), but it seemed Diagnostic Settings were not yet out of preview mode for Azure Government. I tested in a Gov Cloud environment (`fns/dev`) through the portal, through Azure CLI, and by deploying this ARM template in isolation, and found the settings to work each time. So I'm more confident that we won't experience the same issues we once had. 

## This PR has:


iac/arm-templates/blob-storage.json
```diff
@@ -19,6 +19,12 @@
         },
         "sku": {
             "type": "string"
+        },
+        "resourceGroup": {
+            "type": "String"
+        },
```
kingcomma, line 25: *Question/nitpick:* It's a minor detail, but since this is the same value as the `--resource-group` parameter passed to `az deployment`, can we use the built-in `resourceGroup().name` function like we've done in [other ARM templates](https://github.com/18F/piipan/blob/dev/iac/arm-templates/participant-records.json#L55)?

kingcomma (approved):
👍 LGTM!

Issue #1940, opened 2021-10-19
API schema for match record retrieval


cjhskippy:
In order to complete #1656, we first need to provide the frontend with access to match details.  This will be provided by an API.

For the initial launch of the NAC, this API will only be an internal one.

The API needs to provide the information about matches envisioned in this mockup:
<img width="631" alt="Screen Shot 2021-12-29 at 3 33 11 PM" src="https://user-images.githubusercontent.com/1299543/147711271-4b3f03b5-e6de-464c-a9a7-05415d5b2e07.png">

AC:
- [ ] API schema is drafted containing all of these data elements:
  - [ ] Inputs:
    - [ ] A single Match ID
  - [ ] Outputs:
    - [ ] Match status (open / closed)
    - [ ] Match creation date
    - [ ] Name of querying state
    - [ ] Name of matched state
    - [ ] vulnerable status in the querying state
    - [ ] vulnerable status in the matched state
    - [ ] match validity in the querying state
    - [ ] match validity in the matched state
    - [ ] initial action taken by the querying state (OPTIONAL for now - allowed values are currently undefined)
    - [ ] initial action taken by the matched state (OPTIONAL for now - allowed values are currently undefined)
    - [ ] Date of initial action by the querying state
    - [ ] Date of initial action by the matched state
    - [ ] Final disposition by the querying state (OPTIONAL for now - allowed values are currently undefined)
    - [ ] Final disposition by the matched state (OPTIONAL for now - allowed values are currently undefined)
    - [ ] Date and time final disposition was assigned by the initiating state
    - [ ] Date and time final disposition was assigned by the matched state
    - [ ] Case id in the querying state
    - [ ] Case id in the matched state
    - [ ] Participant ID in the querying state
    - [ ] Participant ID in the matched state
    - [ ] Benefits end date in querying state (OPTIONAL for now - field is currently under review)
    - [ ] Benefits end date in querying state (OPTIONAL for now - field is currently under review)
    - [ ] Recent benefits months in querying state
    - [ ] Recent benefits months in matched state

cjhskippy:
Closing.  This has been addressed / was tracked in Jira.

PR #2543, opened 2022-01-05
Add schema to get a match record through new internal API

echappen:
## What’s changing?

- Adds openapi schema for a new Collaboration API for internal use 
- Adds openapi schema to get a single match record from Collaboration API

## Why?

Partially Addresses #1940 

This API schema simply exposes the [match record data we already store](https://github.com/18F/piipan/blob/dev/match/ddl/match-record.sql#L17). #1940 defines API fields for full match resolution, i.e. state-level disposition tracking. Our current data store doesn't yet handle state-level disposition tracking. However, it holds enough data to get an initial Match Record page up in the Query Tool, which are:

- Match status (open / closed)
- Match creation date
- Name of querying state
- Name of matched state
- vulnerable status in the matched state (provided in the original data output)
- Case id in the matched state (provided in the original data output)
- Participant ID in the matched state (provided in the original data output)
- Benefits end date in matched state (provided in the original data output)
- Recent benefits months in matched state (provided in the original data output)

You'll see that a lot of this data comes from the data output of the original match query, which is stored as a `jsonb` field on the match record. I think there's discussion to be had on how heavily we want to rely on this data, or if we should retrieve this data in a different way. 

It's also unclear whether or not the querying state will have equivalent participant data in our system, as shown in the visual for #1940, since the query could be for either a new participant or an existing one. 

## This PR has:

## Anything else?

A directory `widdershins_templates` was added to accommodate custom templating for widdershins, the tool we use to auto-generate our API documentation. It was [recommended](https://github.com/Mermade/widdershins#templates) to copy all templates over instead of piecemeal, and my experiments with custom templating proved that to be preferable. The result is having a bunch of templates we potentially don't need, but the upside is we can have more control over what our API docs look like. It won't be necessary to code review these templates.


echappen:
@kingcomma Wanted to double-check that this internal API would use the same `Authorization: Bearer {token}` schema in the header, as described in our [docs](https://github.com/18F/piipan/blob/dev/docs/securing-internal-apis.md).

echappen:
Closing this in favor of #2548 which adds a Collaboration subsystem.

PR #2547, opened 2022-01-06
Use custom widdershins templates

echappen:
## What’s changing?

- Adds widdershins openapi markdown templates to make them available for customization
- Adds slight modifications to widdershins templates to remove long urls for nested parameters that broke our generated API doc tables. 
- Regenerates API docs to fix long urls issue

## Why?

Our Openapi yaml schema uses many references to modular components. While this helps us when composing yaml templates, it resulted in parameter tables that appeared broken in our generated API docs, due to unusually long urls placed into the tables. 

## This PR has:

## Anything else?

A directory `widdershins_templates` was added to accommodate custom templating for widdershins, the tool we use to auto-generate our API documentation. It was [recommended](https://github.com/Mermade/widdershins#templates) to copy all templates over instead of piecemeal, and my experiments with custom templating proved that to be preferable. The result is having a bunch of templates we potentially don't need, but the upside is we can have more control over what our API docs look like. It won't be necessary to code review these templates.


docs/openapi/generated/duplicate-participation-api/openapi.md
```diff
@@ -58,8 +57,8 @@ Searches all state databases for any participant records that are an exact match
 |Name|In|Type|Required|Description|
 |---|---|---|---|---|
 |From|header|string|false|As in the HTTP/1.1 RFC, used for logging purposes as a means for identifying the source of invalid or unwanted requests. The interpretation of this field is that the request is being performed on behalf of the state government-affiliated person whose email address (or username) is specified here. It is not used for authentication or authorization.|
-|data|body|[object]|true|none|
-|» lds_hash|body|string|true|SHA-512 digest of participant's last name, DoB, and SSN. See docs/pprl.md for details|
+|data|body|object|true|none|
+|» lds_hash|body|object|true|SHA-512 digest of participant's last name, DoB, and SSN. See docs/pprl.md for details|
```
kingcomma, line 61: *Issue:* Did you happen to copy the template files from the Github project? I noticed these parameter types were changed incorrectly — e.g., `data` should be an array of objects and `lds_hash` should be a `string` — and looked into why that happened. Strangely, the latest version of widdershins on npm is `4.0.1` but the latest version on Github is `4.0.0`. The template files between the two versions differ slightly and account for these changes.

That's not the most encouraging finding, but I think we should copy the templates from `4.0.1` to minimize these changes to our generated docs.

echappen: That change in the parameter types was the part of the template that I had to alter in order to fix the long urls. I can also look into the npm vs github discrepancy, but I think this issue is what I changed in the templates:

Typically in the templates, the parameter type value comes from `p.safeType()`. This was producing a long url for nested `$refs` from the yaml template, which visually broke the tables. 

In the `response` partial, I could use `p.type()` but in the parameters partial, I had to use `{{=p.originalType||'object'}}` meaning if the `type` didnt exist, "object" would go in its place. I was assuming that any `$ref`s were related to objects, but that's not the case here. I'll try to think of a way to make this more accurate.

echappen: I ended up reverting back to `safeType` for now since that outputs the expected. My generation of the collaboration API docs probably prompted me to change it, but that's now moot.

kingcomma: Ah, got it! I didn't realize those updates to the template files were made on our end and was incorrectly attributing the differences in the generated files to the difference between the `4.0.1` and `4.0.0` template files.

Does the difference between `p.type` and `p.safeType` also explain the changed types in the responses section (e.g., `array` instead of `[string]` for `recent_benefit_months`)? Those changes aren't incorrect, but it does seem more helpful for an implementor to know a given property is an "array of strings" rather than just an "array".

echappen: It explains those changes too. It'll take some refining, but we're limited to [what's available to us](https://github.com/Mermade/widdershins/blob/main/templates/openapi3/README.md#parameter-template) in the templates. Personally, I think the tables are most helpful for their sentence descriptions—the json examples above give a much better sense of the data types.

kingcomma (approved):
👍

PR #2645, opened 2022-01-28
Bump Npgsql to 6.0.3

kingcomma:
## What’s changing?

Bump Npgsql to 6.0.3

## Why?

Flagged by dependabot. See reference: [Updating .NET dependencies](https://github.com/18F/piipan/blob/main/docs/update-deps.md)


echappen (approved):
Sorry I missed this! Approved!

PR #2664, opened 2022-02-02
Add GetRecordById function to MatchRecordDao

echappen:
## What’s changing?

Adds a function to `MatchRecordDao` to find a Match Record by Match ID. 

## Why?

Partially addresses #1940

This is the first small step to implementing the Collaboration API. This function will be used by the Collaboration API when a state eligibility worker supplies a Match ID from a previous match query, and expects to receive detailed match record information. 

@kingcomma @edwintorres Could you review to see if I've missed any documentation or other places the function should be referenced.

@bgfoshay Feedback is welcome, particularly on whether this function should return an empty array or throw an error if a record isn't found. My thinking was that the Dao's only responsibility is to perform the query, and an error message would be an API service's responsibility. But I'm open to other thoughts. Let me know if I should tag others from Ventera on this.

## This PR has:

## Anything else?

PR also fixes a test name of an existing MatchRecordDao function.


bgfoshay:
I think this looks good and it's easy to follow. The tests look good as well.

thought: Regarding your question about the function returning an empty error vs an error, I agree with you that it the dao's responsibility is simply to execute the query. From its perspective, it's not an error if no record is found. 

question: This is probably a silly question but is it possible that multiple records can exist with the same matchId? Looking at the db, it looks like there's a uniqueness constraint on match_id so it doesn't seem like it'd be possible. The sql query in this code limits the return to one record. Could there ever be a case where there are multiple, and if so, is limiting the result set to a single record the correct action?

echappen:
@bgfoshay To answer your question, you’re correct that our current implementation does not expect multiple records with the same Match ID. I could see that changing in the future if we move to an Event Sourcing model. 

Given our current implementation, it does seem confusing and redundant to have `LIMIT 1` on the query, so I can leave it off for now.

match/src/Piipan.Match/Piipan.Match.Core/DataAccessObjects/MatchRecordDao.cs
```diff
@@ -114,5 +114,39 @@ FROM matches
...
+                    match_id = @MatchId
+                ;";
+
+            using (var connection = await _dbConnectionFactory.Build())
+            {
+                return await connection.QueryAsync<MatchRecordDbo>(sql, new MatchRecordDbo {
+                    MatchId = matchId
+                });
+            }
+        }
```
kingcomma, line 150: *Question:* Can you go into more detail on how you see an event sourcing approach resulting in the need to return an enumerable here?

Since we anticipate the collaboration API being a REST API, I'm wondering if returning a single record would be a better fit. Thinking ahead to the API implementation, the convention for a `GET` request seems to be returning a 404 if a resource does not exist (e.g., [JSON:API](https://jsonapi.org/format/#fetching-resources-responses-404) and [GitLab](https://docs.gitlab.com/ee/api/#status-codes)). Using a method like `QuerySingleAsync` would [throw an exception](https://dapper-tutorial.net/querysingle#first-single-default) if no record is found, which plays nicely with [our current approach](https://github.com/18F/piipan/blob/dev/match/src/Piipan.Match/Piipan.Match.Func.Api/Api.cs#L69-L84) for returning error responses from our APIs.

echappen: Returning a single record from the Dao makes sense for a GET. Are you suggesting throwing the 404 error at the API level or the Dao level? 

With an Event Sourcing approach, I could see the Match ID being the ID used to collect all match record events related to a single Match. But I'm still thinking through whether that would be a new table/model, or if we'd modify the existing Match Record for that approach.

kingcomma: Ah, gotcha! Thanks for that explanation.

For the Dao, I'm suggesting letting the object mapper (Dapper, in this case) throw the exception if a record does not exist. That is the default functionality for Dapper's `QuerySingleAsync` method, so throwing the exception could be achieved by swapping `QueryAsync` with `QuerySingleAsync`. The DAO would be unaware of the exception — it would naively run the query and attempt to return the result. (Same approach as the [UploadDao](https://github.com/18F/piipan/blob/dev/participants/src/Piipan.Participants/Piipan.Participants.Core/DataAccessObjects/UploadDao.cs#L19-L30)).

The exception could then bubble up to the Function app / API level where it would be caught and handled appropriately (i.e., return a 404). This would follow the same pattern as the match API. E.g., it catches a `ValidationException` at the API level and returns the appropriate error response to the client.

echappen:
Resurrecting this PR now that Match Resolution implementation is underway and this functionality is needed for it. I switched the name of the function to `GetRecordByMatchId`, in case a query by Primary Key is ever needed in the future. This function now returns a single record when one is found, and throws an exception when not found, per @kingcomma's request.

echappen:
@kcherukupalli Since Bill is out this week, do you mind reviewing this PR on the Ventera side? This is a supporting function for the Event Sourcing approach being implemented.

kcherukupalli (approved):
Looks good.

PR #2677, opened 2022-02-09
Add ADR for using Event Sourcing for Match Resolution

echappen:
## What’s changing?

Adds ADR proposing to use an Event Sourcing data model for Match Resolution

## Why?

Working on the Match Resolution API opened up discussion topics on data modeling. My goal is for both 18F and Ventera to review this and come to consensus on a path forward for implementing the Match Resolution functionality. 

## This PR has:


docs/adr/0027-use-event-sourcing-for-match-resolution.md
```diff
@@ -0,0 +1,152 @@
...
+Proposed
+
+## Context
+
+When matches are found, government entities must work together to resolve them. Our system needs to provide functionality to enable this collaboration and to audit this activity.
+
+The system will need to:
+- show the details and status of a match ("open" or "closed", etc.)
+- changes the details and status of a match ("open" -> "closed") based on activity
+- show activity history of a match (who did what? when did they do it? Why did they do it?)
```
cjhskippy, line 14: I'm glad that we're thinking about how this can be accomplished with an eye toward reducing the amount of rework that gets done later. However, activity history is something that we'd like to be able to keep out of scope of the initial launch.  The approach as proposed looks like it will front-load most of the history tracking, which increases the scope of the MVP compared to some of the other proposed approaches.

docs/adr/0027-use-event-sourcing-for-match-resolution.md
```diff
@@ -0,0 +1,152 @@
...
+
+#### Benefits of this approach:
+- Single source of truth
+- Auditing is baked into the design; querying Activity history is straightforward
+- Immutability of data eliminates doubt over change history.
+- Deriving both the current state and former state is doable and of equal effort. For example, one could get what the status of a match is right now, or what it was a week ago with equal effort.
+- Schema changes are already time-based; outdated Event records can be discarded; versioning records can mitigate against unnecessary migrations
+
+#### Drawback of this approach:
+- Querying for "state" becomes slightly more complex, could have scalability implications.**
```
cjhskippy, line 68: There is some asymmetry in the value of these two things.  Having the correct current state is paramount; having the correct prior state will have low, occasional value.  If anything goes wrong in the replay chain, we lose the meaning of the current state.  It would be worth it in my mind to sacrifice immutability to trace the state change backward from current state rather than forward from the original state, but I think there may be another option.

Given that these records will be rare, fairly small, and will likely be changed infrequently, I wonder if we might be better off saving the entire contents of every record, rather than just a delta from one record to the next?  In that scenario, you would be able to find the current state at any given time by searching for the most recent event record for any given match ID.

This has the advantage of always preserving the correct current state, retaining an immutable audit history, and might allow us to reduce the MVP scope compared to what I anticipate might be involved in creating deltas, interpreting them, and implementing error checking to detect or resolve race conditions.  My suggestion would involve some additional storage compared to this approach, but if we find that storage costs or scalability makes it worth the effort, we could come back to this proposal as an optimization.

cjhskippy (commented):
I'm really glad to see the foresight into the future complexity that history tracking will bring.  I added some inline comments with another variation that might work for our use case.

docs/adr/0027-use-event-sourcing-for-match-resolution.md
echappen: To clarify, and to document some of that we talked about offline: we don't anticipate this approach being significantly more work than other approaches. Any approach we choose will need to:

- record state-provided data, with timestamps on state actions
- update the status of a match depending on state actions

Any approach we choose will also require changes to the data schema.

Where we anticipate the most complexity with Event Sourcing compared to other approaches is in aggregating domain data for UI purposes. With any approach, data from the database is unlikely to be in the exact format we would need for an API that serves a UI. So regardless of approach, we will need to format data for the UI's purposes. An Event Sourcing model would add perhaps a few lines more code to that formatting step. However, any aggregators we create would be very straightforward to Unit Test: create events, run them through the aggregator, and test the output. It's a completely functional approach. 

An Even Sourcing model could also make a few metrics queries more complex, but doable, but it can also make some queries easier. Overall, we anticipate it being easier to begin with an Event Sourcing model and convert elements to CRUD when the need arises, than it is to tack on an event-based model to a traditional CRUD model. In my experience, the latter leads to the worst of both worlds. 

We don't anticipate utilizing the full capabilities of an Event Sourcing approach in the MVP, but when the need for activity history arises, the underlying data will be there and ready for the using.

bgfoshay (approved):
I think this looks good. I think the design and its pros/cons have been thoroughly considered and this should work well. I also very much agree that this approach can very easily be transitioned to a traditional CRUD model if ever necessary (much easier than transitioning in the opposite direction). I think this design will benefit us now and positions us for even more benefit in the future.

edwintorres (approved):

cjhskippy (approved):

kcherukupalli (approved):
Agree with Event Sourcing approach.

kingcomma:
Jira/NAC-207

Issue #2554, opened 2022-01-10
Skeleton of match lookup web page


cjhskippy:
Make progress towards #1656, doing some of the work that is not currently blocked.  For this slice of progress, we can create the skeleton of the match lookup page, perform input validation, and show that the website is prepared to handle errors (invalid match provided / match lookup website offline).

AC:
- [ ] New match lookup page is added to the Query Tool and Collaboration App
- [ ] URL for the match lookup page includes the match ID
- [ ] Page extracts match ID from url and verifies that match ID meets validation criteria
- [ ] If the match ID provided is not valid, provide user with an appropriate error message
- [ ] If the match ID provided is valid, page attempts to call the Collaboration API to request match details (this will always fail for now, that is okay)
- [ ] Include the same navigation header and footer that we provide on the "Search for SNAP Participants" page.  Mockup below.
- [ ] When the Collaboration API cannot be reached, show the user an appropriate error message - something like "Uh-oh.  This page is unavailable right now.  Try again in a few minutes or contact support for help."
<img width="965" alt="Screen Shot 2022-01-10 at 9 03 27 AM" src="https://user-images.githubusercontent.com/1299543/148808534-562053c7-fca8-4486-84a7-b55e9c92119d.png">

PR #2692, opened 2022-02-17
Added match lookup web page

edwintorres:
## What’s changing?

Created a mockup of the collaboration page. The mockup used the USWDS and try to follow the design without investing time in custom CSS. 

Demo mockup collaboration page at Feb 17, 2022

## Why?

Progress toward  #2554

- [x] New match lookup page is added to the Query Tool and Collaboration App
- [x] URL for the match lookup page includes the match ID
- [ ] Page extracts match ID from url and verifies that match ID meets validation criteria
- [ ] If the match ID provided is not valid, provide user with an appropriate error message
- [ ] If the match ID provided is valid, page attempts to call the Collaboration API to request match details (this will always fail for now, that is okay)
- [x] Include the same navigation header and footer that we provide on the "Search for SNAP Participants" page. Mockup below.
- [ ] When the Collaboration API cannot be reached, show the user an appropriate error message - something like "Uh-oh. This page is unavailable right now. Try again in a few minutes or contact support for help."

## This PR has:

## Anything else?

I suggest moving incomplete subtasks to a different issue related to an invalid match id given that the API and the data structure is in progress.


query-tool/src/Piipan.QueryTool/Pages/Match.cshtml
```diff
@@ -0,0 +1,509 @@
...
+        </script>
+
+        <!-- Protect location warning -->
+        <div id="warning-protect-location" class="grid-row">
+            <div class="tablet:grid-col-12">
+                <div class="usa-alert usa-alert--warning">
+                    <div class="usa-alert__body">
+                        <h4 class="usa-alert__heading">This match involves a vulnerable individual</h4>
+                        <p class="usa-alert__text">
+                            One or both states has indicated that the individual involved in this match is considered vulnerable individual. This means that the individual's location should not be shared. For example, this individual may ne freeing domestic violence.
```
ccaprio97, line 133: Nitpick comment 

"be" in "this individual may ne freeing domestic violence" has a typo

query-tool/src/Piipan.QueryTool/Pages/Match.cshtml.cs
```diff
@@ -0,0 +1,144 @@
+using System;
```
bgfoshay, line 1: Nitpick: Import organization

bgfoshay (commented):
praise: Overall I think this looks like a really clean mockup of the page and the code is straightforward to follow.
question: If this PR is intended to be merged into dev, I believe we'll need tests to provide coverage for the new code?

cjhskippy:
Just recording some conversational details we had about match ID validation.  For this issue, I'm good with any validation that prevents malicious user input from being passed through to the API + frontend.  In this case, it might be best to go with match ID length = 7 characters, and is only composed of alphanumeric characters.

edwintorres:
@bgfoshay : If this PR is intended to be merged into dev, I believe we'll need tests to provide coverage for the new code?

I don't see much value in merging it until there is a backend. Then work with the actions on the page that right now are simulated. 

I can work with the validation to the match id that @cjhskippy added in his comment and put tests for that. 

Thank you for the review @bgfoshay @ccaprio97 @cjhskippy

query-tool/src/Piipan.QueryTool/Pages/Match.cshtml.cs
edwintorres: Sorry @bgfoshay Can you give me more context for this one.

bgfoshay: Was just a minor comment regarding the organization/sorting of the using statements. But upon looking at this again, I realize the two I was concerned about (Piipan.Match.Core.DataAccessObject & Piipan.Match.Core.Exceptions) actually appear to be commented out. So I think my original comment can be ignored.

edwintorres:
I updated this PR to include the validation @cjhskippy mentioned. Added tests to prove invalid characters, the length of the case id, and a valid test case id. Additionally, I improve vscode configuration to debug the test scenarios.

edwintorres:
@bgfoshay @ccaprio97 I'll close this PR, and split on 3 PR.

1. PR to validate the Match Id (match id of length 7 and alphanumeric) 
2. PR with vscode configuration to debug test scenarios 
3. PR with the front-end code, since the design team is reviewing the designs, it may be necessary to wait for this code to be approved.

PR #2695, opened 2022-02-18
Adding global piipan Visual Studio solution

bgfoshay:
## What’s changing?

Adding a global Visual Studio solution file

## Why?

Allows developers who choose to work within Visual Studio to have better IDE experience e.g. click navigation to class/interfaces across apps/projects, debug multiple applications simultaneously, etc.

## This PR has:

## Anything else?
Note: I won't be offended if we don't want this file in the repo. I thought other developers might benefit from it and save them the trouble of creating it themselves. This file is meant for Visual Studio development convenience. It has no effect on the existing csproj files and no effect on compilation and build procedures.


kingcomma:
@bgfoshay Nice! I'm all for this if it improves developer experience. I think even Visual Studio Code users would benefit from some of what you list out, e.g. clicking to navigate between projects.

*Question:* What do you think about moving this file up a level to the root of the repo in keeping with the [project structure](https://gist.github.com/davidfowl/ed7564297c61fe9ab814) we follow? I also think it would be good to rename it to something that's not tied to the name of this specific repository. Maybe `system.sln` or something similar?

If you're up for that, we should also update the [source tree checklist](https://github.com/18F/piipan/blob/dev/docs/source-guidelines.md#source-tree-checklist) with an additional step of adding new projects to the top-level solution.

bgfoshay:
@kingcomma Glad to hear you think it might be beneficial! I have no concerns moving it up a level to the root of the repo. And renaming it makes sense as well. I always struggle to come up with good names 😃 

I also agree that I need to update the source tree checklist as well with this new step. I will make all 3 edits and let you know when the changes are ready for review.

kingcomma (approved):
@bgfoshay Looks great!

PR #2699, opened 2022-02-22
Remove invalid and status fields from matches

echappen:
## What’s changing?

1. Remove invalid and status fields from matches
2. Returns existing match if one is found in query process

## Why? 

Relates to Jira NAC-165

This is in preparation for implementing [match resolution events](https://github.com/18F/piipan/blob/dev/docs/adr/0027-use-event-sourcing-for-match-resolution.md) which will hold this data.

Because there is (temporarily) no longer a concept of a "closed" match, the Match Event Service will simply return any existing match found. This routine for determining whether the latest match is closed in this service will be re-implemented once we add match resolution events and can determine current state from them.

Once this is merged, the next steps in implementing NAC-165 are to:
1. add the `match_res_events` table to the collaboration ddl
1. add a `MatchResEventDao` to handle database queries for this table
1. add a `MatchAggregator` service to compile stateful data from match res events
1. use the Match Aggregator service to determine open/closed status in `MatchEventService.Reconcile` 

## This PR has:

## Anything else?

- Collaboration Postgresql database server will need to be dropped prior to running IAC on deployment


kingcomma:
*Question:* Is there a way to approach this change in a manner that does not require fairly heavy changes to the applications and unit/integration tests?

One idea: could we hard-code an "open" status in the SQL statement that retrieves match records? I think that would eliminate the need for any changes to the applications or tests and preserve current functionality. We could then iteratively refactor once match record events are implemented.

echappen:
> _Question:_ Is there a way to approach this change in a manner that does not require fairly heavy changes to the applications and unit/integration tests?
> 
> One idea: could we hard-code an "open" status in the SQL statement that retrieves match records? I think that would eliminate the need for any changes to the applications or tests and preserve current functionality. We could then iteratively refactor once match record events are implemented.

Yep, can do!

kingcomma:
@bgfoshay @kcherukupalli We had some discussion internally and decided to approach this work in the following order:
- Hold off on dropping these two fields for now
- Start with the match resolution events table schema as its own PR
- Implement the application-level changes to support match resolution events
- Drop these two fields once they are no longer used

echappen:
I'll close this PR then, in favor of the order above

kcherukupalli:
Sure, we can drop those at a later stage.

bgfoshay:
I think that approach makes a lot of sense.

PR #2701, opened 2022-02-24
add database access objects for querying match resolution events

echappen:
## What’s changing?

Adds database access objects for querying match resolution events

## Why?

Relates to Jira NAC-165

This is the next step in implementing [match resolution events](https://github.com/18F/piipan/blob/dev/docs/adr/0027-use-event-sourcing-for-match-resolution.md) based on https://github.com/18F/piipan/pull/2699#issuecomment-1048921838.

For the purposes of match resolution, the only queries needed for now are:
- insert new events
- get all events related to a match, in order of insertion 

Once this is merged, the next step is to add a MatchAggregator service to compile stateful data for matches, which will be in the next PR. 

## This PR has:


match/src/Piipan.Match/Piipan.Match.Core/DataAccessObjects/MatchResEventDao.cs
```diff
@@ -0,0 +1,100 @@
...
+        }
+
+        /// <summary>
+        /// Finds all match resolution events related to a given match ID,
+        /// and orders results by oldest event first
+        /// </summary>
+        /// <param name="matchId">The given match ID</param>
+        /// <param name="timestamp">Will select all events prior to this timestamp. Pass the current timestamp to get all related records. Timestamp is assumed to be UTC</param>
+        /// <returns>IEnumerable of IMatchResEvents</returns>
+        public async Task<IEnumerable<IMatchResEvent>> GetRecords(
```
kcherukupalli, line 73: It might be helpful if we add  Sortby as a parameter and default it to "asc" .

echappen: Good call, will do!

match/src/Piipan.Match/Piipan.Match.Core/DataAccessObjects/MatchResEventDao.cs
```diff
@@ -0,0 +1,111 @@
...
+        /// </summary>
+        /// <param name="matchId">The given match ID</param>
+        /// <param name="timestamp">Will select all events prior to this timestamp. Pass the current timestamp to get all related records. Timestamp is assumed to be UTC</param>
+        /// <param name="sortByAsc">Boolean indicating ascending sort order, defaults to true. Argument of false is descending order</param>
+        /// <returns>IEnumerable of IMatchResEvents</returns>
+        public async Task<IEnumerable<IMatchResEvent>> GetRecords(
+            string matchId,
+            DateTime timestamp,
+            bool sortByAsc = true
+        )
```
kingcomma, line 77: *Question:* How often do you anticipate needing to get only events before a given time vs getting all events? I always get worried about the unintended consequences of having to create timestamps and pass around `DateTime` objects. What do you think about having this method get all events and, if necessary, having a separate method for time-bound operations (e.g., `GetEventsBefore`)?

match/src/Piipan.Match/Piipan.Match.Core/DataAccessObjects/IMatchResEventDao.cs
```diff
@@ -0,0 +1,18 @@
...
+using System.Threading.Tasks;
+using Piipan.Match.Api.Models;
+using Piipan.Match.Core.Models;
+
+namespace Piipan.Match.Core.DataAccessObjects
+{
+    public interface IMatchResEventDao
+    {
+        Task<int> AddRecord(MatchResEventDbo record);
+        Task<IEnumerable<IMatchResEvent>> GetRecords(
```
kingcomma, line 12: *Nit:* `AddEvent` and `GetEvents` are probably more descriptive method names and in line with how we've named DAO methods elsewhere: e.g. `AddParticipants`, `GetUploads`.

match/src/Piipan.Match/Piipan.Match.Core/Models/MatchResEventDbo.cs
```diff
@@ -0,0 +1,21 @@
...
+    /// </summary>
+    public class MatchResEventDbo : IMatchResEvent
+    {
+        public int Id { get; set; }
+        public string MatchId { get; set; }
+        public DateTime InsertedAt { get; set; }
+        public string Actor { get; set; }
+        public string Delta { get; set; }
+        [Column(TypeName = "jsonb")]
+        public string ActorState { get; set; }
```
kingcomma, line 19: *Issue:* I think this column attribute should be applied to `Delta` not `ActorState`?

echappen: Ah yes, I forgot the specs go before the property and not after

match/src/Piipan.Match/Piipan.Match.Core/DataAccessObjects/MatchResEventDao.cs
echappen: I don't see an immediate need for passing timestamps other than current. So I'm fine with setting the timestamp as `now()` in the query for this function, then creating a new `GetEventsBefore` function down the road if need be.

match/src/Piipan.Match/Piipan.Match.Core/DataAccessObjects/IMatchResEventDao.cs
echappen: Agreed, will change!

match/src/Piipan.Match/Piipan.Match.Core/DataAccessObjects/MatchResEventDao.cs
echappen: Or just get all the events and not have a time clause at all

echappen:
This is updated and ready for review

kcherukupalli:
Looks good to me.

echappen:
Thanks @kcherukupalli! Do you or @kingcomma mind hitting "Approve" so I can merge?

bgfoshay (commented):
Question: Do we typically test for the case when a matching record cannot be found?

echappen:
@bgfoshay I've been adding tests for when not found throws an error. In this case, a non-existent match ID would return an empty array. Do you think that's a case I should test for? 

Also, do you think an empty array is an appropriate response here, or would you expect an error?

bgfoshay:
@echappen An empty array makes sense to me. It wouldn't hurt to add a test but that being said, if that's not a pattern we've been following I'm ok not introducing it now.

bgfoshay (approved):

PR #2702, opened 2022-02-24
Cjc nac 163

ccaprio97:
Made Match API take participant Id and Case Id as optional values to add to the input in the match record

## What’s changing?

RequestPerson model now has participant_id and Case_id as optional params

## Why?

Jira issue NAC-163

## This PR has:

### DIFF
--- a/docs/openapi/generated/duplicate-participation-api/openapi.md
+++ b/docs/openapi/generated/duplicate-participation-api/openapi.md
@@ -59,6 +59,8 @@ Searches all state databases for any participant records that are an exact match
 |From|header|string|false|As in the HTTP/1.1 RFC, used for logging purposes as a means for identifying the source of invalid or unwanted requests. The interpretation of this field is that the request is being performed on behalf of the state government-affiliated person whose email address (or username) is specified here. It is not used for authentication or authorization.|
 |data|body|[object]|true|none|
 |» lds_hash|body|string|true|SHA-512 digest of participant's last name, DoB, and SSN. See docs/pprl.md for details|
+|» case_id|body|string|false|Participants Case Number within a State|
+|» participant_id|body|string|false|Participants Participant ID within a state|
 
 > Example responses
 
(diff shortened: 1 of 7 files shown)


bgfoshay (approved):
Looks good.

Question: Assuming further PRs coming next for how new request parameters will be handled? Actually, upon further discussion I'm not sure if other work is required here. Conner reminded me that the entire request is persisted in the "input" column of the matches table. Given that's the case, I think this is a complete solution. @kingcomma Is this understanding correct?

kingcomma:
Correct, this update will cause any incoming `case_id` and `participant_id` values (along with `lds_hash`) to be parsed to a `RequestPerson` instance. That is sufficient to have the values recorded along with the match records in the `input` field.

I think that still leaves the final two acceptance criteria in NAC-163. Assuming those be addressed in future PRs?

ccaprio97:
I can update the API doucmentation now and Yes, I believe there will be another PR to update the query tool to send part id and case id and another PR to make the participant ID required and update the api documentation in github again

kingcomma:
@ccaprio97 Sounds good — were you planning to update the API documentation as part of this PR or a subsequent PR? Either one works, I'll just hold off on my review if you're adding it to this PR.

ccaprio97:
took some trial and error but with Edwin's and Ryan Harvey's help i was able to figure out the API documentation update. So this should be all the updates for this ticket and the next query tool PR will be seperate

kingcomma:
Nice, thanks! The documentation generation process is a little wonky and definitely could be better documented. In this case, it looks like the public-facing ("generated") docs were updated but not the underlying OpenAPI schema that is used to generate those docs. For this PR, can you do the following steps to make sure both the schema and the docs are updated and in sync?

1. Update the [`Person` schema in `match/docs/openapi/schemas/match-query.yaml`](https://github.com/18F/piipan/blob/dev/match/docs/openapi/schemas/match-query.yaml#L36-L44) to include the optional participant_id and case_id fields
2. Run `docs/openapi/tools/generate-docs.bash` to [generate the updated public-facing documentation](https://github.com/18F/piipan/tree/dev/docs/openapi#generating-documentation). This will overwrite the changes already made to the docs in the `generated` directory.

kingcomma:
Also, for the description of both of those fields, let's base them off of the existing descriptions in the [bulk upload file schema](https://github.com/18F/piipan/blob/dev/etl/docs/csv/import-schema.json):

- `participant_id`: Participant's state-specific identifier. Must not be social security number or any personal identifiable information.
- `case_id`: Participant's state-specific case number

ccaprio97:
Hey Michael, sorry for so much back and forth on this PR but the changes you asked for should be done now. Let me know if there's anything you want me to change. Thanks!

kingcomma (approved):
👍 👍 Looks great!

And no need to apologize at all — this back and forth led to some super valuable knowledge transfer.

PR #2703, opened 2022-02-24
Add Match Resolution Aggregator

echappen:
## What’s changing?

Adds models and builder to aggregate stateful match data based on match record events.

## Why?

Relates to Jira NAC-165

This is the next step in implementing [match resolution events](https://github.com/18F/piipan/blob/dev/docs/adr/0027-use-event-sourcing-for-match-resolution.md) based on https://github.com/18F/piipan/pull/2699#issuecomment-1048921838.

Once this is merged, the next step is to use this aggregate in the Active Match Reconciliation process, instead of the soon-to-be-removed `status` field on `matches`. This will be the next PR. 

## This PR has:

## Anything else?

See discussion below about naming.


kingcomma:
Nice! A few thoughts related to your feedback questions:

I would expect the most common interaction with match records, at least to start, will be an application retrieving a record in its current state and acting accordingly. The aggregation functionality would be behind the scenes and not really a relevant detail at that level. That might suggest simplifying the model names by dropping "Aggregate" — maybe something like a `MatchRecordAggregator` that returns a `MatchRecord` with the necessary fields, including `Dispositions` for each participating state.

In terms of what data is passed to the aggregator, I think there's an advantage to the aggregator focusing on replaying events against a record and not have it control/know how the record and events are retrieved. Also keep in mind the match API needs to retrieve an aggregate record without knowing a match ID in advance.

echappen:
@kingcomma Thanks for that feedback! I'll make it so that the Aggregator accepts a `MatchRecordDbo` and a collection of its `MatchResEventDbo`'s so it's up to the caller to collect those objects. 

A name like `XYZAggregator` for the aggregator makes sense to me. For `MatchRecord`, would using a name that's more specific to the context would be helpful, like `MatchResRecord`? The object we're composing here is related to Match Resolution. In the future, there could be match objects that are tailored for other contexts, like for metrics. 

I'm also wondering if it's worth distinguishing through a naming convention objects that are composed of aggregated data versus objects that map more directly to the database. Or is every object that doesn't have a `Dbo` on it assumed to be an aggregate?

kingcomma:
> For `MatchRecord`, would using a name that's more specific to the context would be helpful, like `MatchResRecord`?

I'm definitely not tied to one name or another, was just using `MatchRecord` as an example without `Aggregate` in the name. Totally open to whatever others feel works best. My default is always simpler and my assumption is that in the context of the match and collaboration APIs, "MatchRecord" will always mean the same as "MatchResRecord" or "MatchResolutionRecord". Other sub-systems may have their own classes/interfaces and may not even connect to the same data store.

> I'm also wondering if it's worth distinguishing through a naming convention objects that are composed of aggregated data versus objects that map more directly to the database. Or is every object that doesn't have a `Dbo` on it assumed to be an aggregate?

I was chewing on this too and thought it might be easiest to go with the precedent set by the Kickstarter article we've been referencing. They exclude "aggregate" from the name of the objects themselves (e.g., their subscription aggregate object is called just "Subscription"). I could also see a time at which the aggregation itself is moved to a database view, further blurring the line of what's in / out of the database.

I now see why that article took the time to [reiterate that naming is hard](https://kickstarter.engineering/event-sourcing-made-simple-4a2625113224#c87c) 😅

echappen:
> I could also see a time at which the aggregation itself is moved to a database view, further blurring the line of what's in / out of the database.

Very good point! Here's what I came up with (roughly models the [WIP API schema](https://github.com/18F/piipan/pull/2704/files#diff-e31f4347925f1f97d2f203a7860824f41433c7d0fc33993616f845d793ab7c41R40)):

**MatchRecord** - Top-level aggregate of match resolution data: `status`, etc. Has many `InitialData`; has many `Disposition`. 
**InitialData** - Snapshot of state-specific data when match is found: `case_id`, `participant_id`, etc. 
**Disposition** - Data each state provides in the resolution process: `final_disposition`, etc.

kingcomma:
👍 One more thought: what do you think about `Participant` instead of `InitialData`? It matches how we refer to the data in the match record mock ups ("SNAP participant information") and is maybe a bit more scrutable for developers (as well as API consumers)?

echappen:
`Participant` works for me, thanks!

match/src/Piipan.Match/Piipan.Match.Core/Models/Participant.cs
```diff
@@ -0,0 +1,24 @@
...
+using System.Linq;
+using Piipan.Match.Api.Models;
+using Newtonsoft.Json;
+
+namespace Piipan.Match.Core.Models
+{
+    /// <summary>
+    /// Initial participant data for each related state in a match
+    /// </summary>
+    public class Participant
```
kingcomma, line 12: *Question:* Is there a benefit to having this class implement `IParticipant`? It could help ensure consistency with the data that's written to the match record. An implication at this point is that `RecentBenefitIssuanceDates` wouldn't be available until it has been implemented in the upload/ETL process.

kcherukupalli:
Looks good .

kcherukupalli (approved):
Looks

PR #2728, opened 2022-02-28
Eth/hsts policy

edwintorres:
## What’s changing?

This change adds HSTS to the ConfigureServices in Startup.cs on the applications Dashboard and Query tool. This configuration sets max-age to 31536000 seconds. The HSTS can be configured in the middleware (Startup.cs) or in Azure Front Door. The middleware seems easier to configure based on the environment. Given that we have an Azure Front Door by application, there is no advantage of a single point of configuration with that option.

## Why?

Close NAC-36: HSTS Policy Not Enable. See AppSec report.

## This PR has:


bgfoshay (approved):
Looks good👍

PR #2740, opened 2022-03-04
Remove fields from matches; Use events in match reconciliation

echappen:
## What’s changing?

Removes `invalid` and `status` fields from the `matches` table since these are now on `match_res_events`
Use `MatchResAggregator` in `MatchEventService` to determine whether or not a new open match should be created when a present one is found.

## Why?

Relates to Jira NAC-165

This is the last of the Groundwork for the Match Resolution API. After this, we can build the Function App endpoints for the API. 

## This PR has:


echappen:
@kingcomma In particular, check out the `Reconcile` method in `MatchEventService` where `MatchResAggregator` is now being used. It felt slightly awkward to inject two more dependencies into the service, but testing didn't end up being that difficult.

kingcomma (approved):
👍 Looks good to me!

@echappen Can you run the necessary bits of IaC and add a comment confirming success before merging?

echappen:
@kingcomma IAC core database creation succeeded

PR #2824, opened 2022-03-09
Create and publish boilerplate Function App for Match Resolution in IAC

echappen:
## What’s changing?

- Adds ARM Template for Match Resolution Api Function App
- Deploys ARM Template in IAC script
- Adds Microsoft boilerplate for Match Resolution Api Function App
- Deploys App in IAC script
- Deploys App in build script for Match subsystem
- Adds App to sln file for Match subsystem
- Adds Initial developer documentation for Match Resolution Api

## Why?

Relates to Jira ticket NAC-389

Establishes initial scaffolding for creating and deploying a function app for Match Resolution in IAC. Does not include connecting app to a Virtual Network, or configuring app roles for database connections. I considered those tasks out of scope for this ticket, but if they should be, let me know. 

Tagging @bgfoshay as primary reviewer (if there's another Ventera Engineer who should be tagged please let me know)
Including @kingcomma since he's likely most knowledgeable about IAC requirements and considerations at this point

## This PR has:


echappen:
Ignore the names/routes/etc of the functions in the source code for now. I'll configure those in the next PR.

iac/create-resources.bash
```diff
@@ -376,6 +376,31 @@ main () {
...
+      collabDatabaseConnectionString="$collab_db_conn_str" \
+      cloudName="$CLOUD_NAME" \
+      states="$state_abbrs" \
+      coreResourceGroup="$RESOURCE_GROUP" \
+      eventHubName="$EVENT_HUB_NAME"
+
+  echo "Publish Match Resolution API Function App"
+  try_run "func azure functionapp publish ${MATCH_RES_FUNC_APP_NAME} --dotnet" 7 "../match/src/Piipan.Match/Piipan.Match.Res.Func.Api"
+
+  # TODO: integrate Match Resolution API with VNet
```
kingcomma, line 401: *Todo (non-blocking):* When this integration is done, we will want to update the Function app's ARM template with `"vnetRouteAllEnabled": true` to restrict outbound traffic (since the API is only internal for now).

match/build.bash
```diff
@@ -26,6 +26,16 @@ run_deploy () {
...
+
+  match_res_function_apps=($(get_resources $MATCH_RES_API_TAG "$MATCH_RESOURCE_GROUP"))
+
+  for app in "${match_res_function_apps[@]}"
+  do
+    echo "Publish ${app} to Azure Environment ${azure_env}"
+    pushd ./src/Piipan.Match/Piipan.Match.Res.Func.Api
+      func azure functionapp publish "$app" --dotnet
+    popd
+  done
```
kingcomma, line 38: *Question (non-blocking):* Would it be helpful for developers to easily build, test, deploy this API without also building, testing, deploying the match API?

This feels to me like the first clear instance where having a separate sub-system could be an advantage.

kingcomma (commented):
Just a small request and a couple non-blocking thoughts/questions from me.

Also, some things to think about when addressing naming / structuring in the following PRs:

* If the match resolution API stays a part of the match subsystem, something like `Piipan.Match.Func.ResolutionApi` is more in line with the [subsystem architecture ADR](https://github.com/18F/piipan/blob/dev/docs/adr/0018-standardize-subsystem-software-architecture.md#namespacesubsystemfuncweb).
* What do you think about following the precedent set by the metrics API of one `.cs` file per endpoint/function? E.g., `GetMatch.cs`.

bgfoshay:
Given this is IAC work, I do think it's worthwhile to have Ryan review.

I did have one chore related to the top-level solution file. My fault for not communicating to everyone that I added that file. Other than that, I think this looks pretty good.

jlachlan:
Can you set up the database function ? POTUS announced the whole of government digital strategy today... Maybe just add the function and link it to a like USPS zip code or USGS databases as those will be critical in future for address emergencies

jlachlan (approved):

jlachlan (approved):

match/build.bash
echappen: For CI purposes I would think No. For debugging purposes, maybe. I've just been commenting out the deploys I'm not working on whenever I debug, and it's been easy enough.

echappen:
> * If the match resolution API stays a part of the match subsystem, something like `Piipan.Match.Func.ResolutionApi` is more in line with the [subsystem architecture ADR](https://github.com/18F/piipan/blob/dev/docs/adr/0018-standardize-subsystem-software-architecture.md#namespacesubsystemfuncweb).
> * What do you think about following the precedent set by the metrics API of one `.cs` file per endpoint/function? E.g., `GetMatch.cs`.

Yep, these both make sense. Will update.

bgfoshay (commented):
Just some minor thoughts & comments from me. Only real request (non-blocking) I have is if you could add the project to the top level solution.

iac/arm-templates/function-match-res.json
```diff
@@ -0,0 +1,185 @@
...
+            "identity": {
+                "type": "SystemAssigned"
+            },
+            "dependsOn": [
+                "[resourceId('Microsoft.Storage/storageAccounts', variables('storageAccountName'))]"
+            ],
+            "properties": {
+                "serverFarmId": "[resourceId(parameters('coreResourceGroup'), 'Microsoft.Web/serverfarms', parameters('appServicePlanName'))]",
+                "httpsOnly": true,
+                "siteConfig": {
```
ryangeorgeharvey, line 80: Suggestion (non-blocking): Set minTlsVersion to be 1.2, as recommend by ARM Template BPA.

TA-000011
Latest TLS version should be used in your Function App
Recommendation: Upgrade to the latest TLS version
More information: https://github.com/Azure/template-analyzer/blob/main/docs/built-in-bpa-rules.md/#ta-000011-latest-tls-version-should-be-used-in-your-function-app

Result: Failed
Line: 80

Rules passed: 27/28

iac/arm-templates/function-match-res.json
```diff
@@ -0,0 +1,185 @@
+{
```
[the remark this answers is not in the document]
echappen: Good point—I went back and forth on this. I did a separate file for now, since I wasn't sure if more settings would need to diverge between the two apps as we make our way into new production environments. But if you think it's worth attempting now, I can certainly try.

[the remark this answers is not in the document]
bgfoshay: Any idea on the level of effort? If you think it would be a few hours or so, then I'd say it's worth taking on. But otherwise I would reserve the task for another time.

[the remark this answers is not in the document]
echappen: Getting it code-complete wouldn't take long, but it would need another IAC run to fully test it. I could break this into a separate issue and work on it if there's time left in this sprint.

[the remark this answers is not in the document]
bgfoshay: 👍I think that's a good plan.

match/build.bash
echappen: As for a new subsystem, I think it'll become a lot more clear to me whether Resolution should be its own subsystem as I work through the business logic of the endpoints. So far it's been tightly coupled enough with Match classes to keep it there.

iac/create-resources.bash
echappen: Created a new issue to return to at end of work

PR #2825, opened 2022-03-09
Hoist Function App Api Error builders into Piipan.Shared

echappen:
## What’s changing?

- Removes Api Error Builders from `Piipan.Match.Func.Api` and into `Piipan.Shared`
- Uses Api Error Builders in `Piipan.Match.Res.Func.Api`

## Why?

Relates to Jira Ticket NAC-165
Blocked by #2824 

Attempts to avoid duplicating Api Error building, since the new Match Resolution Api will need catch a lot of the same errors. 

I think there's a need for modularizing how we generate our API responses, because it'll help us conform to our chosen [JSON API schema](https://github.com/18F/piipan/blob/dev/docs/adr/0016-use-json-api-spec-for-apis.md#decision) across our API's. Before I continue down this path, I wanted to get feedback on the approach of adding it to `Piipan.Shared`. 

## This PR has:


echappen:
rough draft, will reconsider this approacj

PR #2826, opened 2022-03-09
Add match collab page

edwintorres:
## What’s changing?

Creates a blank Match page that validate the [MatchId for length 7 and only alphanumeric characters](https://github.com/18F/piipan/pull/2692#issuecomment-1045071033).

## Why?

Progress toward  NAC-132

- [x] New match lookup page is added to the Query Tool and Collaboration App
- [x] URL for the match lookup page includes the match ID
- [x] Page extracts match ID from url and verifies that match ID meets validation criteria
- [x] If the match ID provided is not valid, provide user with an appropriate error message
- [x] Include the same navigation header and footer that we provide on the "Search for SNAP Participants" page. Mockup below.

## This PR has:

## Anything else?

I'm recommending this two acceptance criteria should move to another issue, for next sprint, after we have the API endpoint.
- [ ] If the match ID provided is valid, page attempts to call the Collaboration API to request match details (this will always fail for now, that is okay)
- [ ] When the Collaboration API cannot be reached, show the user an appropriate error message - something like "Uh-oh. This page is unavailable right now. Try again in a few minutes or contact support for help."


query-tool/src/Piipan.QueryTool/Pages/Match.cshtml.cs
```diff
@@ -0,0 +1,102 @@
...
+
+                    Match = Matches.Find(item => item.MatchId == id);
+
+                    //Match ID length = 7 characters
+                    //Reference: https://github.com/18F/piipan/pull/2692#issuecomment-1045071033
+                    if(Match == null || id.Length != 7) {
+                        return RedirectToPage("Error", new { message = "MatchId not found" });
+                    }
+
+                    return Page();
```
bgfoshay, line 57: Question: once you incorporate the Match search, I assume if no match is found here you'll redirect to the NotFound page?

query-tool/src/Piipan.QueryTool/Models/MatchaData.cs
```diff
@@ -0,0 +1,150 @@
+using System;
+using System.Collections.Generic;
+using Newtonsoft.Json;
+using Piipan.Match.Api.Serializers;
+using Piipan.Participants.Api.Models;
+
```
bgfoshay, line 6: Nitpick: Should the file name be "MatchData.cs"?

query-tool/tests/Piipan.QueryTool.Tests/MatchTest.cs
```diff
@@ -0,0 +1,159 @@
...
+
+            // act
+            var result = Assert.IsType<RedirectToPageResult>(pageModel.OnGet("m1$23^45")).PageName;
+
+            // assert
+            Assert.Equal("Error", result);
+        }
+
+        [Fact]
+        public void TestValidMatchId()
```
bgfoshay, line 135: Todo (non-blocking): It might be worth added a test for TestValidMatchIdThatDoesNotExist that would test a valid formatted MatchId that doesn't exist redirects to NotFound page. I know this functionality isn't implemented yet though.

bgfoshay (approved):
I think it looks good. Just a couple minor thoughts

kingcomma:
*Question:* Since work on this page won't be picked up until next sprint, is there a sensible way we can avoid the dip in code coverage without too much effort?

query-tool/src/Piipan.QueryTool/Pages/Match.cshtml.cs
edwintorres: Yes, I was questioning me the same thing; what is  the best approach at this stage. First I created a NotFound page but later I changed to a general Error page. I have three scenarios right now (not valid, not found, empty matchid) on all of those I think the best way to manage it is to redirect to the search page, with a message in case of a match not found or not valid. 

The search page is still not created and I was not sure if I should create that page with this issue. Let me know if you think I should create the search page with this PR and I change the redirects.

bgfoshay: I like that approach (redirecting to the search page with a message). I think that's the cleanest and easiest to maintain going forward. I would not necessarily create that page with this issue though.

edwintorres:
> t

I forgot to delete the NotFound Page. I did that now and added a basic test for the Error message. I hope that help with the code coverage.

PR #2831, opened 2022-03-11
Configure Match Resolution App for Core Database; Integrate into Virtual Network

echappen:
## What’s changing?

- Configures Match Resolution Function App for access to the Core database in IAC
- Integrates app into Virtual Network in IAC

## Why?

Closes Jira ticket NAC-390

These items are needed for the business logic in the `GetMatch` endpoint for the Match Resolution API

Tagging @ryangeorgeharvey as primary reviewer
Tagging @kingcomma: I noticed other function apps used `az functionapp` to configure the VNet. Should I try to set those configs in the ARM template instead?

## This PR has:


iac/create-resources.bash
```diff
@@ -398,8 +398,18 @@ main () {
...
+  az functionapp vnet-integration add \
+    --name "$MATCH_RES_FUNC_APP_NAME" \
+    --resource-group "$MATCH_RESOURCE_GROUP" \
+    --subnet "$FUNC_SUBNET_NAME" \
+    --vnet "$VNET_ID"
+  az functionapp config appsettings set \
+    --name "$MATCH_RES_FUNC_APP_NAME" \
+    --resource-group "$MATCH_RESOURCE_GROUP" \
+    --settings \
+      WEBSITE_CONTENTOVERVNET=1 \
```
ryangeorgeharvey, line 411: Suggestion (non-blocking): Ideally, we can minimize use of bash scripting by maximizing use of ARM templates. I think the app settings (as well as the VNET Integration which you mentioned) should be moved to the function ARM Template, unless there is some reasoning for keeping them separate.

ryangeorgeharvey: Just adding on, I know your time might be limited this sprint and ideally we should make these similar changes to the other functions. Maybe we should make this a separate story to handle all functions?

I'm comfortable approving this, as it's in-line with how functions are configured now.

ryangeorgeharvey (approved):
Approving, but at a minimum we should log a story to refactor ARM Templates for the function apps.

iac/create-resources.bash
echappen: Thanks for pointing this out. In the interest of getting something useable to the FE team, I'll move forward with this and create a JIRA ticket to configure VNet for all our function apps in ARM templates.

kingcomma: Some historical context: We've run into some sequencing challenges when deploying Function apps that live in the virtual network. The network egress rules on the subnet can impact App Service's under-the-hood mechanisms when deploying. Publishing a Function App prior to adding it to the virtual network is an attempt to avoid these issues.

There are likely pure ARM template approaches that will also avoid the issue, but it will take some trial and error. We'll want to be especially careful to test any changes against both existing and new environments/resources.

PR #2848, opened 2022-03-16
First pass at Add Event API endpoint business logic

echappen:
## What’s changing?

- Adds `AddEventApi` endpoint
- Adds Unit tests for endpoint

## Why?

Relates to JIRA NAC 165

## This PR has:


echappen:
oops, forgot to push some cleanup code. Should be good now.

match/src/Piipan.Match/Piipan.Match.Func.ResolutionApi/AddEventApi.cs
```diff
@@ -0,0 +1,238 @@
...
+                    IMatchResEvent? lastEvent = matchResEvents.Result.LastOrDefault();
+                    if (lastEvent is IMatchResEvent)
+                    {
+                        var deltaObj = JsonConvert.DeserializeObject<AddEventRequestData>(lastEvent.Delta);
+                        if (lastEvent.ActorState == state &&
+                            deltaObj is AddEventRequestData &&
+                            JsonConvert.SerializeObject(deltaObj) == JsonConvert.SerializeObject(reqObj.Data))
+                        {
+                            string message = "Duplicate action not allowed";
+                            logger.LogInformation(message);
```
bgfoshay, line 86: Nitpick/Question: Should this be logged at error level?

match/src/Piipan.Match/Piipan.Match.Func.ResolutionApi/AddEventApi.cs
```diff
@@ -0,0 +1,238 @@
...
+                        Delta = JsonConvert.SerializeObject(reqObj.Data)
+                    };
+                    await _matchResEventDao.AddEvent(newEvent);
+                    // determine if match should be closed
+                    DetermineClosure(reqObj, match.Result, matchResEvents.Result, state);
+                    return new OkResult();
+                }
+                catch (StreamParserException ex)
+                {
+                    logger.LogInformation(ex.Message);
```
bgfoshay, line 104: Nitpick/Question: same as previous comment, should all these log statements in catch blocks be error/exception level logging?

match/src/Piipan.Match/Piipan.Match.Func.ResolutionApi/AddEventApi.cs
```diff
@@ -0,0 +1,238 @@
...
+            public async Task<IActionResult> AddEvent(
+                [HttpTrigger(AuthorizationLevel.Function, "patch", Route = "matches/{matchId}/disposition")] HttpRequest req,
+                string matchId,
+                ILogger logger)
+            {
+                LogRequest(logger, req);
+
+                try
+                {
+                    var reqObj = await _requestParser.Parse(req.Body);
```
bgfoshay, line 59: Question: Should some of this function's logic be moved out into a service class(es) to break it up? It's not terribly big and it's organized well. Just wondering if it would help at all with testability?

match/src/Piipan.Match/Piipan.Match.Func.ResolutionApi/AddEventApi.cs
```diff
@@ -0,0 +1,238 @@
...
+            private async void DetermineClosure(
+                AddEventRequest reqObj,
+                IMatchRecord match,
+                IEnumerable<IMatchResEvent> matchResEvents,
+                string state
+            )
+            {
+                if (String.IsNullOrEmpty(reqObj.Data.FinalDisposition)) return;
+
+                IEnumerable<IMatchResEvent> anotherDisposition = matchResEvents.Where(e => {
```
bgfoshay, line 133: Nitpick/Question: For readability purposes, would it be useful to pull out this Linq logic into another method like "RetrieveDispositionsFromOtherStates" or something similar?

Nitpick: Also, given the return is potentially a list of dispositions, the variable name maybe could be maybe be something like "otherDispositions"?

match/src/Piipan.Match/Piipan.Match.Func.ResolutionApi/AddEventApi.cs
```diff
@@ -0,0 +1,238 @@
...
+                    AddEventRequestData? delta = JsonConvert.DeserializeObject<AddEventRequestData>(e.Delta);
+                    return !String.IsNullOrEmpty(delta?.FinalDisposition) &&
+                        e.ActorState != state;
+                });
+                if (anotherDisposition.Any())
```
bgfoshay, line 138: Question: I may be reading wrong, but this looks like it creates a Close event if any other state has provided a final disposition. What if the initiating state issues the final disposition? Should a close event be created for that case?

match/tests/Piipan.Match.Core.IntegrationTests/DbFixture.cs
```diff
@@ -243,5 +247,36 @@ FROM match_res_events
 
             return result;
         }
+
+        public async Task<IEnumerable<IMatchResEvent>> GetEvents(string matchId)
+        {
+            const string sql = @"
```
bgfoshay, line 253: Thought (non-blocking): I wonder if we should consider moving sql statements like this out to a shared location. They're doubly defined in DbFixture and Dao classes. Database schema changes or query changes require changing it in both places.

match/tests/Piipan.Match.Func.ResolutionApi.IntegrationTests/AddEventIntegrationTests.cs
```diff
@@ -0,0 +1,234 @@
...
+                Delta = "{ \"status\": \"closed\" }"
+            };
+            InsertMatchResEvent(matchResEvent);
+
+            var mockRequest = MockRequest("{ \"data\": { \"invalid_match\": true } }");
+            var mockLogger = new Mock<ILogger>();
+            var api = Construct();
+
+            // Act
+            var response = await api.AddEvent(mockRequest.Object, matchId, mockLogger.Object) as UnauthorizedResult;
```
bgfoshay, line 146: Nitpick: If any result is returned other than UnauthorizedResult, the response variable will be null. The assert below will still fail but it'll be due to NullReferenceException. Should we explicitly test for the response type?

match/tests/Piipan.Match.Func.ResolutionApi.IntegrationTests/GetMatchApiIntegrationTests.cs
```diff
@@ -107,5 +117,54 @@ public async void GetMatch_ReturnsIfFound()
...
+            Insert(match);
+            // Act
+            var response = await api.GetMatch(mockRequest.Object, matchId, mockLogger) as JsonResult;
+
+            // Assert first request
+            Assert.Equal(200, response.StatusCode);
+            var resBody = response.Value as ApiResponse;
+            Assert.False(resBody.Data.Dispositions[0].InvalidMatch);
+
+            // Act again
```
bgfoshay, line 153: Thought: It might be worth a comment here regarding what the expectations are for acting again i.e. creating "invalid match" match event results in newly pulled Match request having InvalidMatch=true. Minor thing but I had to re-read a couple times to identify the difference from above.

bgfoshay (commented):
Praise: I had a couple of questions but overall this looks really good and I love the thorough amount of testing you provided.

match/src/Piipan.Match/Piipan.Match.Api/Models/AddEventRequest.cs
```diff
@@ -0,0 +1,47 @@
...
+
+    /// <summary>
+    /// Disposition data submitted by states
+    /// </summary>
+    /// <remarks>
+    /// Nullable properties along with NullValueHandling.Ignore allows
+    /// null key/values to be removed in JSON serialization.
+    /// This is important when inserting match resolution events into the database
+    /// and when merging events in match resolution aggregation.
+    /// Otherwise, null values would override previously present values when merging.
```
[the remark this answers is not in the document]
echappen: Good question—correct, events are immutable/insert-only. I should probably write a better comment for the scenario I'm trying to address here (or change my whole approach!):

I was running into an issue in the Add Event Api that went like this:
1. JSON request data is serialized into a C# object `AddEventRequest` for validation purposes
1. Multiple properties on the C# object end up as null, since a client may only send certain data and not others. (For example, a state may send a `protect_location` flag but not a `final_disposition`.) But the object initializes all the properties present. 
1. the C# object is serialized back into JSON in order to save it as the `delta` on a `match_res_event` database record. When this happens, the JSON string had keys present that were not sent by the client, but which are present on the C# Object. So I was ending up with a series of `match_res_events` with deltas that looked like this:
```
{ "protect_location": true, "final_disposition": null } // etc. for the other null values
{ "protect_location": null, "final_disposition": "foo" }
```
When these deltas are aggregated in the `MatchResAggregator` to get the current state, this would result in an object where subsequent null values would override any previously provided values:
```
{ "protect_location": null, "final_disposition": "foo" } // expected true for protect_location
```

Using `NullValueHandling = NullValueHandling.Ignore` in combination with nullable properties on `AddEventRequest` enabled the JSON serialization to leave off any null values. 

I built the aggregator with the idea that the delta would be saving _only_ the values that are changed, not all the properties, which makes the merging very straightforward. Maybe this is something that should be clarified or enforced somewhere in the codebase. 

As I'm writing this, I'm realizing that this would make it tricky to set a value back to null from something else. (Like if a state wanted to remove their `protect_location` flag). I don't see this being a common scenario, but it should be addressed. I think that would involve distinguishing between null values that are passed from the client from default object values, but am open to ideas.

match/tests/Piipan.Match.Core.IntegrationTests/DbFixture.cs
echappen: Very much agree, I plan to tackle this in my next PR.

match/tests/Piipan.Match.Func.ResolutionApi.IntegrationTests/AddEventIntegrationTests.cs
echappen: I noticed that in the tests too, but I struggled with writing a test against the response without first casting it to the object I needed. Maybe I'm missing a C# trick here?

match/src/Piipan.Match/Piipan.Match.Func.ResolutionApi/AddEventApi.cs
echappen: Good question—my understanding is that all states involved in a match must provide a final disposition in order for a match to be closed, regardless of whether they were the initiator. But I'll tag @cjhskippy to confirm that. 

Currently only 2 states can be involved in a match, so any other disposition (other than the acting state) would suffice for closure here. I could future-proof this a bit more and check against the number of states in the `States` list, in case the number of states involved ever changes.

match/tests/Piipan.Match.Func.ResolutionApi.IntegrationTests/AddEventIntegrationTests.cs
bgfoshay: I think maybe if you do an explicit cast, it'll fail at runtime for you which might be more revealing. 
e.g. something like
UnauthorizedResult response = (UnauthorizedResult)(await api.AddEvent(mockRequest.Object, matchId, mockLogger.Object)); 

But if that doesn't work, this is a minor comment and non-blocking.

match/src/Piipan.Match/Piipan.Match.Func.ResolutionApi/AddEventApi.cs
bgfoshay: Ok, that's my understanding as well. I misread the code. I was looking for logic like

if ( anotherDisposition.Any() && reqObj.Data.FinalDisposition ) -> then close it

But now I see you checked first thing in the function to verify the acting state entered a FinalDisposition.

match/src/Piipan.Match/Piipan.Match.Core/Parsers/AddEventRequestParser.cs
```diff
@@ -0,0 +1,73 @@
...
+                return request;
+            }
+            catch (ValidationException ex)
+            {
+                _logger.LogError($"Found validation errors while parsing: {ex.Errors.ToString()}");
+                throw ex;
+            }
+            catch (Exception ex)
+            {
+                throw new StreamParserException(ex.Message, ex);
```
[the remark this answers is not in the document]
echappen: We broke out validation errors here to add some extra logging, since FluentValidation returns `ex.Errors` that were helpful to see. All parsing errors ultimately return a `BadRequest`, so I think catching the rest in a `StreamParserException` is ok.

match/src/Piipan.Match/Piipan.Match.Func.ResolutionApi/AddEventApi.cs
echappen: I've just refactored this so it's more future proof and, hopefully, easier to read!

match/src/Piipan.Match/Piipan.Match.Core/Parsers/AddEventRequestParser.cs
```diff
@@ -0,0 +1,73 @@
...
+                if (!validationResult.IsValid)
+                {
+                    throw new ValidationException("request validation failed", validationResult.Errors);
+                }
+
+                return request;
+            }
+            catch (ValidationException ex)
+            {
+                _logger.LogError($"Found validation errors while parsing: {ex.Errors.ToString()}");
```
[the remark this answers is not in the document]
echappen: Yes, I think so!

match/tests/Piipan.Match.Func.ResolutionApi.IntegrationTests/AddEventIntegrationTests.cs
echappen: Ah ok, yes this does give a more helpful error message when it fails, where I can see what the actual result object is. Thanks!

match/src/Piipan.Match/Piipan.Match.Api/Models/AddEventRequest.cs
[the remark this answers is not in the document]
bgfoshay: Hmmm... I agree that even though it seems uncommon, we'll have to support the case of users explicitly setting a value to null vs not setting a property/value. In the interest of getting this initial capability in, we might consider splitting this case off into a separate issue.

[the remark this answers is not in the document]
echappen: 👍 I'll make another issue for it, hopefully this can be part of my cleanup work in the next sprint

echappen:
@bgfoshay Thanks so much for your thorough review!

Issue #2841, opened 2022-03-15
Make database fixtures reusable across test suites


echappen:
Thought: To your comment about DbFixture reuse, could we move DbFixture into Piipan.Shared.Tests project? Would probably need to genericize it a bit or templatize or use abstract class. Just a thought.

_Originally posted by @bgfoshay in https://github.com/18F/piipan/pull/2829#discussion_r824890690_

PR #2891, opened 2022-03-23
Use shared Db fixture for participants and match databases

echappen:
## What’s changing?

This creates a shared database fixture responsible for applying the schema, setup, and teardown for the `participants` and `matches` databases. 

## Why?

Relates to #2841 

## This PR has:

### DIFF
--- a/dashboard/src/Piipan.Dashboard/packages.lock.json
+++ b/dashboard/src/Piipan.Dashboard/packages.lock.json
@@ -73,6 +73,14 @@
           "System.Threading.Tasks.Extensions": "4.5.2"
         }
       },
+      "Dapper": {
+        "type": "Transitive",
+        "resolved": "2.0.123",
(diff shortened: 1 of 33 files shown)


echappen:
I ran into a few issues while doing this work. @bgfoshay would you mind gut checking this below, before I continue down my current path…

I’m thinking the type of functionality we’d want to DRY up between our test suites would be:
- Setting up database(s)
- Applying database schema
- Tearing down database(s)

These could be handled in a single fixture. Since each subsystem queries the data differently, the exact querying methods of the database(s) could be handled by each test suite, or it could be added to the shared fixture. I could go either way. 

We have multiple databases throughout Piipan. Having 1 fixture per database would work for all subsystem tests except for the Orchestrator, which needs access to both `participants` and `matches`. So I’d break it out like this:

- MetricsDbFixture
- MatchesDbFixture
- ParticipantsDbFixture (could hold enough state db’s needed)
- OrchestratorDbFixture

Also, related to what I just posted in Teams about the `Piipan.Shared` project, I’m thinking of creating a separate `SharedTests` project to hold these fixtures and possibly other testing-specific functionality.

Any thoughts or concerns before I continue?

echappen:
Implemented in new PR #2907

Issue #2833, opened 2022-03-11
Use Azure Middleware for logging in Function Apps


echappen:
I think there could be some future benefit to moving some of the request logging & exception handling logic into Azure Function middleware (https://github.com/Azure/azure-functions-dotnet-worker/tree/main/samples/CustomMiddleware).

_Originally posted by @bgfoshay in https://github.com/18F/piipan/pull/2829#pullrequestreview-907462875_

PR #2892, opened 2022-03-24
Api Request Logging Middleware

echappen:
## What’s changing?

- Creates Middleware for Api request logging
~~- Implements Middleware into Api's~~

## Why?

Relates to GH issue #2833 

## This PR has:

### DIFF
--- a/shared/src/Piipan.Shared/Logging/ApiRequestLoggingMiddleware.cs
+++ b/shared/src/Piipan.Shared/Logging/ApiRequestLoggingMiddleware.cs
@@ -1,0 +1,36 @@
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Logging;
+using System.Threading.Tasks;
+
+namespace Piipan.Shared.Logging
+{
+    public class ApiRequestLoggingMiddleware
+    {
+        private readonly RequestDelegate _next;
+
+        public ApiRequestLoggingMiddleware(RequestDelegate next)
+        {
+            _next = next;
+        }
+
+        public async Task InvokeAsync(HttpContext context,
+            ILogger<RequestLoggingMiddleware> logger)
+        {
+            logger.LogInformation("Executing request from user {User}", context.User?.Identity.Name);
+
+            HttpRequest request = context.Request;
+            string subscription = request.Headers["Ocp-Apim-Subscription-Name"];
+            if (subscription != null)
+            {
+                logger.LogInformation("Using APIM subscription {Subscription}", subscription);
+            }
+
+            string username = request.Headers["From"];
+            if (username != null)
+            {
+                logger.LogInformation("on behalf of {Username}", username);
+            }
+            await _next(context);
+        }
+    }
+}
--- a/shared/tests/Piipan.Shared.Tests/Logging/ApiRequestLoggingMiddlewareTests.cs
+++ b/shared/tests/Piipan.Shared.Tests/Logging/ApiRequestLoggingMiddlewareTests.cs
@@ -1,0 +1,54 @@
+using System;
+using System.Collections.Generic;
+using System.Security.Claims;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Primitives;
+using Moq;
+using Piipan.Shared.Claims;
+using Xunit;
+
+namespace Piipan.Shared.Logging.Tests
+{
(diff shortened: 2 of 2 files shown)


echappen:
@bgfoshay @kingcomma I think I'm missing how to configure middleware in an Azure Function App for .NET 3. The [example](https://github.com/Azure/azure-functions-dotnet-worker/blob/main/samples/CustomMiddleware/Program.cs) given configures it through a `Program` class, but our function apps are configured through a `Startup` class. (This example also shows [configuration](https://www.re-mark-able.net/blogs/2021/05/09/azure-functions-middleware.html) through a Program class). 

The more I google, it seems like there's [no official way](https://stackoverflow.com/a/69173672) to add middleware to a Function app for this .NET version, but there's a workaround that involves overriding the HttpRequestData. Maybe I'm missing something very obvious!

jlachlan:
Idk just friendly suggestion but saw AWS Mx optimizer that’s actually a log function and also there’s an adobe or oracle one but hasn’t been updated since December 2020 I think. 

If you keep the mx from allowing the middleware in my theory is it should work but idk about the azure set parameters especially with the subnet needing to be near S3 aws. 

The AWS Mx function to log looks like middle ware. Ur they claim it speed up traffic they are QOS in warehouse by doing what you guys just set up with vent and not switching between local domain formats …

Total guess. But you just have to prevent anyone from getting into your local network and mx is the 1990s OMG cable that got rid of crossover blah blah. 

Reply if you want link, was at desk when saw. Also check your normal.doc in the VM startup for office or 365 and make sure it matches the database orgin for font. 

The injections are using those stupid fake spaces or rags that get inserted to pdf covert on adobe unless you remove all the lines and stuff on scans. If you data isn’t moving from pdf to csv or xml pretend we had sales force forced and your not gonna deal with 300ppi scans with coffee stains on your OCR…

Apple is native to Segue I think I saw on doc early for parsing and Microsoft should auto configure the normal.doc to work with whatever your doing…

But the normal.doc is the css for the word doc… and excel and all those engines so things can get real nuts when someone is picking formats from the quick format tab bc that codes in the h1 h2 and bs hidden, so if people use edge for PDF not adobe or random stuff like that it becomes a nightmare bc people save the word doc to PDF by spooling printer not save as doc type etc.

Just waste time making sure the normal.docx in system 32 moves with the user. Bc it’s a hidden file but not locked. 

Maybe lock it… actually. Or track middle ware that references not local normal.doc in office. Depends on how the info is being saved worked on by user vs developer to say. If office 365 and azure and VS etc no CloudKit or adobe just lock all the system user files that transport like normal.doc that can end up creating Boolean VS !

jlachlan:
This all depends on if your using windows365 or Microsoft 365 or enterprise hybrid or full azure etc. 

I'd call Microsoft Enterprise and get custom thing for the following action for your on-site and payroll employees... they've probably duck rapped a million + lines as patches into every theoretical spaces possible then compress and transposed another 1M times. 

If your just hiring the website just figure out the paste as formatted thing and PDF with your CSS to avoid later hell... 

From what I've seen the middleware is sneaking in via fixed risk that can't be totally removed without just picking Microsoft and sending useful time making extension for edge and chrome that can silo off from full stack from mx alg to bs vba .net VS then other random middleware in conjunction doesn't work as they are little hacks or work around a that as a whole pile up. 

Just a user wasted a lot of time helping 55 year old employees with ridiculous nonsense from using too many different binary bases for no reason... 

Make Microsoft handle the middleware via the subnet and mx in same region as your AWS and just pay the money to be in most expensive region ...

I'm trying to figure out what the hell to set up on my end too, can you just buy a ton of windows 365 VMs already set up and utilize the extra hard disks as ram that are already SSD? 

I'd do that get rid of IPV6 reduce the middle ware processing, transposing, and switching between disk formats and processors and transport protocols... 

Almost like a local server and and NFS and a crossover cable and maybe a punch panel would would better than these VM that we don't notice are broke bc we don't have to fix them..

Maybe just give Amazon money to run that gx neural network code on your hard drives at night and just had road copy on windows 355 some how ??

Trying to help if I sound dumb it's bc I'm preventing another day of It work and not production

kingcomma:
@echappen I think this comes down us still being on .NET 3.1. Until we make the jump to .NET 6, we can't make use of isolated/out-of-process functions which are the only type that [support middleware](https://docs.microsoft.com/en-us/azure/azure-functions/dotnet-isolated-process-guide#middleware).

echappen:
Ah right, thanks for the reminder!

PR #2907, opened 2022-03-28
Add project for test db fixtures; use it in integration tests

### LINKED ISSUE #2841: Make database fixtures reusable across test suites

echappen:
## What’s changing?

- Adds `Piipan.Shared.TestFixtures` project
- Moves all test db fixture configs, setup and teardown throughout systems to this project
- Uses project in `Match`, `Metrics`, and `Participants` integration tests

## Why?

Closes #2841 

Tagging @bgfoshay as primary reviewer
Tagging @kingcomma for additional review

## This PR has:


match/tests/Piipan.Match.Func.ResolutionApi.IntegrationTests/DbFixture.cs
```diff
@@ -7,120 +7,16 @@
 using Piipan.Match.Api.Models;
 using Piipan.Match.Core.Models;
 using Piipan.Match.Core.DataAccessObjects;
+using Piipan.Shared.TestFixtures;
 
-namespace Piipan.Match.Core.IntegrationTests
+namespace Piipan.Match.Func.ResolutionApi.IntegrationTests
```
echappen, line 12: This is to fix an earlier error

echappen:
I considered making tests for the fixtures, but that seemed redundant since they should be covered by the integration tests. The `OrchestratorDbFixture` seems to be the culprit.

bgfoshay:
> @bgfoshay @echappen Any ideas for avoiding these test fixtures being included in code coverage reporting?

I admit I'm only somewhat familiar with CodeClimate but can we omit this project from coverage given it contains classes supporting? Other test classes are not included in coverage reports. Maybe it's because they fall under a "tests" folder or within a "test" project?

echappen:
Found the issue—I'm not actually using the fixture in the integration test 🤦‍♀️ . Should be an easy fix.

kingcomma:
@bgfoshay @echappen I think there may also be a higher-level question around our practices for including test project classes in code coverage reports.

For context, we use Coverlet for generating code coverage. Coverlet is automatically configured as part of `dotnet new xunit`. By default, Coverlet does not include test assemblies in its code coverage output. You have to explicitly tell Coverlet to include test assemblies [via a command line option](https://github.com/coverlet-coverage/coverlet/blob/master/Documentation/GlobalTool.md#coverlet-as-a-global-tool) if you want to alter that behavior.

CodeClimate just gets the files that Coverlet outputs and does not currently play a role in determining what is reported and what isn't.

My read of what's happening in this PR is that the DbFixtures were not previously included in code coverage reporting because they were exclusively part of the test projects. Now that they have been moved into `Piipan.Shared`, they are being included in the assemblies for that subsystem and thus included in code coverage.

This doesn't seem like a significant concern currently, but it is a change from our past (implicit) practices. It may be worth spending a small amount of time to explicitly lay out an agreed upon approach so we are consistent.

PR #2914, opened 2022-03-29
Fix: Await determining closure

echappen:
## What’s changing?

Fixing a bug introduced in #2848 where integration tests would intermittently fail due to suspected race condition when adding a new event and closing it right afterwards. The fix waits to determine closure until adding the event is done. 

## This PR has:


kingcomma (approved):
👍

PR #2925, opened 2022-04-01
Query Tool Page Updates

dborgen2:
## What’s changing?

We are updating the query tool per the new designs. Included in this PR is a Blazor component library that we use for the Query Tool and can potentially use for other pages as well.

New pa11y tests were added directly into the Cypress tests to validate the UI is accessible after validation. In two cases, post requests are send into Pa11y with the information we get from setting the input values in the Cypress test. After the post request comes back we can validate that the returned results are accessible as well.

A new Piipan.QueryTool.Client project was created. This is a Blazor Web Assembly project, which is required for the Blazor components to work. The Piipan.QueryTool project, which perhaps should be renamed Piipan.QueryTool.Server project in the future, references this Client project and renders the web assembly components as part of the page that is returned from the server side.

## This PR has:


dborgen2:
Closing this because we are breaking this out into multiple PRs.

PR #2931, opened 2022-04-04
Added host name and match URL to the participant match record

dborgen2:
## What’s changing?

The participant match record returned back from the Duplicate Participation API needs to return a MatchURL field, which should be populated with the associated Query Tool link.

## Why?

This satisfies NAC-229

## This PR has:

### DIFF
--- a/iac/create-resources.bash
+++ b/iac/create-resources.bash
@@ -31,6 +31,8 @@ set_constants () {
   ORCHESTRATOR_FUNC_APP_STORAGE_NAME=${PREFIX}storchestrator${ENV}
 
   PRIVATE_DNS_ZONE=$(private_dns_zone)
+
+  FRONT_DOOR_URI="https://$QUERY_TOOL_FRONTDOOR_NAME"$(front_door_host_suffix)
 }
 
 # Generate the storage account connection string for the corresponding
@@ -363,7 +365,8 @@ main () {
     --resource-group "$MATCH_RESOURCE_GROUP" \
     --settings \
       WEBSITE_CONTENTOVERVNET=1 \
-      WEBSITE_VNET_ROUTE_ALL=1
+      WEBSITE_VNET_ROUTE_ALL=1 \
+      QueryToolUrl="$FRONT_DOOR_URI"
 
   # Create an Active Directory app registration associated with the app.
   # Used by subsequent resources to configure auth
@@ -630,7 +633,7 @@ main () {
       idpClientId="$QUERY_TOOL_APP_IDP_CLIENT_ID" \
       aspNetCoreEnvironment="$PREFIX" \
       frontDoorId="$front_door_id" \
-      frontDoorUri="$front_door_uri"
+      frontDoorUri="$FRONT_DOOR_URI"
 
   echo "Integrating ${QUERY_TOOL_APP_NAME} into virtual network"
   az functionapp vnet-integration add \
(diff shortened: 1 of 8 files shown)


iac/create-resources.bash
```diff
@@ -31,6 +31,8 @@ set_constants () {
   ORCHESTRATOR_FUNC_APP_STORAGE_NAME=${PREFIX}storchestrator${ENV}
 
   PRIVATE_DNS_ZONE=$(private_dns_zone)
+
+  FRONT_DOOR_URI="https://$QUERY_TOOL_FRONTDOOR_NAME"$(front_door_host_suffix)
```
ryangeorgeharvey, line 35: Suggestion (non-blocking): Rename FRONT_DOOR_URI to QUERY_TOOL_URL to avoid confusion with Dashboard.

match/src/Piipan.Match/Piipan.Match.Core/Services/MatchEventService.cs
```diff
@@ -75,17 +76,25 @@ record = _recordBuilder
...
-            };
+                // No existing records
+                participantMatchRecord = new ParticipantMatch(match)
+                {
+                    MatchId = await _recordApi.AddRecord(record)
+                };
+            }
+            if (participantMatchRecord != null)
+            {
+                var queryToolUrl = Environment.GetEnvironmentVariable("QueryToolUri");
```
ryangeorgeharvey, line 94: Suggestion (non-blocking): Standardize naming. Recommend using Url versus Uri.

ryangeorgeharvey (commented):
Remove variable front_door_uri in create-resources.bash on line 603. You are declaring the variable at the top.

Consider standardizing naming of variable. Some places you use URI and others you use URL.

dborgen2:
> Remove variable front_door_uri in create-resources.bash on line 603. You are declaring the variable at the top.
> 
> Consider standardizing naming of variable. Some places you use URI and others you use URL.

Thanks, yea I had that removed but it must have got put back in somehow. Removing it now

iac/create-resources.bash
dborgen2: Good idea, updated in the latest checkin

match/src/Piipan.Match/Piipan.Match.Core/Services/MatchEventService.cs
dborgen2: Agree, fix is in

match/tests/Piipan.Match.Core.Tests/Services/MatchEventServiceTests.cs
```diff
@@ -174,6 +174,8 @@ public async void Resolve_AddsManyRecords()
         public async void Resolve_InsertsMatchId()
         {
             // Arrange
+            var queryToolUrl = "https://tts.test";
```
bgfoshay, line 177: Nitpick: You might consider pulling this up into overall test initialization if it's necessary for all tests in the suite.

bgfoshay (approved):
👍LGTM

match/tests/Piipan.Match.Core.Tests/Services/MatchEventServiceTests.cs
dborgen2: Yea, probably should. Just checked in an update to the test where everything gets set in the constructor.

ryangeorgeharvey (approved):
The IAC changes LGTM.

Issue #103, opened 2020-11-25
Incorporate queue for blob events


ryanhofdotgov:
Event grid trigger expects triggered functions to return in less than 30 seconds. This is not a problem for small file processing like Azure's image resizing tutorial, but will be for ETL’ing a 4M row CSV file.

As a follow up to #100, expand the pipeline from:
Blob → Event Grid → BulkUpload Function
to something like:
Blob → Event Grid → QueueWriter Function → Queue → BulkUpload Function

Other considerations that drive implementation approach:
- blob trigger uses polling mechanism (can be super slow, like 10 minutes to get a trigger) and can miss events under high load
- queue trigger approach requires blob publisher to also publish a queue message

End solution could be similar to what is described in [this blog post](https://medium.com/@adrianivan/processing-files-from-azure-blob-storage-using-event-grid-and-storage-queue-in-a-reactive-way-ef3ad30cd081).

PR #2985, opened 2022-04-06
NAC 273 - Add queue storage to handle bulk upload queues of files

edwintorres:
## What’s changing?

Added IaC code to create the queue storage, modification to the BulkUpload function, and QueueBulkUpload function was added. 

## Why?

NAC-273: Goal to establish a queueing approach to handling the initial Event Grid blob upload event that us currently handled by Bulk Upload function. The general motivation and approach is outlined in #103

## This PR has:


bgfoshay (commented):
Question: While I think this approach will work, I'm wondering whether the new Azure Function necessary? Could we instead add a new EventGrid subscription that directs/adds the event to the Storage Queue you've created? https://docs.microsoft.com/en-us/azure/event-grid/custom-event-to-queue-storage

PR #2990, opened 2022-04-07
Adjustments to DateRange bound handling

kingcomma:
## What’s changing?

- Use inclusive bounding for both upper and lower bounds when setting database value
- Check bound inclusion property when parsing database values to DateRange instances
- Use Dapper in ETL db fixture to remove custom bound parsing
- Additional unit tests for DateRangeListHandler for value parsing / setting

Only changes to `DateRange.cs` were small updates to the documentation comments, but Omnisharp picked up some whitespace edits that make it look like more.

## Why?

NAC-542

Sets the stage for using Npgsql's NodaTime extension when writing complex `daterange[]` values in a `COPY` statement

## This PR has:

## Anything else?

@kcherukupalli Tagging you for review since this builds on your recent work.

@bgfoshay Tagging you for awareness / review.

@edwintorres Tagging you because this includes a small update to the ETL integration tests `DbFixture`. I don't think this will conflict with any of your ETL work, but wanted to give you the heads up just in case it does.

### DIFF
--- a/etl/tests/Piipan.Etl.Func.BulkUpload.IntegrationTests/DbFixture.cs
+++ b/etl/tests/Piipan.Etl.Func.BulkUpload.IntegrationTests/DbFixture.cs
@@ -1,13 +1,10 @@
 using System;
 using System.Collections.Generic;
 using System.Threading;
-using Piipan.Etl.Func.BulkUpload.Models;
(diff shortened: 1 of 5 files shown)


kcherukupalli (approved):
Looks good to me.

PR #2993, opened 2022-04-08
Shared project & Query Tool Match Resolution logic

dborgen2:
The Query Tool now can call the Match Resolution API to pull back information on a match. 

## What’s changing?

- Match Resolution API added to AZ Active Directory, and can now be called from the Query Tool application
- Shared.API project created to make other projects that were dependent on the DateRange class more lightweight. This will allow them to be pulled into Blazor WebAssembly projects and remain small which will help download size and reduce security risks.
- Updated the match endpoint and the Duplication Participant API result to reflect the correct URL, which is /match/*match_id*, instead of /match?id=*match_id*.

## Why?

NAC-72, NAC-559, and NAC-582

## This PR has:


dborgen2:
Looks like I need to add a couple other test cases, but if you want to review early the code is all there

etl/tests/Piipan.Etl.Func.BulkUpload.IntegrationTests/DbFixture.cs
```diff
@@ -1,10 +1,13 @@
 using System;
 using System.Collections.Generic;
 using System.Threading;
-using Dapper;
-using Npgsql;
 using Piipan.Etl.Func.BulkUpload.Models;
+using Npgsql;
```
bgfoshay, line 5: Nitpick: Can you make sure using statements are sorted? Omnisharp in VSCode should do it automatically. There are Visual Studio formatting shortcuts as well.

dborgen2: Sure, I'll just turn the setting on to sort/remove unnecessary usings on save from now on. Thanks for the tip

match/src/Piipan.Match/Piipan.Match.Client/MatchResolutionClient.cs
```diff
@@ -0,0 +1,40 @@
...
+        /// <summary>
+        /// Sends a POST request to the /find_matches endpoint using the API client's configured base URL.
+        /// Includes in the request headers the identifier of the state initiating the request
+        /// </summary>
+        /// <param name="request">A collection of participants to attempt to find matches for</param>
+        /// <param name="initiatingState">The two character identifier of the state initiating the request</param>
+        /// <returns></returns>
+        public async Task<MatchResApiResponse> GetMatch(string matchId)
+        {
+            var (response, _) = await _apiClient.TryGetAsync<MatchResApiResponse>($"matches/{matchId}");
```
bgfoshay, line 36: Question: Just curious if there was a benefit to using TryGetAsync rather than GetAsync?

dborgen2: Yea, that was a bit weird. The Match Resolution API returns a 404 not found if you query on a match ID and it doesn't exist. I assumed that was correct and that I shouldn't change it. I needed to know the difference between a 404 and any other kind of error, so I created TryGetAsync to return a valid object (default<T>) along with the status code, in case different processing needed to be done depending on the status code. The response is dropped in GetAsync, so even though I could have caught the HttpRequestException I wouldn't have been able to check the status code if I would have bubbled the exception up to here

PR #2994, opened 2022-04-08
Use COPY for bulk inserting participant records

kingcomma:
## What’s changing?

- Use `COPY` (via `PostgreSQLCopyHelper`) to improve performance of writing large participant uploads to database
- Increase timeout duration for participant upload transaction to accomodate large uploads
- Introduce `Npgsql.NodaTime` plugin to handle setting the proper value for recent_benefit_issuance_dates field

## Why?

NAC-556

## This PR has:

## Anything else?

@bgfoshay I made some tweaks from the proof-of-concept branch we'd be working on. The primary change is using `PostgreSQLCopyHelper` to abstract away some of the complexity of writing individual rows/values. I think the benefit from a developer's perspective is worthwhile, with the tradeoff that we no longer are adding records within a loop where we could perform other per-participant operations (logging progress, etc).

Curious for your thoughts on this. My thinking is that our short-term extensions of the bulk upload process will not require that functionality, and, when we do need that functionality we'll likely want it decoupled from the database operations anyway. E.g., a discrete step that checks for invalid records, followed by another discrete step for adding valid records to the database.

### DIFF
--- a/etl/src/Piipan.Etl/Piipan.Etl.Func.BulkUpload/packages.lock.json
+++ b/etl/src/Piipan.Etl/Piipan.Etl.Func.BulkUpload/packages.lock.json
@@ -862,6 +862,31 @@
           "Newtonsoft.Json": "12.0.3"
         }
       },
+      "NodaTime": {
+        "type": "Transitive",
+        "resolved": "3.0.9",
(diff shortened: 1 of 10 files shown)


bgfoshay (approved):
Changes look good to me and abstracting some of the complexity away from developers would be good. This approach should still support column encryption. And if we want to inject per participant activities, we could always refactor the library to allow us to inject different row-level strategies.

kingcomma:
Per discussion, merging this PR and creating NAC-602 for the agreed upon follow-on work.

PR #2996, opened 2022-04-11
Fix script permissions and naming

ryangeorgeharvey:
## What’s changing?

Minor changes to `configure-defender-and-policy.bash`

## Why?

File was created on Windows, but will fail to execute in Unix based systems. Changing permissions to allow for execution.
Also fixing minor spelling mistake, and appending Subscription ID to the policy name, to avoid duplicate names.

## This PR has:


ryangeorgeharvey:
$ git commit -m "Fix script permissions and naming"
[developer-permissions b75e7eea] Fix script permissions and naming
1 file changed, 0 insertions(+), 0 deletions(-)
mode change 100644 => 100755 iac/configure-defender-and-policy.bash

ryangeorgeharvey:
Tagging @kingcomma for review.

iac/configure-defender-and-policy.bash
```diff
@@ -37,7 +37,7 @@ main () {
 
   echo "Assigning ${CIS_POLICY_SET_DEFINITION_NAME} to Subscription ${SUBSCRIPTION_ID}"
   az policy assignment create \
-    --name "CIS Microsoft Azure Foundations Benchmark v1.3.0" \
+    --name "CIS Microsoft Azure Foundations Benchmark v1.3.0 - ${SUBSCRIPTION_ID}" \
```
kingcomma, line 40: *Question (non-blocking):* What are the implications of a naming conflict? Would it cause an error, or is this to help with read-ability / find-ability?

kingcomma (approved):
👍 Just a non-blocking question that shouldn't hold up merging.

iac/configure-defender-and-policy.bash
ryangeorgeharvey: Naming is not unique, so it would not cause an error. This is mainly to help with filtering if needed and to match Azure naming, although Scope will show the differences. My original thinking was if we nested environments under subscriptions, we could name these per environment, but for now I left it at subscription ID.

I uploaded a screenshot to demonstrate that same naming does not cause errors. Also going to make one minor commit to change the naming to match Azure.  
image
image

ryangeorgeharvey: @kingcomma Thanks for calling this out, I had missed applying the subscription ID change when testing. However, I realized that using subscription ID in either format is causing issues. `The policy assignment name length must not exceed '64' characters.`

I'm going to go with my original thinking, and just use ENV name. This will also cover us, if we chose to share subscriptions at FNS.

ryangeorgeharvey: image

kingcomma: Got it, thanks for the explanation!

PR #3081, opened 2022-04-14
CJC-558

ccaprio97:
## What’s changing?

Making search_reason be a required field that throws an exception if one of the four valid reasons is not given. 
Valid reasons: Application, Recertification, New_Household_member, Other

## Why?

jira ticket 558

## This PR has:


ccaprio97:
generate-docs.bash not ran with included files right now. Current git setup deletes and re adds all lines. Will need to look into git config

match/src/Piipan.Match/Piipan.Match.Core/Parsers/OrchMatchRequestParser.cs
```diff
@@ -67,12 +67,11 @@ public async Task<OrchMatchRequest> Parse(Stream stream)
                         request.Data[i].SearchReason = request.Data[i].SearchReason.ToLower();
                         if (!Enum.IsDefined(typeof(ValidSearchReasons), request.Data[i].SearchReason))
                         {
-                            request.Data[i].SearchReason = ValidSearchReasons.other.ToString();
+                            throw new Exception("Submitted option for 'search_reason' is not valid. Valid options are 'application', 'recertification', 'new_household_member', or 'other'.");
```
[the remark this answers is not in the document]
ccaprio97: we could. I thought about going through and listing them but thought it made it more complicated. But ya we can change it to that if everyone thinks its better.

dborgen2:
We should probably add a test for throwing this exception in [OrchMatchRequestParserTests](https://github.com/18F/piipan/blob/bd836d2f326b9515f1adba99ba5fa516c82a28d4/match/tests/Piipan.Match.Core.Tests/Parsers/OrchMatchRequestParserTests.cs#L89)

ccaprio97:
build is failing and not sure why. Still looking into it

match/src/Piipan.Match/Piipan.Match.Core/Parsers/OrchMatchRequestParser.cs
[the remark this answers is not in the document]
dborgen2: Yea, I get that. I just wouldn't want this list to be out of date if we added another one, that would be hard to catch. I think it just ends up being those 3 lines above assuming the code I wrote works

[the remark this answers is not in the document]
ccaprio97: added tests that look for an exception in latest commit

match/tests/Piipan.Match.Core.Tests/Parsers/OrchMatchRequestParserTests.cs
```diff
@@ -117,12 +117,45 @@ public async Task ValidSearchReason(string searchReason, ValidSearchReasons expe
             Assert.Equal(expectedSearchReason.ToString(), request.Data[0].SearchReason);
         }
 
+        [Theory]
+        [InlineData("ASDF", ValidSearchReasons.other)]
+        [InlineData("New household member", ValidSearchReasons.other)]
+        [InlineData("", ValidSearchReasons.other)]
+        [InlineData(null, ValidSearchReasons.other)]
+        public async Task InvalidSearchReason(string searchReason, ValidSearchReasons expectedSearchReason)
```
dborgen2, line 125: We don't need this parameter for this test, since we are now expecting an error

docs/openapi/generated/duplicate-participation-api/openapi.yaml
```diff
@@ -49,7 +49,7 @@ paths:
                         description: Participant's state-specific case number
                       search_reason:
                         type: string
-                        description: 'User''s reason for running match query. Must be either ''Application'', ''Recertification'', or ''New Household Member'''
+                        description: 'The case action or other valid reason for performing the NAC search.  Valid options are ''application'', ''recertification'', ''new_household_member'', or ''other'''
```
dborgen2, line 52: I think search_reason should now be marked required after line 38

match/src/Piipan.Match/Piipan.Match.Api/Models/OrchMatchRequest.cs
```diff
@@ -32,6 +32,7 @@ public class RequestPerson
         public string CaseId { get; set; }
 
         [JsonProperty("search_reason",
+            Required = Required.Always,
```
bgfoshay, line 35: Will this cause backwards compatibility issues? We still haven't marked participant_id or case_id required. Have we double-checked with Andrew that this is ok?

ccaprio97: It did cause issues with tests and stuff that i had to fix. As for talking with andrew, no but he wrote the ticket that had this. 
image

bgfoshay: 👍. We just need to make sure to call this out when we make the release.

bgfoshay (approved):
👍LGTM

dborgen2 (approved):
👍

PR #3082, opened 2022-04-14
adding vulnerable individual to match api

ccaprio97:
## What’s changing?

Adding vulnerable_individual as an optional parameter to match api

## Why?

https://e-jira.fns.usda.net/jira-prod/browse/NAC-98

## This PR has:

API Doc updates to all params example, and match api 
comments on new code
tests for all cases and null case
generated docs

## Anything else?

the checking of the vulnerable individual is not in the validator because it is still accepted if it is not valid

the manual checking of yes no and blank is because this will most likely never change and therefore doesn't need an enum

### DIFF
--- a/docs/openapi/generated/duplicate-participation-api/openapi.yaml
+++ b/docs/openapi/generated/duplicate-participation-api/openapi.yaml
@@ -52,7 +52,7 @@ paths:
                         type: string
                         description: 'The case action or other valid reason for performing the NAC search.  Valid options are ''application'', ''recertification'', ''new_household_member'', or ''other'''
                       vulnerable_individual:
-                        type: string
+                        type: boolean
                         description: Location protection flag for vulnerable individuals. True values indicate that the individual’s location must be protected from disclosure to avoid harm to the individual. Apply the same protections to true and null values.
             examples:
               Single person:
--- a/docs/openapi/generated/duplicate-participation-api/openapi.yaml
+++ b/docs/openapi/generated/duplicate-participation-api/openapi.yaml
@@ -53,7 +53,7 @@ paths:
                         description: 'The case action or other valid reason for performing the NAC search.  Valid options are ''application'', ''recertification'', ''new_household_member'', or ''other'''
                       vulnerable_individual:
                         type: string
-                        description: Location protection flag for vulnerable individuals. 'yes' values indicate that the individual’s location must be protected from disclosure to avoid harm to the individual.
+                        description: Location protection flag for vulnerable individuals. True values indicate that the individual’s location must be protected from disclosure to avoid harm to the individual. Apply the same protections to true and null values.
             examples:
               Single person:
                 description: 'An example request to query a single person, with values for all fields'
--- a/docs/openapi/generated/duplicate-participation-api/openapi.yaml
+++ b/docs/openapi/generated/duplicate-participation-api/openapi.yaml
@@ -63,7 +63,7 @@ paths:
                       case_id: CaseNumber12345
                       participant_id: ParticipantId12345
                       search_reason: application
-                      vulnerable_individual: 'yes'
+                      vulnerable_individual: true
               Multiple persons:
                 description: A request with multiple persons
                 value:
--- a/docs/openapi/generated/duplicate-participation-api/openapi.yaml
+++ b/docs/openapi/generated/duplicate-participation-api/openapi.yaml
@@ -51,6 +51,9 @@ paths:
                       search_reason:
                         type: string
                         description: 'The case action or other valid reason for performing the NAC search.  Valid options are ''application'', ''recertification'', ''new_household_member'', or ''other'''
+                      vulnerable_individual:
+                        type: string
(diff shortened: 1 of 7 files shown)


docs/openapi/generated/duplicate-participation-api/openapi.yaml
```diff
@@ -51,6 +51,9 @@ paths:
                       search_reason:
                         type: string
                         description: 'The case action or other valid reason for performing the NAC search.  Valid options are ''application'', ''recertification'', ''new_household_member'', or ''other'''
+                      vulnerable_individual:
+                        type: string
+                        description: Location protection flag for vulnerable individuals. True values indicate that the individual’s location must be protected from disclosure to avoid harm to the individual. Apply the same protections to true and null values.
```
bgfoshay, line 56: If we're keeping this as a "string" value type, we should clarify "Yes/No" values rather than "True"

match/docs/openapi/schemas/match-query.yaml
```diff
@@ -52,6 +52,9 @@ Person:
     search_reason:
       type: string
       description: "The case action or other valid reason for performing the NAC search.  Valid options are 'application', 'recertification', 'new_household_member', or 'other'"
+    vulnerable_individual:
+      type: string
+      description: "Location protection flag for vulnerable individuals. True values indicate that the individual’s location must be protected from disclosure to avoid harm to the individual. Apply the same protections to true and null values."
```
bgfoshay, line 57: If we're keeping this as a "string" value type, we should clarify "Yes/No" values rather than "True"

ccaprio97: oh that's my bad. So the ticket didn't come with an api docs description so i took it from the other vulnerable_individual api doc that is held as a boolean so it came with that. 

Will update to yes/no

match/tests/Piipan.Match.Core.Tests/Parsers/OrchMatchRequestParserTests.cs
```diff
@@ -149,6 +149,76 @@ public async Task InvalidSearchReason(string searchReason)
             await Assert.ThrowsAsync<StreamParserException>(() => parser.Parse(BuildStream(JsonConvert.SerializeObject(orchMatchRequest))));
         }
 
+        [Theory]
```
bgfoshay, line 152: We probably should have tests with different capitalization as well i.e. "YES" or "Yes". As well as a test when the property is null.

match/src/Piipan.Match/Piipan.Match.Core/Parsers/OrchMatchRequestParser.cs
```diff
@@ -72,6 +72,17 @@ public async Task<OrchMatchRequest> Parse(Stream stream)
                             throw new Exception($"Submitted option for 'search_reason' is not valid. Valid options are {validOptionsString}.");
                         }
                     }
+                    //checking if vulnerable individual field is valid. 
+                    //If not set to empty but still accept the request
+                    if (request.Data[i].VulnerableIndividual != null)
+                    {
+                        request.Data[i].VulnerableIndividual = request.Data[i].VulnerableIndividual.ToLower();
+                        string[] vulnerableIndividualResponses = { "yes", "no", "" };
+                        if (!vulnerableIndividualResponses.Contains(request.Data[i].VulnerableIndividual))
```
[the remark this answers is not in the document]
ccaprio97: we can ask andrew but I went with not failing it because it is an optional field. Similar to how we did the search_reason before. 

I think andrew is out today not sure

match/tests/Piipan.Match.Core.Tests/Parsers/OrchMatchRequestParserTests.cs
ccaprio97: there are tests for null. they come later in the file. I can add capitalized tests if youd like

bgfoshay: ok. Yes, I'd add a capitalization test.

match/src/Piipan.Match/Piipan.Match.Core/Parsers/OrchMatchRequestParser.cs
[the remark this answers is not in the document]
bgfoshay: I thought we changed the functionality for search_reason so that it'd fail if inputted values didn't match one of the allowed enum values?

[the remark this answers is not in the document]
ccaprio97: we did this sprint. originally we allowed it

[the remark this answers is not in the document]
ccaprio97: we are going to go with failing it if its not a bool or null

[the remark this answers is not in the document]
bgfoshay: 👍

bgfoshay (approved):
👍LGTM

PR #3083, opened 2022-04-15
First round of updates for query tool and match search tool

dborgen2:
## What’s changing?

We are updating the query tool per the new designs.

New pa11y tests were added directly into the Cypress tests to validate the UI is accessible after validation. In several cases, post requests are send into Pa11y with the information we get from setting the input values in the Cypress test. After the post request comes back we can validate that the returned results are accessible as well.

A new Piipan.QueryTool.Client project was created. This is a Blazor Web Assembly project, which is required for the Blazor components to work. The Piipan.QueryTool project references this Client project and renders the web assembly components as part of the page that is returned from the server side.

## Why?

NAC-210 & NAC-448

## This PR has:


dborgen2:
Having some issues with circleci, but I will wait until Krishna merges his .net 6 PR anyway before merging to dev. Either way, the review for this can start if anyone has time

components/src/Piipan.Components/Forms/UsaForm.razor.cs
```diff
@@ -13,29 +13,44 @@ public partial class UsaForm
         [Inject] protected IJSRuntime JSRuntime { get; set; } = default!;
         private bool HasErrors => currentErrors?.Count() > 0;
         private EditContext editContext;
-        private bool ShowAlertBox { get; set; } = false;
+        private bool showAlertBox = false;
```
bgfoshay, line 16: Question: why the switch to lower-camelcase?

components/src/Piipan.Components/Forms/UsaInputSSN.razor.cs
```diff
@@ -11,6 +13,7 @@ namespace Piipan.Components.Forms
     public partial class UsaInputSSN
     {
         [Inject] protected IJSRuntime JSRuntime { get; set; } = default!;
+        Timer timer = new Timer();
```
bgfoshay, line 16: Question: There's a lot of logic around this timer and it's not immediately obvious what's going on. Do you think we could add some documentation in the code, or provide this timer with a more intention-revealing name, or move the logic out into a separate method call?

query-tool/tests/Piipan.QueryTool.Tests/MatchTest.cs
```diff
@@ -187,7 +148,15 @@ public async Task TestValidMatchIdThatDoesNotExist()
...
         private Mock<IMatchResolutionApi> SetupMatchResolutionApi()
         {
-            var apiReturnValue = MatchResApiResponseBuilder.Build();
+            var apiReturnValue = new MatchResApiResponse
+            {
+                Data = new MatchResRecord
+                {
+                    States = new string[] { "ea", "eb" },
+                    Initiator = "ea",
+                    MatchId = "m123456"
```
bgfoshay, line 157: Nitpick: Since this Id is critical and duplicated across many of the tests, should we pull it out into a constant or variable?

components/src/Piipan.Components/Forms/UsaInputSSN.razor.cs
dborgen2: Yes, good callout. I was hiding the SSN characters immediately after being typed, but Corey said I should wait about a second to hide them so a person could see they typed it in correctly. I'll definitely update it with some comments and choose a better name so it's more obvious.

query-tool/tests/Piipan.QueryTool.Tests/MatchTest.cs
dborgen2: Yea, probably a good idea. I'll do that

query-tool/tests/Piipan.QueryTool.Tests/IndexTest.cs
```diff
@@ -129,9 +131,8 @@ public async void MatchSetsResults()
                     }
                 });
 
-            var requestPii = new PiiRecord
+            var requestPii = new Client.Models.PiiRecord
```
bgfoshay, line 134: Nitpick: Could you have used a using statement?

dborgen2: Yes, good catch. There was a period of time when I was working where PiiRecord was in both the server and client projects (I was moving it), and had to preface it like this. But then after I deleted the one on the server this can be simplified. I'll fix it

query-tool/tests/Piipan.QueryTool.Tests/Extensions/ModelStateExtensions.cs
```diff
@@ -0,0 +1,30 @@
...
+            var results = new List<ValidationResult>();
+
+            if (!Validator.TryValidateObject(model, context, results, true))
+            {
+                pageModel.ModelState.Clear();
+                foreach (ValidationResult result in results)
+                {
+                    var key = (string.IsNullOrEmpty(boundProperty) ? "" : boundProperty + ".") +
+                        result.MemberNames.FirstOrDefault() ?? "";
+                    pageModel.ModelState.AddModelError(key, result.ErrorMessage);
```
bgfoshay, line 24: Question: I may find this after reviewing the rest of the code, but outside of the testing framework, what typically adds the errors to the ModelState? Is there a reason to add them differently in tests?

dborgen2: Sure, if you look at Index.cshtml.cs lines 32-33, you'll see this:

```
[BindProperty]
public PiiRecord Query { get; set; } = new PiiRecord();
```

This BindProperty is special for ASP.NET and will bind the submitted form to this property. When we are doing it through tests, we are simply creating an instance of the controller class and calling a function, thus the ASP.NET framework doesn't intervene

query-tool/src/Piipan.QueryTool/Pages/Match.cshtml.cs
```diff
@@ -34,36 +39,67 @@ public class MatchModel : BasePageModel
...
+                }
+                catch (Exception exception)
+                {
+                    _logger.LogError(exception, exception.Message);
+                    RequestErrors.Add(new("", "There was an error running your search. Please try again."));
+                }
+            }
+            else
+            {
+                var keys = ModelState.Keys;
```
bgfoshay, line 92: Question: Is it possible to Post with an invalid state after having already posted a valid request? i.e. AvailableMatches is populated already and now you're reposting with an invalid state? If so, should the AvailableMatches get cleared out?

dborgen2: If the controller state stayed consistent (i.e., were were storing into a session variable or something) this would be true. But each request creates its own controller instance so this is not possible. Thus the AvailableMatches is reset to null after each request. The only variable that is populated at the start of the request is the Query variable, since it is bound to the form.

bgfoshay: Gotcha. The logic in the valid condition threw me off as it was newing up the AvailableMatches

query-tool/tests/Piipan.QueryTool.Tests/Extensions/ModelStateExtensions.cs
bgfoshay: 👍

bgfoshay (approved):

query-tool/src/Piipan.QueryTool/Pages/Match.cshtml.cs
dborgen2: Yea, that was a bit weird. It was because the client does different things depending on if it's null, empty, or populated

PR #3084, opened 2022-04-15
Participant uploads generate unique lookup Id.

kcherukupalli:
## What’s changing?
Updating  a unique id(eTag) in Uploads table.  This would be used in future for any state inquries.

## Why?

#nac-412 

## This PR has:

- [ x] **Developer documentation** (_for new build/test/API changes or complex portions of the system_)

- [x ] **Changes to IAC** (_add any deployment steps that require manual intervention to the draft release notes_)


bgfoshay (approved):
👍

kingcomma:
Nice! A couple questions:

* Looks like this is confirmation that we can work with the `ETag` and not have to generate our own unique identifier?
* It looks like this gets a little ahead of NAC-412 and solves the approach for saving the identifier. What's the planned approach for the rest of the AC (returning the identifier in API responses, documentation, etc)?

bgfoshay:
@kingcomma Yes, this work confirms we can work with ETag as a unique identifier rather than generating our own. And it is a little ahead of NAC-412 because it takes care of the persistence/back-end aspect of the work. Krishna is addressing the rest of the AC in a subsequent PR (introducing BulkUploadResponse object that includes the upload id, updating Bulk Upload function to provide the response, and updating to the documentation)

kingcomma:
@bgfoshay Great! Thanks for the clarification. When working on the rest of the AC, we'll likely need to keep in mind that the upload process is asynchronous with the BulkUpload function. The response comes from the blob storage rest API and does not involve/invoke the BulkUpload function.

PR #3129, opened 2022-04-20
Component Library Testing & X-Frame-Options

dborgen2:
The blazor component library wasn't correctly getting picked up by the build.bash script. After adding it, we noticed that it would be an overall drop in code coverage, so I needed to add a few more tests.

A separate issue relating to X-Frame-Options: DENY was not getting picked up by the application. Moving this logic to add the header before the routing seems to fix it.

## What’s changing?

Updated the following

1. Updated build.bash to run the components project build/test. Nothing to deploy, so that was intentionally left out.
2. Added more tests to the component library so it passes code coverage
3. Updated the QueryTool test project to not reference the Components test project
4. Moved the X-Frame-Options above the routing so they get picked up.

## Why?

NAC-505 & NAC-630

## This PR has:


query-tool/src/Piipan.QueryTool/Startup.cs
```diff
@@ -147,15 +148,8 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
             app.UseEndpoints(endpoints =>
             {
                 endpoints.MapRazorPages();
-                endpoints.MapControllers();
```
bgfoshay, line 150: Question: Was this configuration removed intentionally?

dashboard/src/Piipan.Dashboard/Startup.cs
```diff
@@ -95,6 +96,12 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
             }
 
             app.UseHttpsRedirection();
+            app.Use(async (context, next) =>
+            {
+                context.Response.Headers.Add("X-Frame-Options", "DENY");
```
bgfoshay, line 101: Question: The placement of this differs from the QueryTool? Is that intentional?

query-tool/src/Piipan.QueryTool/Startup.cs
dborgen2: It was. I added MapControllers when I was under the impression we were using the MVC pattern when I was adding Blazor components, and then I realized we were not. It didn't hurt anything, but was unneccessary

dashboard/src/Piipan.Dashboard/Startup.cs
dborgen2: I think I updated it to be relatively in the same spot for both (after UseHttpsRedirection but before UseStaticFiles). That said, there is additional logic in query tool for handling error situations that we might also want to add to Dashboard. I didn't do that as part of this but if it's useful I could throw it in there for sure.

bgfoshay: It's a good point about the additional logic for the query tool error handling, but don't make that change now.

bgfoshay (approved):
👍

PR #3130, opened 2022-04-21
Refactor bulk insert functionality to isolate from ParticipantDao

kingcomma:
## What’s changing?

- Creates new interface for defining bulk insert functionality
- Moves PostgreSQL-specific copy functionality to implementation of new
  interface

## Why?

Partially addresses NAC-602. Will follow with a PR that updates the ADR for using COPY to callout this functionality as something that will need to be swapped if/when using a different database system.

## This PR has:


bgfoshay (approved):
This looks good! My only minor/non-blocking comment would be that now that we've encapsulated the database insertion logic behind an interface, whether there would be any value to moving ParticipantDaoTests & ParticipantServiceTests to unit tests (rather than integration) and mocking out IParticipantBulkInsertHandler.

kingcomma:
Great idea — I think there's definitely potential value in that our integration tests take a little extra effort to run (hunt down the command, time for docker to start up). Since the `ParticipantDao` still has the `GetParticipants` method that benefits from an integration test, it might be worth keeping as is for now. Hopefully we can also improve the integration test developer experience in the future (maybe something like a single docker compose file and associated `build.bash` command).

PR #3131, opened 2022-04-21
CJC-593

ccaprio97:
## What’s changing?

Updating API docs to accurately reflect current api 

## Why?

[_Link to specific issue (ex: Closes #123)_](https://e-jira.fns.usda.net/jira-prod/browse/NAC-593)

### DIFF
--- a/match/docs/openapi/schemas/match-query.yaml
+++ b/match/docs/openapi/schemas/match-query.yaml
@@ -90,6 +90,12 @@ MatchRequestExamples:
           participant_id: "ParticipantId12345"
           search_reason: "application"
           vulnerable_individual: true
+  SingleRequired:
+    description: "An example request to query a single person, with values for only required fields"
+    value:
+      data:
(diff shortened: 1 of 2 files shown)


ccaprio97:
@adabbraccio-ventera  the docs have been updated. Wanted you to approve this ticket on whether or not we need to update the widdershin templates so that the .md file also contains the other match api request examples that are in the openapi.yaml file

match/docs/openapi/schemas/match-query.yaml
```diff
@@ -16,6 +16,8 @@ MatchRequest:
       examples:
         Single person:
           $ref: '#/MatchRequestExamples/Single'
+        Single required person:
```
bgfoshay, line 19: Question/Nitpick: Interested to hear what others think? But are there other options we could consider for "Single required person"? This made me think it was referring to an individual being required as opposed to a request for a single person that contains only the required fields.

ccaprio97: Single person with required fields? Dont know how long it can be

bgfoshay (approved):

match/docs/openapi/schemas/match-query.yaml
adabbraccio-ventera: I agree with Bill. Your suggestion Conner was along the lines of what I was thinking.

PR #3138, opened 2022-04-22
Updates to Query Tool errors

dborgen2:
It was decided that if there were multiple errors with a certain field, we would show all of the errors instead of just the first one. In addition, the wording for when the user doesn't type in any letters is confusing, as it said "Normalized Last Name must be at least 1 character long." Now, we'll just say "Last Name must start with a letter" and update the requirements for that field appropriately.

## What’s changing?

- The UsaSSN attribute validation now returns all errors instead of just the first one
- The UsaName attribute now just simply checks to make sure the first non-space character they've typed is a letter
- The alert box was updated to show all errors on multiple list items when the returned error has more than one line
- The error above the field on a form was updated to show all errors instead of just the first one. Line breaks separate the errors.
- The SSN input field was widened by 10 more pixels so that the last digit wouldn't get cut off when the field had an error.

## Why?

Closes NAC-655

## This PR has:

### DIFF
--- a/components/src/Piipan.Components/Validation/ValidationConstants.cs
+++ b/components/src/Piipan.Components/Validation/ValidationConstants.cs
@@ -8,43 +8,48 @@
     public class ValidationConstants
     {
         /// <summary>
+        /// Used for a placeholder for the name of whatever field is being validated.
+        /// </summary>
+        public const string ValidationFieldPlaceholder = "@@@";
+
+        /// <summary>
         /// Used when a field is invalid, such as an invalid date
         /// </summary>
-        public const string InvalidMessage = "@@@ is invalid";
+        public const string InvalidMessage = $"{ValidationFieldPlaceholder} is invalid";
 
         /// <summary>
         /// Used for when a field is required, and is not filled in
         /// </summary>
-        public const string RequiredMessage = "@@@ is required";
+        public const string RequiredMessage = $"{ValidationFieldPlaceholder} is required";
 
         /// <summary>
         /// Used for social security number validation, when the SSN is not in the correct format
         /// </summary>
-        public const string SSNInvalidFormatMessage = "@@@ must have the form ###-##-####";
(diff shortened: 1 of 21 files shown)


components/tests/Piipan.Components.Tests/Validation/UsaNameAttributeTests.cs
```diff
@@ -46,7 +46,7 @@ public void CorrectErrorMessageForNormalizedNameTooShort()
 
             // Assert
             Assert.False(result);
-            Assert.Equal(ValidationConstants.NormalizedNameTooShortMessage, attribute.ErrorMessage);
+            Assert.Equal(ValidationConstants.MustStartWithALetter, attribute.ErrorMessage);
```
bgfoshay, line 49: Question/Nitpick: Should the test name change as well?

dborgen2: Yep, good point. Just updated

components/src/Piipan.Components/Validation/ValidationConstants.cs
```diff
@@ -40,7 +40,7 @@ public class ValidationConstants
         /// <summary>
         /// Used for names when validating they would have at least one character after normalization
         /// </summary>
-        public const string NormalizedNameTooShortMessage = "Normalized @@@ must be at least 1 character long.";
+        public const string MustStartWithALetter = "@@@ must start with a letter";
```
bgfoshay, line 43: Nitpick: You might consider pulling the placeholder/token out into it's own constant that can be reused, for example, in calls like @splitMessage.Replace("@@@", Label)

bgfoshay (approved):
I had one other nitpick you could consider but otherwise I think this looks good.

components/src/Piipan.Components/Validation/ValidationConstants.cs
dborgen2: Thanks, you're definitely right. Just updated the project to use a constant.

PR #3146, opened 2022-04-26
Updating package lock using a different node version

dborgen2:
@ryangeorgeharvey and @kcherukupalli noticed that they were getting npm install errors for the query tool project. I was not getting any issues until I downgraded to node version 14.19.1, which is the same version that they were on and what Azure CloudShell was using.

## What’s changing?

The package-lock file was rebuilt with node version 14.19.1, and no longer has any issues.

## Why?

Build issues described above

## This PR has:

### DIFF
--- a/docs/iac.md
+++ b/docs/iac.md
@@ -7,7 +7,7 @@
 - [.NET 6.0 SDK](https://dotnet.microsoft.com/download)
 - `bash` shell >= 4.1, `/dev/urandom` – included in macOS, Linux, Git for Windows, Azure Cloud Shell
 - `psql` client for PostgreSQL – included in Azure Cloud Shell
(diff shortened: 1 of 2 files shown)


bgfoshay (approved):
LGTM

Question (non-blocking): Do we have it documented in the repo anywhere that our expectation is that people run with Node v14.19.1? Might be worth documenting it as a pre-requisite so others don't run into this.

dborgen2:
> LGTM
> 
> Question (non-blocking): Do we have it documented in the repo anywhere that our expectation is that people run with Node v14.19.1? Might be worth documenting it as a pre-requisite so others don't run into this.

Good point. Updated the "Prerequisites" documentation with this note.

PR #3147, opened 2022-04-26
Front door ARM Template changes for Security Headers

dborgen2:
## What’s changing?

The Strict-Transport-Security and X-Frame-Options headers were sometimes missing from the query tool and dashboard application due to Front Door/EasyAuth quirks. The easiest and most straightforward solution is to make changes to Front Door itself to pass these header values in the response.

## Why?

Closes NAC-477, NAC-458, and NAC-505

## This PR has:

### DIFF
--- a/iac/arm-templates/front-door-app-service.json
+++ b/iac/arm-templates/front-door-app-service.json
@@ -65,7 +65,8 @@
     "variables": {
         "location": "global", /* front-door location must be global */
         "wafLocation": "global" /* waf location must be global */,
-        "eventhubName": "[concat(parameters('prefix'), '-evh-monitoring-', parameters('env'))]"
+        "eventhubName": "[concat(parameters('prefix'), '-evh-monitoring-', parameters('env'))]",
+        "securityHeaderRulesName": "SecurityHeaderRules"
     },
     "resources": [
         {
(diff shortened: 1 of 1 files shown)


edwintorres:
I'm curious, it was missing on all requests to the applications?

dborgen2:
> I'm curious, it was missing on all requests to the applications?

No, it was very strange. When the user was not logged in, they would be redirected to the login screen but before they were they were given a 200 response. That 200 response didn't have the "preload" portion of the header, and I think was sent by something other than the webapp itself. Perhaps front door or easy auth?

When the user was logged in, the header was identified twice, one with preload and one without. Again, something other than the web app was putting it in there. So I figured just override the value on the front door end, and that seemed to work.

iac/arm-templates/front-door-app-service.json
```diff
@@ -214,19 +214,43 @@
...
                                 }
                             },
+                            "rulesEngine": {
+                                "id": "[resourceId('Microsoft.Network/frontdoors/rulesengines', parameters('frontDoorName'), 'SecurityHeaderRules')]"
+                            },
+                            "enabledState": "Enabled"
+                        }
+                    },
+                    {
+                        "name": "HttpRedirect",
```
ryangeorgeharvey, line 236: Nitpick (non-blocking): Change name to lower camelCase 'httpRedirect' to match `routingRule1` naming convention.

iac/arm-templates/front-door-app-service.json
```diff
@@ -260,6 +284,40 @@
                     }
                 ]
             }
+        },
+        {
+            "type": "Microsoft.Network/frontdoors/rulesengines",
+            "apiVersion": "2020-05-01",
+            "name": "[concat(parameters('frontDoorName'), '/SecurityHeaderRules')]",
```
ryangeorgeharvey, line 291: Nitpick (non-blocking): Consider making `SecurityHeaderRules` an ARM variable, as you refer to it multiple places in the ARM.

ryangeorgeharvey (approved):
Minor code review, otherwise LGTM.

iac/arm-templates/front-door-app-service.json
dborgen2: Yea, good idea. Never used ARM template variables before but I think I updated it correctly

iac/arm-templates/front-door-app-service.json
dborgen2: 👍

PR #3150, opened 2022-04-28
NAC-159 Delete Blob After Processing

edwintorres:
## What’s changing?

When participants' records are uploaded, delete the blob file from the staging area. Additionally, BlobClientStream was improved. 

## Why?

NAC-159

## This PR has:


etl/src/Piipan.Etl/Piipan.Etl.Func.BulkUpload/BulkUpload.cs
```diff
@@ -55,20 +56,26 @@ public class BulkUpload
...
+                if(myQueueItem == null || myQueueItem.Length == 0){
+                    
                     log.LogError("No input stream was provided");
                 }
+                else {
+                    var blockBlobClient =  _blobStream.Parse(myQueueItem, log);
+
+                    Stream input = new System.IO.MemoryStream();
+
+                    var response = blockBlobClient.DownloadTo(input);
```
bgfoshay, line 68: Question: This approach did not work for me when I tried deploying it. Does it work for you once deployed? I'm curious if there's something I'm doing wrong. In order to get it to work, I had to use

blockBlobClient.OpenRead() instead of creating a new MemoryStream and calling DownloadTo.

etl/src/Piipan.Etl/Piipan.Etl.Func.BulkUpload/BulkUpload.cs
```diff
@@ -55,20 +56,26 @@ public class BulkUpload
...
+
+                    var blobProperties = blockBlobClient.GetProperties().Value;
+
+                    if (input != null)
+                    {
+                        var participants = _participantParser.Parse(input);
+                        await _participantApi.AddParticipants(participants,  blobProperties.ETag.ToString())
+                                .ContinueWith(t => _blobStream.DeleteBlobAfterProcessing(t, blockBlobClient, log));
```
bgfoshay, line 76: Question: What happens when the upload fails? Does the blob still get deleted? 

Related Question: Just want to verify, if/when the blob deletion fails, this does not affect the participants added to the database. i.e. they remain in the db and are not rolled back.

etl/src/Piipan.Etl/Piipan.Etl.Func.BulkUpload/BulkUpload.cs
edwintorres: Ok, I think I was able to replicate the error. When using DownloadTo a record is created in the `upload` table but not in `participants` when I change it to OpenReadAsync a record is created in both tables. I did not receive an Exception when using DownloadTo.  

Is this what are you experiencing? 

I'll change it to OpenReadAsync, but need to fix some UT.

bgfoshay: Yes, that's exactly the behavior I was seeing.

etl/src/Piipan.Etl/Piipan.Etl.Func.BulkUpload/BulkUpload.cs
```diff
@@ -55,20 +56,26 @@ public class BulkUpload
...
-                    // Can get here if Function does not have
-                    // permission to access blob URL
+                if(myQueueItem == null || myQueueItem.Length == 0){
+                    
                     log.LogError("No input stream was provided");
                 }
+                else {
+                    var blockBlobClient =  _blobStream.Parse(myQueueItem, log);
+
+                    Stream input = await blockBlobClient.OpenReadAsync();
```
edwintorres, line 66: @bgfoshay here is the last version with the OpenReadAsync which in my test solved the problem assign records to the participant table.

etl/src/Piipan.Etl/Piipan.Etl.Func.BulkUpload/Parsers/BlobClientStream.cs
```diff
@@ -30,54 +32,45 @@ public class BlobClientStream : IBlobClientStream
...
+        public virtual BlockBlobClient GetBlob(string blobName, string connectionString = "BlobStorageConnectionString") {                
 
-        public virtual BlockBlobClient GetBlob(string blobName) {                
-
-            return new BlockBlobClient(Environment.GetEnvironmentVariable("BlobStorageConnectionString"), "upload", blobName);
+            return new BlockBlobClient(Environment.GetEnvironmentVariable(connectionString), "upload", blobName);
         }
 
-        public virtual Response<BlobProperties> GetBlobProperties(BlockBlobClient blob) {                
+        public virtual StorageBlobCreatedEventData ParseEvents(string input){
```
bgfoshay, line 40: Question: Does this method need to be public?

edwintorres: No, I think I put all public virtual for the UTs and that one was not needed. I push the change in a minute.

bgfoshay (approved):
Looks good! 👍

PR #3172, opened 2022-05-03
Adding new API endpoint for getting all of the matches and corresponding View

dborgen2:
## What’s changing?

We need to add an API endpoint in the Match Resolution API Azure Function to get all of the existing matches. We then need to display these matches on a page in the Query Tool web application.

## Why?

Closes NAC-169

## This PR has:

### DIFF
--- a/match/src/Piipan.Match/Piipan.Match.Core/DataAccessObjects/MatchRecordDao.cs
+++ b/match/src/Piipan.Match/Piipan.Match.Core/DataAccessObjects/MatchRecordDao.cs
@@ -1,5 +1,3 @@
-using System.Collections.Generic;
-using System.Threading.Tasks;
 using Dapper;
 using Microsoft.Extensions.Logging;
 using Npgsql;
@@ -7,6 +5,8 @@ using Piipan.Match.Api.Models;
 using Piipan.Match.Core.Exceptions;
 using Piipan.Match.Core.Models;
 using Piipan.Shared.Database;
+using System.Collections.Generic;
+using System.Threading.Tasks;
 
 namespace Piipan.Match.Core.DataAccessObjects
 {
@@ -114,6 +114,26 @@ namespace Piipan.Match.Core.DataAccessObjects
             }
         }
 
+        public async Task<IEnumerable<IMatchRecord>> GetMatchesList()
+        {
+            const string sql = @"
+                SELECT
+                    match_id,
+                    created_at,
+                    initiator,
+                    states,
+                    hash,
+                    hash_type::text,
+                    input::jsonb,
+                    data::jsonb
+                FROM matches;";
+
+            using (var connection = await _dbConnectionFactory.Build())
+            {
+                return await connection.QueryAsync<MatchRecordDbo>(sql);
+            }
+        }
+
         /// <summary>
         /// Finds a Match Record by Match ID
         /// </summary>
@@ -141,7 +161,8 @@ namespace Piipan.Match.Core.DataAccessObjects
 
             using (var connection = await _dbConnectionFactory.Build())
             {
-                return await connection.QuerySingleAsync<MatchRecordDbo>(sql, new MatchRecordDbo {
+                return await connection.QuerySingleAsync<MatchRecordDbo>(sql, new MatchRecordDbo
+                {
                     MatchId = matchId
                 });
             }
--- a/match/src/Piipan.Match/Piipan.Match.Core/DataAccessObjects/MatchRecordDao.cs
+++ b/match/src/Piipan.Match/Piipan.Match.Core/DataAccessObjects/MatchRecordDao.cs
@@ -114,7 +114,7 @@ namespace Piipan.Match.Core.DataAccessObjects
             }
         }
 
-        public async Task<IEnumerable<IMatchRecord>> GetMatchesList()
+        public async Task<IEnumerable<IMatchRecord>> GetMatches()
         {
             const string sql = @"
                 SELECT
--- a/match/src/Piipan.Match/Piipan.Match.Func.ResolutionApi/GetMatchesListApi.cs
+++ b/match/src/Piipan.Match/Piipan.Match.Func.ResolutionApi/GetMatchesListApi.cs
@@ -1,0 +1,91 @@
+﻿using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.Azure.WebJobs;
+using Microsoft.Azure.WebJobs.Extensions.Http;
+using Microsoft.Extensions.Logging;
+using Piipan.Match.Api.Models.Resolution;
+using Piipan.Match.Core.Builders;
+using Piipan.Match.Core.DataAccessObjects;
+using Piipan.Match.Func.ResolutionApi.Models;
+using System;
+using System.Linq;
+using System.Net;
+using System.Threading.Tasks;
+
+namespace Piipan.Match.Func.ResolutionApi
+{
+
+    /// <summary>
+    /// Azure Function implementing Get Match endpoint for Match Resolution API
+    /// </summary>
+    public class GetMatchesListApi
+    {
+        private readonly IMatchRecordDao _matchRecordDao;
+        private readonly IMatchResEventDao _matchResEventDao;
+
+        private readonly IMatchResAggregator _matchResAggregator;
+
+        public GetMatchesListApi(
+            IMatchRecordDao matchRecordDao,
+            IMatchResEventDao matchResEventDao,
+            IMatchResAggregator matchResAggregator)
+        {
+            _matchRecordDao = matchRecordDao;
+            _matchResEventDao = matchResEventDao;
+            _matchResAggregator = matchResAggregator;
+        }
+
+        [FunctionName("GetMatchesList")]
+        public async Task<IActionResult> GetMatchesList(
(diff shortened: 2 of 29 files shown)


match/src/Piipan.Match/Piipan.Match.Core/DataAccessObjects/MatchRecordDao.cs
```diff
@@ -114,6 +114,26 @@ FROM matches
             }
         }
 
+        public async Task<IEnumerable<IMatchRecord>> GetMatchesList()
```
bgfoshay, line 117: Thought: I think GetMatches() is probably sufficient. Is using "List" might be redundant?

match/src/Piipan.Match/Piipan.Match.Func.ResolutionApi/GetMatchesListApi.cs
```diff
@@ -0,0 +1,91 @@
...
+            IMatchResAggregator matchResAggregator)
+        {
+            _matchRecordDao = matchRecordDao;
+            _matchResEventDao = matchResEventDao;
+            _matchResAggregator = matchResAggregator;
+        }
+
+        [FunctionName("GetMatchesList")]
+        public async Task<IActionResult> GetMatchesList(
+            [HttpTrigger(AuthorizationLevel.Anonymous, "get", Route = "list")] HttpRequest req,
```
bgfoshay, line 40: Question: based on REST API conventions should this route be "matches"? i.e. GET /matches for all matches, GET /matches/{id} for a specific match

match/src/Piipan.Match/Piipan.Match.Core/DataAccessObjects/MatchRecordDao.cs
```diff
@@ -114,6 +114,26 @@ FROM matches
...
+                SELECT
+                    match_id,
+                    created_at,
+                    initiator,
+                    states,
+                    hash,
+                    hash_type::text,
+                    input::jsonb,
+                    data::jsonb
+                FROM matches;";
```
bgfoshay, line 129: Question: Should we be concerned with the performance of a query like this? I honestly wouldn't anticipate millions of match records but it is slightly dangerous if we were to pull that many records into memory.

match/src/Piipan.Match/Piipan.Match.Func.ResolutionApi/GetMatchesListApi.cs
```diff
@@ -0,0 +1,91 @@
...
+            ILogger logger)
+        {
+            LogRequest(logger, req);
+            try
+            {
+                var matches = await _matchRecordDao.GetMatchesList();
```
bgfoshay, line 46: Question: Does Dapper work with Linq-To-Entities? With Entity Framework and Linq-To-Entities, this query would only every pull in match ids into memory. But otherwise, it'll pull the entire set of match records into memory. Given we're only interested in the match ids, would there be any benefit to rewriting the method/query to GetListOfAllMatchIds() ?

match/src/Piipan.Match/Piipan.Match.Func.ResolutionApi/GetMatchesListApi.cs
```diff
@@ -0,0 +1,91 @@
...
+            IMatchResEventDao matchResEventDao,
+            IMatchResAggregator matchResAggregator)
+        {
+            _matchRecordDao = matchRecordDao;
+            _matchResEventDao = matchResEventDao;
+            _matchResAggregator = matchResAggregator;
+        }
+
+        [FunctionName("GetMatchesList")]
+        public async Task<IActionResult> GetMatchesList(
```
bgfoshay, line 39: Thought: My fault for not following up on this with Andrew/Skippy earlier, but we should discuss whether pagination is needed with this method.

match/src/Piipan.Match/Piipan.Match.Func.ResolutionApi/GetMatchesListApi.cs
```diff
@@ -0,0 +1,91 @@
...
+            }
+            catch (Exception ex)
+            {
+                logger.LogInformation(ex.Message);
+                return InternalServerErrorResponse(ex);
+            }
+        }
+
+        private void LogRequest(ILogger logger, HttpRequest request)
+        {
```
bgfoshay, line 60: Thought (non-blocking): We should note that the methods like this in the JIRA ticket for Azure Function middleware. Eventually it'd be nice to refactor and pull them into a Shared middleware library.

match/tests/Piipan.Match.Core.IntegrationTests/MatchRecordDaoTests.cs
```diff
@@ -292,5 +291,40 @@ public void GetRecordByMatchId_ThrowsExceptionIfNotFound()
                 Assert.ThrowsAsync<System.InvalidOperationException>(() => dao.GetRecordByMatchId(id));
             }
         }
+
```
bgfoshay, line 294: Question: Do we need a test if no records are found?

match/tests/Piipan.Match.Core.IntegrationTests/MatchResEventDaoTests.cs
```diff
@@ -275,5 +271,66 @@ public async void GetEvents_ReturnsExpectedEventValues()
                 conn.Close();
             }
         }
+
```
bgfoshay, line 274: Question: same comment as the one I made for MatchRecordDao, should we consider tests for no records found or for an exception?

bgfoshay (commented):
Overall: Nice work on this! Definitely a lot of good work here. I had a couple minor comments on adding a few more tests. I had two larger concerns

1) Should be easy to fix but I think we need to update the route to "matches"
2) I'm interested in your thoughts but the more I think about it, the more I think it might be worth checking with Andrew/Skippy about the potential performance concerns of this endpoint and whether we need to incorporate pagination.

match/src/Piipan.Match/Piipan.Match.Func.ResolutionApi/GetMatchesListApi.cs
dborgen2: You're right. I'll rename it to /matches

match/src/Piipan.Match/Piipan.Match.Func.ResolutionApi/GetMatchesListApi.cs
dborgen2: No pagination needed (yet) after talking with Andrew. Something to keep in mind for the future

match/src/Piipan.Match/Piipan.Match.Core/DataAccessObjects/MatchRecordDao.cs
dborgen2: No pagination needed (yet) after talking with Andrew. Something to keep in mind for the future.

I definitely agree with you though, this could get very costly.

match/src/Piipan.Match/Piipan.Match.Func.ResolutionApi/GetMatchesListApi.cs
bgfoshay: You're right. I missed that it's used as first parameter to MatchResAggregator.Build method

match/tests/Piipan.Match.Core.IntegrationTests/MatchRecordDaoTests.cs
dborgen2: Yes, I think adding a test for no records is probably a good idea. Unlike GetRecordByMatchId, an exception will not be thrown if there is no record, so I don't think we can test for that here. I do test for an exception in the GetMatchesApiTests.cs unit test where we mock up the database call, which should be sufficient here.

match/tests/Piipan.Match.Core.IntegrationTests/MatchResEventDaoTests.cs
dborgen2: I added a test for no records found, but for the same reason as above no need to check for an exception with this test

dborgen2:
> Overall: Nice work on this! Definitely a lot of good work here. I had a couple minor comments on adding a few more tests. I had two larger concerns
> 
> 1. Should be easy to fix but I think we need to update the route to "matches"
> 2. I'm interested in your thoughts but the more I think about it, the more I think it might be worth checking with Andrew/Skippy about the potential performance concerns of this endpoint and whether we need to incorporate pagination.

Awesome, thanks for reviewing it. I think I've got everything covered that you mentioned. I definitely agree that we should consider paging in the future, but from my conversation with Andrew we didn't need to worry about it for this item.

match/tests/Piipan.Match.Core.IntegrationTests/MatchRecordDaoTests.cs
bgfoshay: 👍

match/tests/Piipan.Match.Core.IntegrationTests/MatchResEventDaoTests.cs
bgfoshay: 👍

bgfoshay (approved):
Sorry it took me so long to get to this. LGTM

PR #3198, opened 2022-05-06
CJC-416

ccaprio97:
## What’s changing?

Changing status in upload table to 'Complete' on successful upload

## Why?

NAC-416

## This PR has:

## Anything else?

Unaware of any documentation to update because all affected code is internal. If some needs to be added where would that go?

### DIFF
--- a/participants/src/Piipan.Participants/Piipan.Participants.Core/Services/ParticipantService.cs
+++ b/participants/src/Piipan.Participants/Piipan.Participants.Core/Services/ParticipantService.cs
@@ -71,6 +71,7 @@ namespace Piipan.Participants.Core.Services
                 }
             }
             catch (Exception ex) {
+                _logger.LogError(ex, ex.Message);
                 await _uploadDao.UpdateUploadStatus(upload, UploadStatuses.FAILED.ToString());
             }
         }
--- a/participants/src/Piipan.Participants/Piipan.Participants.Core/Services/ParticipantService.cs
+++ b/participants/src/Piipan.Participants/Piipan.Participants.Core/Services/ParticipantService.cs
@@ -63,6 +63,7 @@ namespace Piipan.Participants.Core.Services
                 });
 
                 await _participantDao.AddParticipants(participantDbos);
+                await _uploadDao.UpdateUploadStatus(upload);
                 scope.Complete();
             }
         }
--- a/participants/src/Piipan.Participants/Piipan.Participants.Core/Services/ParticipantService.cs
+++ b/participants/src/Piipan.Participants/Piipan.Participants.Core/Services/ParticipantService.cs
@@ -66,12 +66,12 @@ namespace Piipan.Participants.Core.Services
                     });
 
                     await _participantDao.AddParticipants(participantDbos);
-                    await _uploadDao.UpdateUploadStatus(upload, UploadStatuses.Complete.ToString());
+                    await _uploadDao.UpdateUploadStatus(upload, UploadStatuses.COMPLETE.ToString());
                     scope.Complete();
                 }
             }
             catch (Exception ex) {
-                await _uploadDao.UpdateUploadStatus(upload, UploadStatuses.Failed.ToString());
+                await _uploadDao.UpdateUploadStatus(upload, UploadStatuses.FAILED.ToString());
             }
         }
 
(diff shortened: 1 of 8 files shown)


bgfoshay (changes requested):
I'm not sure this is the right approach. I think we need to move the initial record creation outside the transaction. And it should still set to "Uploading". Then we'll need a new statement, at the end of the transaction (after participants are finished being copied over) to update the upload record's status to "Complete" (or "Failed" when appropriate). Call me if you want to discuss.

ccaprio97:
changed the original place back to 'uploading' and added new method after addParticipants. Returns the Id even though it doesnt need to. will probably need to have the id being returned when we eventually check for failed or successful plus it doesnt hurt

participants/src/Piipan.Participants/Piipan.Participants.Core/Services/ParticipantService.cs
```diff
@@ -63,6 +63,7 @@ public async Task AddParticipants(IEnumerable<IParticipant> participants,string
                 });
 
                 await _participantDao.AddParticipants(participantDbos);
+                await _uploadDao.UpdateUploadStatus(upload);
```
bgfoshay, line 66: Issue: I think you also need to move the following line outside/above the transaction scope.

participants/src/Piipan.Participants/Piipan.Participants.Core/Services/ParticipantService.cs
```diff
@@ -50,20 +51,27 @@ public async Task AddParticipants(IEnumerable<IParticipant> participants,string
...
+                        UploadId = upload.Id
+                    });
+
+                    await _participantDao.AddParticipants(participantDbos);
+                    await _uploadDao.UpdateUploadStatus(upload, UploadStatuses.Complete.ToString());
+                    scope.Complete();
+                }
+            }
+            catch (Exception ex) {
+                await _uploadDao.UpdateUploadStatus(upload, UploadStatuses.Failed.ToString());
```
bgfoshay, line 74: Thought: You should log the error here so that if one occurs, we can see what happened.

bgfoshay: I think it should be a simple change and if so, we should log it as part of this PR. Unless you have concerns with doing so?

participants/src/Piipan.Participants/Piipan.Participants.Core/Services/ParticipantService.cs
```diff
@@ -50,20 +51,28 @@ public async Task AddParticipants(IEnumerable<IParticipant> participants,string
...
             // an increased time out duration to avoid System.TimeoutException
-            using (TransactionScope scope = new TransactionScope(
-                TransactionScopeOption.Required,
-                TimeSpan.FromSeconds(600),
-                TransactionScopeAsyncFlowOption.Enabled))
-            {
-                var upload = await _uploadDao.AddUpload(uploadIdentifier);
-
-                var participantDbos = participants.Select(p => new ParticipantDbo(p)
+            var upload = await _uploadDao.AddUpload(uploadIdentifier);
```
bgfoshay, line 54: Thought: Do we have a unit test case for upload record created but participants failed?

ccaprio97: check AfterException_AddParticipantsRollsTranactionBack

participants/tests/Piipan.Participants.Core.IntegrationTests/ParticipantServiceTests.cs
```diff
@@ -84,7 +84,7 @@ public async void AfterException_AddParticipantsRollsTranactionBack(int nPartici
                 catch (Exception)
                 {
                     long expectedNewUploadId = ++lastUploadId;
-                    long actualLastUploadId = GetLastUploadId();
+                    long actualLastUploadId = GetLastSuccessfulUploadId();
```
bgfoshay, line 87: Thought: Aside from verifying the most recent successful uploadId, I think you should verify that the latest upload has a status = FAILED.

shared/src/Piipan.Shared.TestFixtures/ParticipantsDbFixture.cs
```diff
@@ -129,5 +129,35 @@ public Int64 GetLastUploadId()
...
+            {
+                conn.ConnectionString = ConnectionString;
+                conn.Open();
+                result = conn.ExecuteScalar<Int64>("SELECT MAX(id) FROM uploads WHERE status='COMPLETE'");
+                conn.Close();
+            }
+            return result;
+        }
+
+        public Int64 GetLastFailedUploadId()
```
bgfoshay, line 148: Question: can you change/consolidate GetLastSuccessfulUploadId/GetLastFailedUploadId -> GetLatestUploadByStatus(string desiredStatus)?

bgfoshay (approved):
👍Thanks for addressing all of my comments!

PR #3206, opened 2022-05-10
Use increased database timeout settings for ETL

kingcomma:
## What’s changing?

ETL process requires increased timeout durations to avoid exceptions during long-running bulk load operations. These settings had previously been mistakenly applied to the connection string used by the Orchestrator function app rather than the ETL function app.

I re-ran IaC to confirm the ETL app is configured with the proper settings and the Orchestrator app is reverted to the default settings it used previously.

## Why?

Fixing misconfiguration introduced by #2994.

## This PR has:

- ~[ ] **Commented source code** (_required on public classes and methods, and elsewhere as appropriate_)~
- ~[ ] **Automated unit tests** (_to maintain or increase level of code coverage_)~


bgfoshay (approved):
👍

PR #3209, opened 2022-05-10
Disabling match responses for certain states

dborgen2:
## What’s changing?

We need a way to disable returning back matches to caller. Two possible options emerged: Put it in a configuration file, or a database table. We also needed to consider the possibility that we might want to disable it per state, instead of disabling it as a whole.

This PR shows a proof of concept of how we would do it in a configuration file. We already had the states.csv anyway, so I propose adding a third column to the CSV, which is set to TRUE if the state should be disabled from returning back matches, or FALSE if the state should process as normal. If it makes more sense to switch this around and make it an "enabled" flag instead of a "disabled" flag, we would simply negate these values.

In my opinion, the configuration file makes the most sense for a couple of reasons:

1. We already had the states.csv, so this is simply one more column on there
2. Most importantly, the speed of the check to see if we should be disabling matches would be faster, since a call to the database is a round-trip call that we would not have to make if it was in a configuration file.
3. No need to add a new table or SQL calls

It's also important to note that you can alter this on the fly in Azure by changing the configuration in the Orchestrator function, and if you do then the Azure Function will restart and pick up the new value.

The main drawback to this approach is that if it is changed in Azure and not updated in States.csv, then next time we deploy the configuration will get overridden.

If the team agrees on this approach, I'll add to the documentation and test suite

## Why?

Research for NAC-627

## This PR has:

- [] **Automated unit tests** (_to maintain or increase level of code coverage_)


match/src/Piipan.Match/Piipan.Match.Func.Api/Api.cs
```diff
@@ -63,6 +64,15 @@ public class MatchApi
                 var response = await _matchApi.FindMatches(request, initiatingState);
                 response = await _matchEventService.ResolveMatches(request, response, initiatingState);
 
+                // If our initiating state is in the disabled states list, simply return that there were no matches.
+                string disabledMatchingStates = Environment.GetEnvironmentVariable("StatesToDisableMatches");
```
bgfoshay, line 68: Thought: Should we consider moving this out of the Find call? Or caching it (after reading it the first time) so that we don't need to parse it with every call to Find?

match/src/Piipan.Match/Piipan.Match.Func.Api/Api.cs
```diff
@@ -63,6 +64,15 @@ public class MatchApi
...
+                string[] disabledStates = disabledMatchingStates?.Split(',') ?? new string[0];
+
+                // If we wanted to disable all states, perhaps we could check for if this array contains "all"?
+                if (disabledStates.Contains(initiatingState.ToLower()))
```
bgfoshay, line 72: Question: Should we consider an additional condition that the matching state is also not in disabledStates? e.g. the initiating state is enabled, but a match is found with a state that is disabled?

dborgen2: Yea, that's definitely a valid question. I'll ask Andrew, and if he says we should do that I'll make sure to implement it when creating the final PR

match/src/Piipan.Match/Piipan.Match.Func.Api/Api.cs
dborgen2: From what I'm seeing, a new MatchApi instance is created every time the function is called, so moving it to the constructor wouldn't really be of any benefit, and in fact may potentially hurt response time if we ever have another function in the file that doesn't depend on this environment variable.

That being said, it should speed it up a bit to use a MemoryCache singleton, so I believe the final implementation should utilize that.

match/src/Piipan.Match/Piipan.Match.Func.Api/Api.cs
bgfoshay: 👍

cjhskippy (changes requested):

iac/arm-templates/function-orch-match.json
```diff
@@ -34,6 +34,9 @@
         },
         "states": {
             "type": "String"
+        },
+        "statesToDisableMatches": {
```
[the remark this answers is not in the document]
dborgen2: Yep, not a problem. I was wondering about that.. I'll reverse it

iac/states.csv
```diff
@@ -1,7 +1,7 @@
...
-EB,Echo Bravo
-EC,Echo Charlie
-IA,Iowa
-LA,Louisiana
-MA,Massachusetts
-MT,Montana
+EA,Echo Alpha,FALSE
+EB,Echo Bravo,FALSE
+EC,Echo Charlie,FALSE
+IA,Iowa,FALSE
```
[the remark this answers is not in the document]
dborgen2: Currently no, but I think this would be a great idea. I can add that in the final code

match/src/Piipan.Match/Piipan.Match.Func.Api/Api.cs
dborgen2: Added MemoryCache

iac/states.csv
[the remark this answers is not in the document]
dborgen2: FYI, after talking with Ryan there's nothing to do for this PR. He said they will maintain separate states.csv files depending on where they are publishing. So the result is the same, just nothing needing to be updated for this PR.

match/src/Piipan.Match/Piipan.Match.Func.Api/Api.cs
```diff
@@ -63,6 +68,27 @@ public class MatchApi
                 var response = await _matchApi.FindMatches(request, initiatingState);
                 response = await _matchEventService.ResolveMatches(request, response, initiatingState);
 
+                // If our initiating state is in the enabled states list return the result, otherwise return nothing.
+                var enabledStatesList = _memoryCache.GetOrCreate("EnabledStates", (entry) =>
+                {
+                    string enabledStates = Environment.GetEnvironmentVariable("EnabledStates");
```
bgfoshay, line 74: Question: Should you convert to lowercase for consistent comparison purposes?

match/tests/Piipan.Match.Func.Api.IntegrationTests/ApiIntegrationTests.cs
```diff
@@ -90,6 +91,7 @@ static Mock<HttpRequest> MockRequest(string jsonBody)
         static MatchApi Construct()
         {
             Environment.SetEnvironmentVariable("States", "ea");
+            Environment.SetEnvironmentVariable("EnabledStates", "ea,eb");
```
bgfoshay, line 94: Thought: Might be good to test with various capitalization. But I haven't finished review so you might have this elsewhere.

match/tests/Piipan.Match.Func.Api.IntegrationTests/ApiIntegrationTests.cs
```diff
@@ -185,6 +189,29 @@ public async void ApiReturnsEmptyMatchesArray()
...
+            ClearParticipants();
+            Insert(record);
+
+            // Act
+            var response = await api.Find(mockRequest.Object, logger);
+            var result = response as JsonResult;
+            var resultObject = result.Value as OrchMatchResponse;
+
+            // Assert
+            Assert.Empty(resultObject.Data.Results[0].Matches);
```
bgfoshay, line 212: Question: Do we need to verify that a match record was still created, even though an empty response was provided.

dborgen2: Good idea. Just added an assert for that

match/src/Piipan.Match/Piipan.Match.Func.Api/Api.cs
dborgen2: Good idea. It was lowercasing in create-resources.bash, but in case they edited it directly in Azure we should lowercase it.

bgfoshay (approved):
👍

PR #3211, opened 2022-05-10
CJC-639

ccaprio97:
## What’s changing?

Making the search reason validation add to the list of errors on the participant level instead of failing the entire match request

## Why?

NAC ticket 639

## This PR has:

### DIFF
--- a/match/src/Piipan.Match/Piipan.Match.Core/Services/MatchService.cs
+++ b/match/src/Piipan.Match/Piipan.Match.Core/Services/MatchService.cs
@@ -5,6 +5,8 @@ using Piipan.Match.Api;
 using Piipan.Match.Api.Models;
 using Piipan.Match.Core.Extensions;
 using Piipan.Participants.Api;
+using Piipan.Match.Core.Enums;
+using System;
 
 namespace Piipan.Match.Core.Services
 {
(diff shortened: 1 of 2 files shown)


match/src/Piipan.Match/Piipan.Match.Core/Services/MatchService.cs
```diff
@@ -39,6 +41,20 @@ public async Task<OrchMatchResponse> FindMatches(OrchMatchRequest request, strin
...
+                {
+                    response.Data.Errors.Add(
+                        new OrchMatchError
+                        {
+                            Index = i,
+                            Code = "Invalid Search Reason",
+                            Detail = "Search Reason: " + person.SearchReason.ToString()
+                        });
+                    continue;
+                }
```
kingcomma, line 57: *Question:* Does this approach offer advantages over adding a new validation rule to the existing [`RequestPersonValidator`](https://github.com/18F/piipan/blob/dev/match/src/Piipan.Match/Piipan.Match.Core/Validators/RequestPersonValidator.cs)?

ccaprio97: originally tried, misunderstood how the validator worked, now validating through the IValidator

kingcomma: Nice! I was hoping that would be possible. Seems like it's a little less code to wrangle and keeps the validation handling contained.

PR #3229, opened 2022-05-11
Updating nightly pa11y testing

dborgen2:
## What’s changing?

As part of the nightly build, circleci currently runs pa11y tests before running cypress tests. These tests are timing out, and don't need to occur anyway since we started calling pa11y tests inside of our cypress tests. Therefore, the following changed:

- Removed pa11y tests from being explicitly called from circleci. These will now happen as part of cypress testing
- Updated the dashboard app to call pa11y from cypress. This already existed in the query tool app.
- Updated one of the query tool cypress tests to create a match and grab the match ID, instead of relying on a match ID that the integration tests were using. After further review, this will likely fail since the integration tests dispose of that record when they're done.

## Why?

Closes NAC-734

## This PR has:


bgfoshay (approved):
LGTM

dashboard/src/Piipan.Dashboard/cypress.json
```diff
@@ -1,6 +1,7 @@
 {
-  "viewportHeight": 1000,
-  "viewportWidth": 1280,
-  "screenshotOnRunFailure": false,
-  "video": false
+    "viewportHeight": 1000,
+    "viewportWidth": 1280,
+    "screenshotOnRunFailure": false,
```
ryangeorgeharvey, line 4: Suggestion (non-blocking): Set `screenshotOnRunFailure` and `video` to `true`. I have this enabled in GH actions, and it can be helpful for developers if things fail or it's needed for audit purposes. We will get more value of it when fully e2e.

ryangeorgeharvey (approved):
LGTM, only suggestion is enabling video/screenshots output for the cypress.

ryangeorgeharvey:
I know you didn't make the change, but you could also revert the 30 second sleep timer in circleCI.

PR #3267, opened 2022-05-19
Participant upload error logging

dborgen2:
## What’s changing?

When the participant upload gets changed to "Failed", we should be logging information about the file so that we can try to identify what caused the upload to fail.

Per NAC-577, that data includes:

- State that uploaded
- Timestamp of upload processing start
- Timestamp of upload processing fail
- Error and location of the error (eg. stacktrace)
- Name of failed file

We also need to redact out any PII (LDS Hash, Participant ID, Case Number) from the error text or file name. A redaction service was created for this purpose.

A small IAC change was created to add the State to the Bulk Upload Azure Function. We could have gotten this information from the database connection string, but it seemed quite hacky, so I figured adding it as a configuration setting would be the best bet.

## Why?

Closes NAC-577

## This PR has:

### DIFF
--- a/etl/src/Piipan.Etl/Piipan.Etl.Func.BulkUpload/BulkUpload.cs
+++ b/etl/src/Piipan.Etl/Piipan.Etl.Func.BulkUpload/BulkUpload.cs
@@ -1,20 +1,12 @@
 // Default URL for triggering event grid function in the local environment.
 // http://localhost:7071/runtime/webhooks/EventGrid?functionName={functionname}
-using System;
-using System.Collections.Generic;
-using System.IO;
-using System.Threading;
-using System.Threading.Tasks;
-using Azure.Messaging.EventGrid;
-using Azure.Storage.Blobs;
-using Azure.Storage.Blobs.Models;
 using Microsoft.Azure.WebJobs;
-using Microsoft.Azure.WebJobs.Extensions.EventGrid;
 using Microsoft.Extensions.Logging;
-using Newtonsoft.Json;
-using Newtonsoft.Json.Linq;
 using Piipan.Etl.Func.BulkUpload.Parsers;
 using Piipan.Participants.Api;
+using System;
+using System.IO;
+using System.Threading.Tasks;
 
 namespace Piipan.Etl.Func.BulkUpload
 {
@@ -74,10 +66,10 @@ namespace Piipan.Etl.Func.BulkUpload
                     if (input != null)
                     {
                         var participants = _participantParser.Parse(input);
-                        await _participantApi.AddParticipants(participants, blobProperties.ETag.ToString())
+                        await _participantApi.AddParticipants(participants, blobProperties.ETag.ToString(), blockBlobClient.Name)
                                 .ContinueWith(t => _blobStream.DeleteBlobAfterProcessing(t, blockBlobClient, log))
                                 .ContinueWith(t => _participantApi.DeleteOldParticpants());
-                       
+
                     }
                 }
             }
--- a/etl/src/Piipan.Etl/Piipan.Etl.Func.BulkUpload/BulkUpload.cs
+++ b/etl/src/Piipan.Etl/Piipan.Etl.Func.BulkUpload/BulkUpload.cs
@@ -1,12 +1,12 @@
 // Default URL for triggering event grid function in the local environment.
 // http://localhost:7071/runtime/webhooks/EventGrid?functionName={functionname}
+using System;
+using System.IO;
+using System.Threading.Tasks;
 using Microsoft.Azure.WebJobs;
 using Microsoft.Extensions.Logging;
 using Piipan.Etl.Func.BulkUpload.Parsers;
 using Piipan.Participants.Api;
-using System;
-using System.IO;
-using System.Threading.Tasks;
 
 namespace Piipan.Etl.Func.BulkUpload
 {
@@ -57,7 +57,7 @@ namespace Piipan.Etl.Func.BulkUpload
                 {
                     var blockBlobClient = _blobStream.Parse(myQueueItem, log);
 
-                    Stream input = await blockBlobClient.OpenReadAsync();
+                    using Stream input = await blockBlobClient.OpenReadAsync();
 
                     log.LogInformation($"Input lenght: {input.Length} Position: {input.Position}");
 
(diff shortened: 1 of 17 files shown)


participants/src/Piipan.Participants/Piipan.Participants.Core/Services/ParticipantService.cs
```diff
@@ -1,14 +1,15 @@
...
-using System.Linq;
-using System.Threading.Tasks;
-using System.Transactions;
 using Microsoft.Extensions.Logging;
 using Piipan.Participants.Api;
 using Piipan.Participants.Api.Models;
 using Piipan.Participants.Core.DataAccessObjects;
-using Piipan.Participants.Core.Models;
 using Piipan.Participants.Core.Enums;
+using Piipan.Participants.Core.Models;
```
bgfoshay, line 6: Nitpick: Looks like your visual studio import organization settings are different from the omnisharp (https://github.com/18F/piipan/blob/daf64b8732665936885d269193ad7393e734391f/docs/source-guidelines.md#source-code-conventions) convention. Is there a way you can adjust it so that we don't churn on imports moving around?

etl/src/Piipan.Etl/Piipan.Etl.Func.BulkUpload/BulkUpload.cs
```diff
@@ -1,20 +1,12 @@
...
-using Azure.Storage.Blobs;
-using Azure.Storage.Blobs.Models;
 using Microsoft.Azure.WebJobs;
-using Microsoft.Azure.WebJobs.Extensions.EventGrid;
 using Microsoft.Extensions.Logging;
-using Newtonsoft.Json;
-using Newtonsoft.Json.Linq;
 using Piipan.Etl.Func.BulkUpload.Parsers;
 using Piipan.Participants.Api;
+using System;
```
bgfoshay, line 7: Nitpick: Nice to remove some of the unused imports but it looks like your visual studio import organization settings are different from the omnisharp (https://github.com/18F/piipan/blob/daf64b8732665936885d269193ad7393e734391f/docs/source-guidelines.md#source-code-conventions) convention. Is there a way you can adjust it so that we don't churn on imports moving around?

etl/src/Piipan.Etl/Piipan.Etl.Func.BulkUpload/Startup.cs
```diff
@@ -7,6 +6,7 @@
 using Piipan.Participants.Core.DataAccessObjects;
 using Piipan.Participants.Core.Extensions;
 using Piipan.Shared.Database;
+using System;
```
bgfoshay, line 9: Nitpick: same comment about imports organization

etl/tests/Piipan.Etl.Func.BulkUpload.Tests/BulkUploadTests.cs
```diff
@@ -17,6 +9,11 @@
 using Piipan.Participants.Api;
 using Piipan.Participants.Api.Models;
 using Piipan.Shared.API.Utilities;
+using System;
```
bgfoshay, line 12: Nitpick: same comment about imports organization

participants/src/Piipan.Participants/Piipan.Participants.Core/Services/ParticipantService.cs
```diff
@@ -70,9 +77,26 @@ public async Task AddParticipants(IEnumerable<IParticipant> participants,string
                     scope.Complete();
                 }
             }
-            catch (Exception ex) {
-                _logger.LogError(ex, ex.Message);
+            catch (Exception ex)
+            {
                 await _uploadDao.UpdateUploadStatus(upload, UploadStatuses.FAILED.ToString());
+
+                // Log the redacted error string as an error.
```
bgfoshay, line 84: Thought: You may consider pulling this whole block out into a separate method. Maybe even a separate class whose responsibility is simply to take a list of participants and log exceptions.

participants/src/Piipan.Participants/Piipan.Participants.Core/Services/ParticipantService.cs
```diff
@@ -47,20 +51,23 @@ public async Task<IEnumerable<IParticipant>> GetParticipants(string state, strin
...
 
-        public async Task AddParticipants(IEnumerable<IParticipant> participants,string uploadIdentifier)
+        public async Task AddParticipants(IEnumerable<IParticipant> participants, string uploadIdentifier, string fileName)
         {
+            DateTime uploadTime = DateTime.UtcNow;
             // Large participant uploads can be long-running processes and require
             // an increased time out duration to avoid System.TimeoutException
             var upload = await _uploadDao.AddUpload(uploadIdentifier);
-            try {
+            var participantList = participants.ToList();
```
bgfoshay, line 60: Question: This is going to pull all participants into memory rather than using delayed execution. Is that the intent?

participants/tests/Piipan.Participants.Core.IntegrationTests/ParticipantServiceTests.cs
```diff
@@ -1,9 +1,10 @@
-using System;
-using System.Linq;
 using Microsoft.Extensions.Logging;
 using Moq;
 using Piipan.Participants.Core.DataAccessObjects;
 using Piipan.Participants.Core.Services;
+using Piipan.Shared.Deidentification;
```
bgfoshay, line 5: Nitpick: same comment about imports

participants/tests/Piipan.Participants.Core.Tests/Services/ParticipantServiceTests.cs
```diff
@@ -1,19 +1,26 @@
-using System;
-using System.Collections.Generic;
-using System.Linq;
+using Microsoft.Extensions.Logging;
```
bgfoshay, line 1: Nitpick: same comment about imports

participants/src/Piipan.Participants/Piipan.Participants.Core/Services/ParticipantService.cs
dborgen2: It was, because as it reads the file it advances the read marker. So if it fails, for example, on the 2nd record, when we try to iterate in the catch block it would miss the first two records.

However, I think reading them all in at once will have a big side effect when we're dealing with huge files. Do you think we only need to scan for PII on the current record rather than all of them?

etl/src/Piipan.Etl/Piipan.Etl.Func.BulkUpload/BulkUpload.cs
dborgen2: Yea, looks like my import settings did not have "Place 'System' directives first when sorting usings". Options -> Text Editor -> C# -> Using Directives.

Checking that box and saving these files over again to place them in the right order.

participants/src/Piipan.Participants/Piipan.Participants.Core/Services/ParticipantService.cs
```diff
@@ -70,26 +74,45 @@ public async Task AddParticipants(IEnumerable<IParticipant> participants,string
                     scope.Complete();
                 }
             }
-            catch (Exception ex) {
-                _logger.LogError(ex, ex.Message);
+            catch (Exception ex)
+            {
```
bgfoshay, line 78: Question: Did you mean to remove the logging statement?

dborgen2: Yes. Now the errorCallback gets called, which comes back to the BulkUpload function. BulkUpload then resets the spot in the CSV, re-reads the participants, and calls the participant service LogParticipantsUploadError which ends up logging the redacted exception. I didn't really like using a callback like that, but felt it was the cleanest way.

bgfoshay: 🤦‍♂️ Of course. I didn't trace through the callback but that makes sense.

participants/src/Piipan.Participants/Piipan.Participants.Core/Services/ParticipantService.cs
dborgen2: Updated the function to not call ToList() right away. Instead, if an error occurs, a callback is sent back to the BulkUpload. BulkUpload now resets the CSV pointer to the first row, reads the records at that time, and then logs the redacted error.

bgfoshay (approved):
LGTM

PR #3273, opened 2022-05-23
Expanding location roles

dborgen2:
## What’s changing?

We are going to use the location claims to restrict access to a couple areas of the web application:

1. The search screen. The user will only be able to search if their location is a state, not a region or national office. That state will then be passed down to the API as their initiating state.
2. The match list screen. Only the national office will be able to view the match list screen.

Another small but purposeful change that was made to the view's model constructors was to accept a IServiceProvider instead of injecting everything as separate parameters. This makes code maintainability easier, as each subclass only needs to call `base(serviceProvider)` instead of passing in each injected instance. When new types are injected we no longer have to update each file, but instead only the base class.

## Why?

Closes 679 and 787

## This PR has:

### DIFF
--- a/query-tool/src/Piipan.QueryTool.Client/Components/QueryForm.razor
+++ b/query-tool/src/Piipan.QueryTool.Client/Components/QueryForm.razor
@@ -26,6 +26,8 @@
         {
             serverErrorList = new(QueryFormData.ServerErrors.Select(n => (n.Property, n.Error)));
         }
+
+        // Only locations with length 2, which are states, can perform a search.
         canPerformSearch = QueryFormData.Location?.Length == 2;
     }
     private Task SubmitForm(bool valid)
--- a/query-tool/src/Piipan.QueryTool.Client/Components/QueryForm.razor
+++ b/query-tool/src/Piipan.QueryTool.Client/Components/QueryForm.razor
@@ -11,27 +11,22 @@
 
 
 @code {
-    [Parameter] public PiiRecord Query { get; set; } = new();
+    [Parameter] public QueryFormModel QueryFormData { get; set; } = new();
 
-    /// <summary>
-    /// The search response that is collected after the user hits "Search" and an API call is made
-    /// </summary>
-    [Parameter] public ParameterBase<OrchMatchResponseData> QueryResult { get; set; }
-    [Parameter] public string Token { get; set; }
-    [Parameter] public List<ServerError> ServerErrors { get; set; }
-
-    private bool NoResults => QueryResult?.Data != null &&
-        (QueryResult?.Data.Results.Count == 0 || QueryResult?.Data.Results[0].Matches.Count() == 0);
+    private bool NoResults => QueryFormData.QueryResult != null &&
+        (QueryFormData.QueryResult.Results.Count == 0 || QueryFormData.QueryResult.Results[0].Matches.Count() == 0);
     private bool showNoResultsAlert = true;
     private bool searching = false;
+    private bool canPerformSearch = false;
     List<(string Property, string Error)> serverErrorList;
     protected override void OnInitialized()
     {
         base.OnInitialized();
-        if (ServerErrors?.Count > 0)
+        if (QueryFormData.ServerErrors?.Count > 0)
         {
-            serverErrorList = new(ServerErrors.Select(n => (n.Property, n.Error)));
+            serverErrorList = new(QueryFormData.ServerErrors.Select(n => (n.Property, n.Error)));
         }
+        canPerformSearch = QueryFormData.Location?.Length == 2;
     }
     private Task SubmitForm(bool valid)
     {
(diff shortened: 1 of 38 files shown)


query-tool/src/Piipan.QueryTool/Pages/BasePageModel.cs
```diff
@@ -1,38 +1,55 @@
...
 
         public override void OnPageHandlerExecuting(PageHandlerExecutingContext context)
         {
-            if ((string.IsNullOrEmpty(State) || string.IsNullOrEmpty(Role)) &&
+            if ((string.IsNullOrEmpty(Location) || string.IsNullOrEmpty(Role)) &&
                 (!context.HandlerMethod?.MethodInfo.CustomAttributes.Any(n => n.AttributeType == typeof(IgnoreAuthorizationAttribute)) ?? false))
             {
                 context.HttpContext.Response.StatusCode = 403;
-                context.Result = RedirectToPage("/NotAuthorized");
+                context.Result = RedirectToPage(NotAuthorizedPageName);
```
bgfoshay, line 31: Question: Should you instead call RedirectToUnauthorized() here?

query-tool/src/Piipan.QueryTool.Client/Components/QueryForm.razor
```diff
@@ -11,27 +11,22 @@
...
-    /// <summary>
-    /// The search response that is collected after the user hits "Search" and an API call is made
-    /// </summary>
-    [Parameter] public ParameterBase<OrchMatchResponseData> QueryResult { get; set; }
-    [Parameter] public string Token { get; set; }
-    [Parameter] public List<ServerError> ServerErrors { get; set; }
-
-    private bool NoResults => QueryResult?.Data != null &&
-        (QueryResult?.Data.Results.Count == 0 || QueryResult?.Data.Results[0].Matches.Count() == 0);
+    private bool NoResults => QueryFormData.QueryResult != null &&
```
bgfoshay, line 16: Question: Should we consider null conditional operator around the usages in this file of QueryFormData? e.g. QueryFormData?.ServerErrors?.Count

query-tool/src/Piipan.QueryTool/Pages/Index.cshtml.cs
```diff
@@ -1,14 +1,12 @@
-﻿using Microsoft.AspNetCore.Mvc;
+﻿using System;
```
bgfoshay, line 1: Nitpick: import order

query-tool/src/Piipan.QueryTool.Client/Components/QueryForm.razor
```diff
@@ -11,27 +11,22 @@
...
     protected override void OnInitialized()
     {
         base.OnInitialized();
-        if (ServerErrors?.Count > 0)
+        if (QueryFormData.ServerErrors?.Count > 0)
         {
-            serverErrorList = new(ServerErrors.Select(n => (n.Property, n.Error)));
+            serverErrorList = new(QueryFormData.ServerErrors.Select(n => (n.Property, n.Error)));
         }
+        canPerformSearch = QueryFormData.Location?.Length == 2;
```
bgfoshay, line 29: Question: What does this check mean? length == 2?

query-tool/src/Piipan.QueryTool/Pages/BasePageModel.cs
dborgen2: Yes, good catch

query-tool/src/Piipan.QueryTool.Client/Components/QueryForm.razor
dborgen2: While it wouldn't hurt anything, it might be a bit overkill. The property is instantiated immediately and is never currently set to null. That being said, it is currently possible that somewhere else in the program it could be set to null. I think I'd rather remove the setter on QueryFormData so that it's impossible for it to be null.

bgfoshay: That's a good point. Given that's the case, I like the idea of removing the setter

query-tool/src/Piipan.QueryTool.Client/Components/QueryForm.razor
dborgen2: This is probably overkill, because if Location is null we would have been redirected to the Unauthorized page. If it somehow was null though, calling .Length on it would throw a null reference exception. So ?.Length only checks length if it's not null.

If it is null, the left side of the equation returns null. Null == 2 is false, so canPerformSearch would be false if Location was null.

query-tool/src/Piipan.QueryTool/Pages/Index.cshtml.cs
```diff
@@ -19,49 +17,54 @@ public class IndexModel : BasePageModel
...
             {
                 try
                 {
                     _logger.LogInformation("Query form submitted");
 
                     string digest = _ldsDeidentifier.Run(
-                        Query.LastName,
-                        Query.DateOfBirth.Value.ToString("yyyy-MM-dd"),
-                        Query.SocialSecurityNum
+                        QueryFormData.Query.LastName,
```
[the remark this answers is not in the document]
dborgen2: Same thing here, let's just remove the setter on QueryFormData

query-tool/src/Piipan.QueryTool.Client/Components/QueryForm.razor
dborgen2: Sorry, I think I misread this. You're probably asking why we're doing it. Andrew said that states and only states will be two characters long, and that regions and the national office will be more than 2 characters long. I'll make a comment

bgfoshay: Yes, I was wondering more about why. I think a comment will help make it more clear.

query-tool/src/Piipan.QueryTool/Pages/List.cshtml.cs
```diff
@@ -30,6 +31,10 @@ public class ListModel : BasePageModel
 
         public async Task<IActionResult> OnGet()
         {
+            if (!States?.Contains("*") ?? true)
```
bgfoshay, line 34: Question/Thought: I had to read this a couple times to follow the logic. 

Would a simple function if(StatesListDoesNotContainWildcard) help make it more readable? Or StatesListDoesNotContainWildcardOrIsNull

query-tool/src/Piipan.QueryTool/Pages/NotAuthorized.cshtml.cs
```diff
@@ -1,21 +1,25 @@
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.AspNetCore.Mvc.RazorPages;
 using Piipan.Shared.Authorization;
-using Piipan.Shared.Claims;
+using System;
```
bgfoshay, line 4: Nitpick: imports order

shared/tests/Piipan.Shared.Tests/Locations/LocationProviderTests.cs
```diff
@@ -0,0 +1,76 @@
...
+        {
+            // Arrange
+            var options = Options.Create(locationOptions);
+            var locationProvider = new LocationsProvider(options);
+
+            // Act
+            var states = locationProvider.GetStates(null);
+
+            // Assert
+            Assert.Null(states);
```
bgfoshay, line 73: Question: Thinking about the GetStates return, is there any particular advantage to returning null? vs returning empty array or throwing an exception? Just curious

dborgen2: Yea, that's an interesting coding philosophy question. I personally don't think an exception be thrown unless something occurs that is not an expected use case as it is costly and needs to be handled by the caller. Having no location to me is an expected use case.

I may be in the minority but I generally like returning null so that the extra heap allocation for instantiating a list is not done. But obviously the time/memory this takes is very small and if we think an empty list should be returned I am perfectly fine with that.

bgfoshay (approved):
LGTM

PR #3286, opened 2022-05-26
NAC-574 Timestamp Successful State Participant Uploads in Metrics DB

edwintorres:
## What’s changing?

New participant successful uploads to the database are logged in the Metrics database using the EvenGrid, a new custom topic and a new function called BulkUploadMetricsDB. The BulkUploadMetrics was renamed as BulkUploadMetricsStorage. Changes to the VPN are needed for these changes to work.

## Why?

NAC-574. Capture the participant upload timestamps, so the we can reference the health and status of uploads over time. 

## This PR has:


etl/src/Piipan.Etl/Piipan.Etl.Func.BulkUpload/packages.lock.json
```diff
@@ -1,2074 +1,2095 @@
...
+      "WindowsAzure.Storage": {
+        "type": "Transitive",
+        "resolved": "9.3.1",
+        "contentHash": "NooNF4glP6BO7U4dno/xSfiEVVIv6OFcFfisX24Us2CZa9NQR3TSVEj9eVUlM5rLat5H9CHxk6M/mNSIaq7Vrw==",
+        "dependencies": {
+          "NETStandard.Library": "1.6.1",
+          "Newtonsoft.Json": "10.0.2"
+        }
+      },
+      "piipan.metrics.api": {
```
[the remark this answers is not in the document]
edwintorres: This part of the strategy is to re-use the ParticipantUpload created for the Metrics database which is the one the process is inserting the records. If separate, it will be needed to maintain the same object in two places.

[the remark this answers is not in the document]
edwintorres: I think that's a good approach. I was talking to @kingcomma on how to design the system in a way that the metrics layer is independent and you can replace it at any time. But the implications of that I think were outside of the scope of this task.

iac/create-metrics-resources.bash
```diff
@@ -21,17 +21,43 @@ set_constants () {
   # Metrics Collection Info
   COLLECT_APP_FILEPATH=Piipan.Metrics.Func.Collect
   COLLECT_STORAGE_NAME=${PREFIX}st${METRICS_COLLECT_APP_ID}${ENV}
-  COLLECT_FUNC=BulkUploadMetrics
+  COLLECT_FUNC=BulkUploadMetricsStorage
```
bgfoshay, line 24: Thought/Nitpick: Is there a more intention-revealing name we could use for these functions? One of them creates a metrics upload record upon initial processing. The other is used to update an existing record, whether its on success, failure, or possibly even could be used for periodic processing updates. What about CreateBulkUploadMetricsRecord and UpdateBulkUploadMetricsRecord? Or something similar? I'm open to other thoughts. Just trying to think of something that clearly communicates the responsibility of the functions.

iac/create-metrics-resources.bash
```diff
@@ -124,36 +150,55 @@ main () {
       }
     ]'
 
+
+  # Subscribe each dynamically created event blob topic to this function
+  METRICS_PROVIDERS=/subscriptions/${SUBSCRIPTION_ID}/resourceGroups/${RESOURCE_GROUP}/providers
+  SUBS_RESOURCE_GROUP=$RESOURCE_GROUP
+
+  # Create a Subscription to upload events that get routed to Function
+  custom_topic_name=evgt-custom-$ENV
```
bgfoshay, line 159: Thought: Similar to naming comment above, is there something more descriptive we could use for defining these topics instead of "custom"?

iac/iac-common.bash
```diff
@@ -60,6 +60,11 @@ UPLOAD_ENCRYPT_KEY_SHA_KV=upload-payload-key-sha
 # used in the bulk upload Azure Function.
 STATE_STR_KEY=State
 
+# Name of environment variable used to pass event grid connection
+# strings to app or function code
+EVENTGRID_CONN_STR_ENDPOINT=EventGridEndPoint
```
bgfoshay, line 65: Nitpick: I know I keep harping on naming, but should we consider changing these too to something more specific? Like UPDATE_METRICS_EVENTGRID_CONN_STR_ENDPOINT? I'm concerned it's too generic as is and I can see us adding additional event grid endpoints in the future.

metrics/src/Piipan.Metrics/Piipan.Metrics.Api/IParticipantUploadWriterApi.cs
```diff
@@ -5,6 +5,6 @@ namespace Piipan.Metrics.Api
 {
     public interface IParticipantUploadWriterApi
     {
-        Task<int> AddUpload(string state, DateTime uploadedAt);
+        Task<int> AddUpload(string state, DateTime uploadedAt, DateTime createdAt, string status, string uploadID, string etlStep);
```
bgfoshay, line 8: Question: could we pass the ParticipantUpload object here instead of all of the parameters individually? That way if ParticipantUpload changes in the future, we're not necessarily having to change the signature of this interface.

metrics/ddl/metrics.sql
```diff
@@ -3,7 +3,11 @@ BEGIN;
 CREATE TABLE IF NOT EXISTS participant_uploads(
     id serial PRIMARY KEY,
     state VARCHAR(50) NOT NULL,
-    uploaded_at timestamptz NOT NULL
+    uploaded_at timestamptz NOT NULL,
+    created_at  timestamptz NOT NULL,
+    status text NOT NULL,
+    upload_identifier text,
+    etl_step text
```
bgfoshay, line 10: Question: Just curious, what is etl_step for?

bgfoshay: Issue: Based on the AC in the JIRA ticket, we need to record the timestamp at processing completion. We might also consider an "updated_at" value to reflect when the record was updated.

metrics/src/Piipan.Metrics/Piipan.Metrics.Core/DataAccessObjects/ParticipantUploadDao.cs
```diff
@@ -80,28 +80,41 @@ FROM participant_uploads
                 });
             }
         }
-
-        public async Task<int> AddUpload(string state, DateTime uploadedAt)
+        
+        public async Task<int> AddUpload(string state, DateTime uploadedAt, DateTime createdAt, string status, string uploadID, string etlStep)
```
bgfoshay, line 84: Issue: It looks like you're only creating new records. Should we be updating pre-existing records instead? i.e. recording the new values like status and completion time?

metrics/src/Piipan.Metrics/Piipan.Metrics.Func.Collect/BulkUploadMetricsDB.cs
```diff
@@ -0,0 +1,78 @@
...
+        }
+
+        /// <summary>
+        /// Listens for BulkUpload events when users upload participants;
+        /// write meta info to Metrics database
+        /// </summary>
+        /// <param name="eventGridEvent">storage container blob creation event</param>
+        /// <param name="log">handle to the function log</param>
+
+        [FunctionName("BulkUploadMetricsDB")]
```
bgfoshay, line 31: Nitpick: See comments above regarding function naming.

metrics/src/Piipan.Metrics/Piipan.Metrics.Func.Collect/BulkUploadMetricsDB.cs
```diff
@@ -0,0 +1,78 @@
...
+            ILogger log)
+        {
+            log.LogInformation(eventGridEvent.Data.ToString());
+            try
+            {
+                DateTime uploadedAt = eventGridEvent.EventTime.DateTime;
+
+                //Prepare JSON string to be deserialize
+                var input = eventGridEvent.Data.ToString();
+                if (input.Substring(0, 1) == "\"") {
```
bgfoshay, line 43: Question: Can we put this logic in a method and add some documentation around what it's doing?

iac/create-metrics-resources.bash
edwintorres: BulkUploadMetricsStorage = create a record when a file is uploaded to the storage.
BulkUploadMetricsDB = create a record when a file is added to the database.

metrics/ddl/metrics.sql
edwintorres: Maybe here there is some confusion. Right now there is not an update. The process insert a record when the file in inserted in the storage, and insert another record when the participants are inserted in the database. To identify what part of the process ran, you need to use the etl_step. I sent a screenshot vis Teams which could help to understand this part.

metrics/src/Piipan.Metrics/Piipan.Metrics.Func.Collect/BulkUploadMetricsDB.cs
edwintorres: I thought about that, this was confusing to me because I was having different behavior so I really wanted to have a conversation with this one. When I run the test in the Azure environment the data in the EventGrid arrived with weird characters, an additional quote at the beginning and in the end. When running it on the rest I didn't have this behavior. I assumed is related to the encoding and the strings.

metrics/src/Piipan.Metrics/Piipan.Metrics.Func.Collect/BulkUploadMetricsStorage.cs
```diff
@@ -39,7 +39,7 @@ public BulkUploadMetrics(IParticipantUploadWriterApi participantUploadWriterApi)
                 string state = ParseState(eventGridEvent);
                 DateTime uploadedAt = eventGridEvent.EventTime.DateTime;
 
-                int nRows = await _participantUploadWriterApi.AddUpload(state, uploadedAt);
+                int nRows = await _participantUploadWriterApi.AddUpload(state, uploadedAt, uploadedAt, "Test", "Rest", "BulkUpload Storage" );
```
bgfoshay, line 42: Question: Are these parameters correct?

metrics/src/Piipan.Metrics/Piipan.Metrics.Func.Collect/BulkUploadMetricsDB.cs
```diff
@@ -0,0 +1,78 @@
...
+                input = input.Replace("\\u0022", "\"");
+                input = input.Replace("\"\"", "\"");
+
+                log.LogInformation(input);
+
+                var participantUpload = JsonConvert.DeserializeObject<ParticipantUpload>(input);
+
+                checkParticipantUpload(participantUpload);
+
+                int nRows = await _participantUploadWriterApi.AddUpload(
```
bgfoshay, line 56: Question: Will nRows ever be anything other than 1?

iac/create-metrics-resources.bash
bgfoshay: I updated the naming to reflect creating metrics records vs updating the status

metrics/src/Piipan.Metrics/Piipan.Metrics.Api/IParticipantUploadWriterApi.cs
bgfoshay: Updated to utilize ParticipantUpload object.

metrics/ddl/metrics.sql
bgfoshay: Slightly refactored the schema. Added column to capture completion time, status, upload_identifier. Also added columns to capture the number of participants uploaded and possible error messages.

metrics/src/Piipan.Metrics/Piipan.Metrics.Func.Collect/BulkUploadMetricsStorage.cs
bgfoshay: Removed hard-coded values and passed in real values of the upload.

metrics/src/Piipan.Metrics/Piipan.Metrics.Func.Collect/BulkUploadMetricsDB.cs
bgfoshay: I was able to remove this logic. I added code to Bulk Upload function to serialize the ParticipantUpload and convert it to BinaryData before putting in the EventGrid event. Then I deserialize to a ParticipantUpload object on the Metrics function side.

metrics/src/Piipan.Metrics/Piipan.Metrics.Core/DataAccessObjects/ParticipantUploadDao.cs
```diff
@@ -38,20 +47,32 @@ public async Task<Int64> GetUploadCount(string? state)
...
+                    error_message ErrorMessage
                 FROM participant_uploads";
 
             if (!String.IsNullOrEmpty(state))
             {
                 sql += $" WHERE lower(state) LIKE lower(@state)";
             }
 
-            sql += " ORDER BY uploaded_at DESC";
+            sql += " ORDER BY completed_at DESC";
```
dborgen2, line 75: Question: Just double checking we are supposed to be changing this logic to sort on completed_at. Will completed_at being potentially null cause any issues?

metrics/src/Piipan.Metrics/Piipan.Metrics.Func.Collect/UpdateBulkUploadMetrics.cs
```diff
@@ -0,0 +1,62 @@
...
+                log.LogInformation(String.Format("Number of rows inserted={0}", nRows));
+            }
+            catch (ArgumentException ex)
+            {
+                log.LogError(ex, "Failed updating Bulk Upload status & metrics.");
+                throw;
+            }
+        }
+
+        void checkParticipantUpload(ParticipantUpload upload){
```
dborgen2, line 57: Nitpick (unblocking): We usually use Pascal casing for naming functions

metrics/tests/Piipan.Metrics.Core.IntegrationTests/ParticipantUploadDaoTests.cs
```diff
@@ -186,13 +190,45 @@ public async Task AddUpload_InsertsRecord()
...
+
+            // Assert
+            uploadCount = await dao.GetUploadCount("ea");
+            Assert.Equal(1, uploadCount);
+            
+            Assert.Equal(uploadedAt, results.ToList()[0].UploadedAt);
+        }
+
+        [Fact]
+        public async Task UpdateUpload_InsertsRecord()
```
dborgen2, line 208: Nitpick Question (non-blocking): The name of this function confused me before I read through the code. I thought it was going to insert a record after calling update, but it doesn't appear to be doing that. Should it be called UpdateUpload_UpdatesRecord?

participants/src/Piipan.Participants/Piipan.Participants.Core/Services/ParticipantService.cs
```diff
@@ -64,6 +73,12 @@ public async Task AddParticipants(IEnumerable<IParticipant> participants, string
             // Large participant uploads can be long-running processes and require
             // an increased time out duration to avoid System.TimeoutException
             var upload = await _uploadDao.AddUpload(uploadIdentifier);
+
+            var participanUploadMetrics = new ParticipantUpload();
+            participanUploadMetrics.State = Environment.GetEnvironmentVariable("State");
```
dborgen2, line 78: Question: I think this Environment Variable should be passed into the function rather than pulling it in here. While right now the ParticipantService is used exclusively by the BulkUpload service, it shouldn't be dependent on an EnvironmentVariable that is set in the BulkUpload function, in my opinion.

participants/src/Piipan.Participants/Piipan.Participants.Core/Services/ParticipantService.cs
```diff
@@ -78,19 +93,44 @@ public async Task AddParticipants(IEnumerable<IParticipant> participants, string
...
+                    await _participantPublishUploadMetric.PublishUploadMetric(participanUploadMetrics);    
+
                     scope.Complete();
                 }
             }
             catch (Exception ex)
             {
                 await _uploadDao.UpdateUploadStatus(upload, UploadStatuses.FAILED.ToString());
+
+                participanUploadMetrics.CompletedAt = DateTime.UtcNow;
```
dborgen2, line 118: Question: Do we need to set ParticipantUploadMetrics status to failed here?

participants/tests/Piipan.Participants.Core.Tests/Services/ParticipantPublishUploadMetricTests.cs
```diff
@@ -0,0 +1,41 @@
...
+                Status = UploadStatuses.COMPLETE.ToString(),
+                UploadIdentifier = "UploadIdentifier",
+                UploadedAt = DateTime.UtcNow,
+                CompletedAt = DateTime.UtcNow,
+                ParticipantsUploaded = 50
+
+            };
+
+            // Act
+            await participantPublishUploadMetric.PublishUploadMetric(metric);
```
dborgen2, line 36: Question: Do we need to Assert anything here, or do we assume the function works if no errors are thrown?

metrics/src/Piipan.Metrics/Piipan.Metrics.Core/DataAccessObjects/ParticipantUploadDao.cs
bgfoshay: Good catch. We should keep this as uploaded_at for this method.

metrics/src/Piipan.Metrics/Piipan.Metrics.Func.Collect/CreateBulkUploadMetrics.cs
```diff
@@ -69,5 +82,21 @@ private string ParseState(EventGridEvent eventGridEvent)
                 throw new FormatException("State not found", ex);
             }
         }
+
+        private string ParseUploadId(EventGridEvent eventGridEvent)
+        {
+            try
+            {
+                var jsondata = eventGridEvent.Data.ToString(); ;
```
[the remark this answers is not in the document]
bgfoshay: 👍

metrics/tests/Piipan.Metrics.Core.IntegrationTests/ParticipantUploadDaoTests.cs
bgfoshay: Yes, definitely needed to be renamed. Thanks for catching.

participants/src/Piipan.Participants/Piipan.Participants.Core/Services/ParticipantService.cs
```diff
@@ -78,19 +93,44 @@ public async Task AddParticipants(IEnumerable<IParticipant> participants, string
...
+
+                string uploadDBStatus = JsonConvert.SerializeObject(participanUploadMetrics);
+
+                await _participantPublishUploadMetric.PublishUploadMetric(participanUploadMetrics);
+
                 errorCallback?.Invoke(ex);
             }
         }
 
+        private static long MapUlongToLong(ulong ulongValue)
```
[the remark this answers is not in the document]
bgfoshay: 👍

participants/src/Piipan.Participants/Piipan.Participants.Core/Services/ParticipantService.cs
bgfoshay: Good catch and I agree it should be passed. in. I'll update this.

participants/src/Piipan.Participants/Piipan.Participants.Core/Services/ParticipantService.cs
bgfoshay: Yes, definitely. Not sure how I missed that.

participants/tests/Piipan.Participants.Core.Tests/Services/ParticipantPublishUploadMetricTests.cs
bgfoshay: It would be good to add a check. I'll update this.

dborgen2 (approved):
👍

participants/src/Piipan.Participants/Piipan.Participants.Core/Services/ParticipantPublishUploadMetric.cs
```diff
@@ -0,0 +1,56 @@
...
+        public EventGridPublisherClient _client;
+
+        public ParticipantPublishUploadMetric(){
+
+                //Create event grid client to publish metric data
+                _client = new EventGridPublisherClient(
+                    new Uri(Environment.GetEnvironmentVariable("EventGridEndPoint")),
+                    new AzureKeyCredential(Environment.GetEnvironmentVariable("EventGridKeyString")),
+                    default
+                );
```
kingcomma, line 24: *Thought (non-blocking):* As the approach to publishing metrics from Function Apps is refined, would there be a benefit to using [Event Grid output bindings](https://docs.microsoft.com/en-us/azure/azure-functions/functions-bindings-event-grid-output?tabs=in-process%2Cextensionv3&pivots=programming-language-csharp) where possible? That might remove the need for another custom service that has to be explicitly configured, and make testing a bit easier by only needing to pass around an `IAsyncCollector<EventGridEvent>` instance.

participants/src/Piipan.Participants/Piipan.Participants.Core/Services/ParticipantPublishUploadMetric.cs
```diff
@@ -0,0 +1,56 @@
...
+            //Otherwise, EventGridEvent serializes it according to class property names rather than JsonProperty attributes
+            var binaryData = BinaryData.FromString(result);
+
+            // Add EventGridEvents to a list to publish to the topic
+            EventGridEvent egEvent =
+                new EventGridEvent(
+                    "Add participation",
+                    "Upload to the database",
+                    "1.0",
+                    binaryData);
```
kingcomma, line 48: *Thought (non-blocking):* Another future-minded thought — would there be benefit to establishing a standard structure for metrics events? Perhaps the initial set of metrics won't be very large and bespoke events wouldn't be too unwieldy. But there may still be benefits in terms making it easier to publish and consume each new event that gets added.

Maybe this would be good fodder for an ADR?

participants/src/Piipan.Participants/Piipan.Participants.Core/Services/ParticipantService.cs
```diff
@@ -78,15 +93,36 @@ public async Task AddParticipants(IEnumerable<IParticipant> participants, string
...
+                    participanUploadMetrics.CompletedAt = DateTime.UtcNow;
+                    participanUploadMetrics.ParticipantsUploaded = Convert.ToInt64(count);
+                    participanUploadMetrics.UploadIdentifier = uploadIdentifier;
+
+                    string uploadDBStatus = JsonConvert.SerializeObject(participanUploadMetrics);
+
+                     _logger.LogInformation(uploadDBStatus);
+
+                    await _participantPublishUploadMetric.PublishUploadMetric(participanUploadMetrics);    
+
```
kingcomma, line 110: *Question:* If the metric event publishing resulted in an exception, would the database transaction be rolled back? If so, under what circumstances might that happen?

bgfoshay: As it stands now, yes, it would get rolled back. I will double check whether that's desired behavior or not. This could happen if for whatever reason, the function cannot connect to Event Grid.

participants/src/Piipan.Participants/Piipan.Participants.Core/Services/ParticipantPublishUploadMetric.cs
bgfoshay: Yes, I think so. As we move forward and capture different areas of metrics, it would be useful to have a standard paradigm for this work.

participants/src/Piipan.Participants/Piipan.Participants.Core/Services/ParticipantPublishUploadMetric.cs
bgfoshay: This is a great idea and honestly did not even cross my mind when I picked up this work. That would definitely simplify things.

iac/create-metrics-resources.bash
```diff
@@ -124,44 +150,72 @@ main () {
...
   az functionapp config appsettings set \
     --resource-group "$RESOURCE_GROUP" \
     --name "$METRICS_COLLECT_APP_NAME" \
     --settings \
       $DB_CONN_STR_KEY="$db_conn_str" \
       $CLOUD_NAME_STR_KEY="$CLOUD_NAME" \
-    --output none
+      $EVENTGRID_CONN_STR_ENDPOINT="$eventgrid_endpoint" \
+      $EVENTGRID_CONN_STR_KEY="$eventgrid_key_str"
+    # --output none
```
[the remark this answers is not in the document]
bgfoshay: Good catch. That looks like a mistake. I'll fix this

PR #3296, opened 2022-05-30
Updated Participant Service tests to include a better exception test with the redaction serve

dborgen2:
## What’s changing?

Per comments made by @kingcomma, we should have a better test in the participant service tests that shows we can handle the actual redaction of an exception instead of just a mock.

Also after our discussion we noticed that if the error was due to the parsing and not the upload, that the parsing would continue to fail in the error logging. Thus we should catch any parsing errors and still redact the information that we were previously able to parse.

## Why?

Fixes concerns for NAC-577

## This PR has:

### DIFF
--- a/participants/src/Piipan.Participants/Piipan.Participants.Core/Services/ParticipantService.cs
+++ b/participants/src/Piipan.Participants/Piipan.Participants.Core/Services/ParticipantService.cs
@@ -88,11 +88,13 @@ namespace Piipan.Participants.Core.Services
 
             string[] redactedStrings = new string[3];
 
+            int redactionIndex = 0;
             try
             {
                 // for each participant, redact out their information from the current error string
                 foreach (var participant in participants)
                 {
+                    redactionIndex++;
                     redactedStrings[0] = participant.LdsHash;
                     redactedStrings[1] = participant.ParticipantId;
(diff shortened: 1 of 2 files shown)


participants/src/Piipan.Participants/Piipan.Participants.Core/Services/ParticipantService.cs
```diff
@@ -87,13 +87,22 @@ public void LogParticipantsUploadError(ParticipantUploadErrorDetails errorDetail
...
+                }
+            }
+            catch
             {
-                redactedStrings[0] = participant.LdsHash;
-                redactedStrings[1] = participant.ParticipantId;
-                redactedStrings[2] = participant.CaseId;
-                uploadErrorString = _redactionService.Redact(uploadErrorString, redactedStrings);
+                // If it still errors in here, nothing we can do to redact it or any of the following records.
+                // But we need to continue on and redact everything else.
```
bgfoshay, line 105: Question: Is it worth logging anything here to indicate that it failed during redaction?

dborgen2: Yea, that would probably help us identify the error. Good idea... updated

bgfoshay (approved):
LGTM

PR #3298, opened 2022-05-31
Column level Encrypt/Decrypt

kcherukupalli:
## What’s changing?

Implemented with Column level Encrypt/Decrypt using AES with CBC.

Implement AES key from Azure key vault is still pending.

## Why?

In order to provide more secured data.

#nac-66

## This PR has:

- [ x] **Commented source code** (_required on public classes and methods, and elsewhere as appropriate_)

- [ x] **Changes to IAC** (_add any deployment steps that require manual intervention to the draft release notes_)


participants/src/Piipan.Participants/Piipan.Participants.Core/Services/ParticipantService.cs
```diff
@@ -43,7 +46,11 @@ public async Task<IEnumerable<IParticipant>> GetParticipants(string state, strin
                 var participants = await _participantDao.GetParticipants(state, ldsHash, upload.Id);
 
                 // Set the participant State before returning
-                return participants.Select(p => new ParticipantDto(p) { State = state });
+                return participants.Select(p => new ParticipantDto(p) { 
+                    State = state,
```
dborgen2, line 50: Question? Not really sure why ParticipantDto has LDS hash as a property, but since it does do we need to Decrypt LDS Hash here?

participants/src/Piipan.Participants/Piipan.Participants.Core/Services/ParticipantService.cs
```diff
@@ -66,9 +73,12 @@ public async Task AddParticipants(IEnumerable<IParticipant> participants, string
                 {
                     var participantDbos = participants.Select(p => new ParticipantDbo(p)
                     {
-                        UploadId = upload.Id
+                        UploadId = upload.Id,
+                        LdsHash = _cryptographyClient.EncryptToBase64String(p.LdsHash),
```
dborgen2, line 77: Question: When we do the upload redaction (ParticipantService), we will have re-read the file and the properties we are trying to redact would not be encrypted at that point. Should we attempt to redact both the unencrypted and encrypted properties?

participants/src/Piipan.Participants/Piipan.Participants.Core/Services/ParticipantService.cs
bgfoshay: Good question. The answer to decrypting is 'no'. The thought is that we never need to decrypt the LDS Hash with the intention of keeping it more secure by not revealing the decrypted value anywhere. The answer to why the property is included on the Dto to begin with is murkier. I don't see value in having it there but it's a legacy property that's been there a while.

participants/src/Piipan.Participants/Piipan.Participants.Core/Services/ParticipantService.cs
bgfoshay: Good question. I'll need to talk with Andrew. I'm not as concerned with redacting the encrypted properties. But we need to cover the unencrypted ones

dborgen2: Makes sense. Ok, I'll approve this, and if he decides we need that we can always add it in later

dborgen2 (approved):
👍

PR #3311, opened 2022-06-06
Microsoft Graph Migration

ryangeorgeharvey:
## What’s changing?

Changes need to support the Microsoft Graph migration, which is included as part of Azure CLI 2.37

## Why?

Azure CLI 2.37 
Please Refer: https://docs.microsoft.com/en-us/cli/azure/microsoft-graph-migration

## This PR has:


iac/create-resources.bash
```diff
@@ -130,11 +136,11 @@ main () {
         --name "$SIEM_RECEIVER" \
         --role "Reader" \
         --scopes "/subscriptions/${SUBSCRIPTION_ID}" \
-        --query "[0].objectId" \
+        --query "[0].id" \
         --output tsv)
 
     # Wait bit to avoid "InvalidPrincipalId" on app registration use below
-    sleep 60
+    sleep 120
```
bgfoshay, line 143: Question: Just curious, is the new version of CLI slower?

iac/iac-common.bash
```diff
@@ -118,6 +115,20 @@ OIDC_APPS=("$QUERY_TOOL_APP_NAME" "$DASHBOARD_APP_NAME")
...
+  type=$(az account show --query user.type --output tsv)
+
+  if [ "${type}" = "servicePrincipal" ]; then
+    app_id=$(az account show --query user.name --output tsv)
+    CURRENT_USER_OBJID=$(az ad sp show --id "${app_id}" --query id --output tsv)
+  else
+    CURRENT_USER_OBJID=$(az ad signed-in-user show --query id --output tsv)
+  fi
+}
+current_user_objid
```
bgfoshay, line 130: Question: is this line (current_user_objid) supposed to be here?

iac/create-core-databases.bash
```diff
@@ -39,6 +39,7 @@ main () {
     --name "$PG_SECRET_NAME" \
     --file /dev/stdin \
     --query id
+    #--value "$PG_SECRET" \
```
bgfoshay, line 42: Nitpick: Should we remove this comment?

iac/iac-common.bash
ryangeorgeharvey: Yes, it's need to invoke the function. So I define the function, then call it right after.

iac/create-core-databases.bash
bgfoshay: Ah, ok. Got it. Thanks

iac/create-resources.bash
ryangeorgeharvey: The underlying creation of SP is done though Microsoft Graph now. I'm not sure if it's directly related, but I was seeing some failures when creating the SP for RBAC, getting invalid principle. So I simply changed the timeout, and that seemed to give it enough time. 

I didn't want to spend a long time digging into the timing, and figured just extending it would be a quick fix.

iac/iac-common.bash
bgfoshay: whoops, you're right. Didn't register with me that that was the name of the function. Got it now

iac/iac-common.bash
```diff
@@ -342,6 +353,7 @@ set_oidc_secret () {
     --name "$secret_name" \
     --file /dev/stdin \
     --query id > /dev/null
+    #--value "$value"
```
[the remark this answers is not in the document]
bgfoshay: nm. Saw your comment above.

iac/create-resources.bash
```diff
@@ -202,6 +208,7 @@ main () {
     --name "$PG_SECRET_NAME" \
     --file /dev/stdin \
     --query id
+    #--value "$PG_SECRET" \
```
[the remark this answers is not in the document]
bgfoshay: nm. Saw your comment above.

iac/create-resources.bash
```diff
@@ -79,6 +79,12 @@ main () {
   verify_cloud
   set_constants
 
+  # Only show errors from the CLI ouput, mostly to avoid 'WARNING: This command
+  # or command group has been migrated to Microsoft Graph API. Please carefully
+  # review all breaking changes introduced during this migration:
+  # https://docs.microsoft.com/cli/azure/microsoft-graph-migration'
+  az config set core.only_show_errors=true
```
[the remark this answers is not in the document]
ryangeorgeharvey: That's a good point. Yes it would disable those warnings. However, the current CLI will spit out hundreds of warnings with the new migration. Even if the command is updated and working correctly, the CLI will give the warning just to notify the user that there could be issues.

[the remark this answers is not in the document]
ryangeorgeharvey: I could suppress for each command, but that would be tedious and could still suppress valid warnings. I could also do conditional suppress, meaning we would show it for certain environments, and not others?

The warnings produce a lot of noise though, when trying to debug the actual deployments. So I thought it helpful to remove for that reason.

iac/configure-easy-auth.bash
```diff
@@ -101,7 +111,7 @@ configure_aad_app_reg () {
   app_uri=$(app_id_uri "$app")
   az ad app update \
     --id "$object_id" \
-    --available-to-other-tenants false \
+    --sign-in-audience "AzureADMyOrg" \
```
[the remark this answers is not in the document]
ryangeorgeharvey: https://docs.microsoft.com/en-us/azure/active-directory/develop/supported-accounts-validation

iac/create-resources.bash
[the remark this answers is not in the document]
bgfoshay: Hmmmm, can we run the cli as is for the first command (or for an arbitrary one) just so that we see the warning before turning it off for everything? I agree that we don't want to suppress for each command. That's way too tedious. Showing it during DEV environment might be an option.

bgfoshay (approved):
👍

PR #3315, opened 2022-06-06
Providing support for accessing State_Info table

ccaprio97:
## What’s changing?

Making Core and Api Projects to access the state_info table in the core db

## Why?

NAC-594

## This PR has:


shared/src/Piipan.Shared.TestFixtures/StateInfoDbFixture.cs
```diff
@@ -0,0 +1,115 @@
...
+namespace Piipan.Shared.TestFixtures
+{
+    public class StateInfoDbFixture : IDisposable
+    {
+        public readonly string ConnectionString;
+        public readonly NpgsqlFactory Factory;
+
+        public StateInfoDbFixture()
+        {
+            ConnectionString = "Server=cc-psql-core-cjc.postgres.database.azure.com;Database=collaboration;Port=5432;User Id=piipanadmin@cc-psql-core-cjc;Password=qEAqNcaRG6yCaXbmna6HHY2OUWZn6aKEhSkyfGeH8i03AYNBhhY1R9Wha81c7e48aA1!;";
```
dborgen2, line 15: This connection string I think is here by mistake

states/Tests/Piipan.States.Core.Integration.Tests/.gitignore
```diff
@@ -0,0 +1,433 @@
+## Ignore Visual Studio temporary files, build results, and
```
dborgen2, line 1: The other folders just have one .gitignore file at the main level. I'm wondering if we have too many of these .gitignore files?

system.sln
```diff
@@ -140,6 +140,18 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution
...
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Piipan.States.Api", "states\src\Piipan.States\Piipan.States.Api\Piipan.States.Api.csproj", "{08CBC755-9724-41B2-AB08-189BE5A67B40}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Piipan.States.Core", "states\src\Piipan.States\Piipan.States.Core\Piipan.States.Core.csproj", "{7ADE7269-D297-4CB9-8258-E5CD171BD28F}"
+EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Tests", "Tests", "{FEC0F4AF-9F64-42C4-A7CC-848F37F9EAA0}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Piipan.States.Core.Integration.Tests", "states\Tests\Piipan.States.Core.Integration.Tests\Piipan.States.Core.Integration.Tests.csproj", "{72B0FC2A-8ED9-44C3-9F17-EB00A76CCAC8}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "TestProject1", "TestProject1\TestProject1.csproj", "{B2544E83-A671-452D-A2E6-EC7DBE6E1995}"
```
dborgen2, line 153: Question: Is this project here by accident? I tried getting your branch and got an error with this project

dborgen2:
I believe we also need to make changes to:
- https://github.com/18F/piipan/blob/dev/build.bash
- https://github.com/18F/piipan/blob/dev/.circleci/config.yml
- https://github.com/18F/piipan/blob/dev/.github/dependabot.yml

To be consistent with the other folders, we probably should add a states.sln and build.bash inside the States folder.

states/src/Piipan.States/Piipan.States.Api/Models/IState.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Piipan.States.Api.Models
+{
+    public interface IState
```
dborgen2, line 9: The more I think about it, the more I think we should probably have this be a class. Since the JSON results will be returned to multiple projects it's probably imperative they are serialized/deserialized the same in all cases.

If you make this a class and put [JsonProperty("state_abbreviation")], etc. that should solidify our serialization. Then you wouldn't need StateInfoDbo either

dborgen2, line 9: The more I think about it, the more I think we should probably have this be a class. Since the JSON results will be returned to multiple projects it's probably imperative they are serialized/deserialized the same in all cases.

If you make this a class and put [JsonProperty("state_abbreviation")], etc. that should solidify our serialization. Then you wouldn't need StateInfoDbo either

ccaprio97:
@dborgen2  So this is basically everything. I still need to go through and remove some of the unused using lines and you will probably want me to change the get all states test because it feels messy. But its 6 so I have to go bike and take zeus out. Will finish the fixes tomorrow

states/Tests/Piipan.States.Core.Integration.Tests/.gitignore
ccaprio97: participants has a git ignore in the integration tests folder and another in the core tests folder

states/src/Piipan.States/Piipan.States.Api/Models/IState.cs
ccaprio97: um....I think this is more a @bgfoshay question then me since its on how we want to do it

states/Tests/Piipan.States.Core.Integration.Tests/StateInfoDaoTests.cs
```diff
@@ -0,0 +1,110 @@
...
+                .ReturnsAsync(() =>
+                {
+                    var conn = Factory.CreateConnection();
+                    conn.ConnectionString = ConnectionString;
+                    return conn;
+                });
+
+            return factory.Object;
+        }
+
```
bgfoshay, line 33: Thought: It would probably be useful to have tests for when GetStateByName is called with a name that doesn't exist. Something like GetStateByName_ThrowsExceptionIfNotFound. Same thing for GetStateById

states/src/Piipan.States/Piipan.States.Core/DataAccessObjects/StateInfoDao.cs
```diff
@@ -0,0 +1,59 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Piipan.States.Core.Models;
+using Dapper;
+
+namespace Piipan.States.Core.DataAccessObjects
+{
+    public class StateInfoDao : IStateInfoDao
```
bgfoshay, line 13: Nitpick: We need to make to add documentation for public classes & methods. I know not all classes currently have this, but we need to do a better job moving forward. See MatchResEventDao.cs as an example.

states/Tests/Piipan.States.Core.Integration.Tests/.gitignore
dborgen2: Hm... ok. Unfortunately there doesn't seem to be a consistent pattern then. Oh well, this should be fine then

states/src/Piipan.States/Piipan.States.Api/Models/IState.cs
bgfoshay: I agree that adding JsonProperty to the Dto will solidify serialization. I wouldn't necessarily collapse the dto and dbo into one class at this point though. I only say that because it's not uncommon for a dto to differ from a dbo in more ways than just naming. 

To Dan's point about the interface, I think it's better to use the Dto class itself in consuming projects (i.e. don't use the interface). The interface is really not serving much purpose beyond forcing the the dbo/dto objects to be aligned. Again, that might not always be the case though.

dborgen2: Yep, I'm good with two separate classes in case we need to add something to the dbo object in the future. We would just need to make sure to add the same JsonProperty attributes to both classes to keep them in sync.

ccaprio97: Did not do DTO this will be done in a later PR that involves service class

ccaprio97: Did not do DTO this will be done in a later PR that involves service class

states/tests/Piipan.States.Core.Integration.Tests/Properties/launchSettings.json
```diff
@@ -0,0 +1,10 @@
+{
+  "profiles": {
```
bgfoshay, line 2: Nitpick: Don't include this file in the commit. Should be ignored.

states/src/Piipan.States/Piipan.States.Core/Models/StateInfoDbo.cs
```diff
@@ -0,0 +1,44 @@
...
+    public class StateInfoDbo : IState
+    {
+        public string Id { get; set; }
+        public string State { get; set; }
+        public string StateAbbreviation { get; set; }
+        public string Email { get; set; }
+        public string? Phone { get; set; }
+        public string Region { get; set; }
+
+        public override bool Equals(Object obj)
```
dborgen2, line 22: Question (non-blocking): Should we in general consider [Records](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/tutorials/records) for these types of things? They have a built in Equals method that we won't have to maintain. More of a general question as I've seen this pattern done a few times.

dborgen2 (approved):
LGTM, just a non-blocking question on the use of records

states/src/Piipan.States/Piipan.States.Core/Models/StateInfoDbo.cs
bgfoshay: This is a great point. The less code we have to maintain, the better. This is a good candidate for Records that we should consider in a subsequent PR.

PR #3320, opened 2022-06-08
Authorization logic for location and role to be at application level

dborgen2:
## What’s changing?

The authorization for location and role was added a few weeks back at the page level. After discussion with @kingcomma, this should be done at the application level by looking at the required claims.

The possible values for location and role are very large, but they all start with the same prefixes. So the appsettings now allows a combination of a set value and a wildcard. The AuthorizationPolicyBuilder was updated to reflect this.

## Why?

Closes NAC-857

## This PR has:

### DIFF
--- a/shared/src/Piipan.Shared/Authorization/AuthorizationPolicyBuilder.cs
+++ b/shared/src/Piipan.Shared/Authorization/AuthorizationPolicyBuilder.cs
@@ -1,5 +1,5 @@
-using System.Collections.Generic;
 using System.Linq;
+using System.Text.RegularExpressions;
 using Microsoft.AspNetCore.Authorization;
 
 namespace Piipan.Shared.Authorization
@@ -22,8 +22,21 @@ namespace Piipan.Shared.Authorization
                     // if the allowed values includes a wildcard, allow any value for this claim type
                     if (rcv.Values.Any(v => v == "*"))
                     {
-                        innerBuilder.RequireAssertion(context => {
+                        innerBuilder.RequireAssertion(context =>
+                        {
                             return context.User.Claims.Any(c => c.Type == rcv.Type);
+                        });
+                    }
+                    // If any values contain a wildcard, use a regex expression to test for wildcard
+                    else if (rcv.Values.Any(v => v.Contains("*")))
+                    {
+                        innerBuilder.RequireAssertion(context =>
+                        {
+                            return context.User.Claims.Any(c => c.Type == rcv.Type && rcv.Values.Any(v =>
+                            {
+                                var valueAsRegex = "^" + Regex.Escape(v).Replace("\\*", ".+") + "$";
+                                return Regex.IsMatch(c.Value, valueAsRegex);
+                            }));
                         });
                     }
                     else
@@ -32,7 +45,7 @@ namespace Piipan.Shared.Authorization
                     }
                 }
             }
-            
+
             innerBuilder.RequireAuthenticatedUser();
 
             return innerBuilder.Build();
--- a/shared/tests/Piipan.Shared.Tests/Locations/LocationProviderTests.cs
+++ b/shared/tests/Piipan.Shared.Tests/Locations/LocationProviderTests.cs
@@ -7,11 +7,7 @@ namespace Piipan.Shared.Locations.Tests
     {
         LocationOptions locationOptions = new LocationOptions
         {
-            Map = new LocationMapping[]
-            {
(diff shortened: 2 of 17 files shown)


shared/tests/Piipan.Shared.Tests/Locations/LocationProviderTests.cs
```diff
@@ -31,19 +27,20 @@ public void GetNationalStates()
...
-            var states = locationProvider.GetStates("Midwest");
+        //    // Act
+        //    var states = locationProvider.GetStates("Midwest");
 
-            // Assert
-            Assert.Equal(new string[] { "WI", "IA", "MN" }, states);
-        }
+        //    // Assert
+        //    Assert.Equal(new string[] { "WI", "IA", "MN" }, states);
+        //}
```
kingcomma, line 43: *Thought (non-blocking):* When the time comes to build in support for regions, it might be worth considering if the mapping of a region code to a list of states is considered "mostly-constant" and how that might [affect where it is stored](https://thoughtbot.com/blog/extract-mostly-constant-data-from-the-database).

shared/src/Piipan.Shared/Authorization/AuthorizationPolicyBuilder.cs
```diff
@@ -22,17 +22,30 @@ public static AuthorizationPolicy Build(AuthorizationPolicyOptions options)
...
+                    {
+                        innerBuilder.RequireAssertion(context =>
+                        {
+                            return context.User.Claims.Any(c => c.Type == rcv.Type && rcv.Values.Any(v =>
+                            {
+                                var valueAsRegex = "^" + Regex.Escape(v).Replace("\\*", ".+") + "$";
+                                return Regex.IsMatch(c.Value, valueAsRegex);
+                            }));
+                        });
+                    }
```
kingcomma, line 41: *Question (non-blocking):* Did this work take you into the [EasyAuth middleware's IClaimsTransformer functionality](https://github.com/dasiths/NEasyAuthMiddleware#customizing-it) at all?

Thinking towards the future when our authz needs become more complex, I'm wondering if it would allow us to convert the custom claim types and values we're stuck with into something standardized and easier to work with. E.g., `{"custom_role_claim": "PREFIX_role-name"}` becomes something like `{"role": "role-name"}`. We'd have to be careful about unintended consequences, but it would open the door to relying on ASP.NET's built-in [role-based](https://docs.microsoft.com/en-us/aspnet/core/security/authorization/roles?view=aspnetcore-6.0) or [claims-based](https://docs.microsoft.com/en-us/aspnet/core/security/authorization/claims?view=aspnetcore-6.0) authorization and limit the need for us to do custom work.

kingcomma (commented):
Nice! Love seeing the reduced need for us to write custom authorization logic. I added a couple comments/questions inline.

One other non-blocking thought: if dealing with all the potential number of roles and location values becomes cumbersome, we may be able to get away with limiting the list of possible values to the initial roles and locations we know we need now and come up with a more flexible solution down the road.

I'll keep my reviewing limited to these non-blocking comments and hand-off approval to @bgfoshay or others.

shared/tests/Piipan.Shared.Tests/Locations/LocationProviderTests.cs
dborgen2: Yea, good question. After discussing with @bgfoshay and @ccaprio97 they went ahead with adding regions to the database: (see below PRs for reference)
https://github.com/18F/piipan/pull/3315
https://github.com/18F/piipan/pull/3300

I have some concern about fetching it from the database, but I think we may want to pull regions in using a MemoryCache, like we do for the EnabledStates environment variable here: https://github.com/18F/piipan/blob/dev/match/src/Piipan.Match/Piipan.Match.Func.Api/Api.cs#L72

If we do it this way, we will just need to know we need to restart the services that consume that service if the database ever changes. It sounds like it's very infrequent so we probably can do this, but we would need to make sure to document that well. In the case above if the EnvironmentVariable changes the Azure Function restarts on its own, which causes the MemoryCache to be cleared, so it wasn't an issue there.

bgfoshay: I think the MemoryCache is a good idea and we could set expiration times on the cache entries. Then you can have it automatically re-query once the keys expire.

bgfoshay (approved):
LGTM

PR #3368, opened 2022-06-16
Adding role checks for match detail and match ID search

dborgen2:
## What’s changing?

For the match detail and match ID search, we don't want to show results if the user is using a location that does not match either the initiating or matching state on the found match. If it is the case that they shouldn't have access to the match, we need to log a request to access unauthorized data.

## Why?

Closes NAC tickets 807, 809, 861, and 741

## This PR has:


match/tests/Piipan.Match.Func.ResolutionApi.IntegrationTests/GetMatchApiIntegrationTests.cs
```diff
@@ -117,11 +130,47 @@ public async void GetMatch_ReturnsCorrectSchemaIfFound()
...
             Assert.Equal(expected, resString);
             // Assert the created date that is returned is nearly identical to the actual current time
             Assert.True((createdDate - matchCreateDate).Value.TotalMinutes < 1);
         }
+
+        [Fact]
+        public async void GetMatch_ReturnsCorrectSchemaIfNotAuthorized()
```
bgfoshay, line 140: Nitpick (non-blocking): Is this test verifying Not Found result is returned for an unauthorized user? And if so, should this test be named something like GetMatch_ReturnsNotFoundWhenUnauthorized?

shared/src/Piipan.Shared/Http/AuthorizedJsonApiClient.cs
```diff
@@ -82,9 +82,13 @@ public async Task<TResponse> GetAsync<TResponse>(string path)
             return JsonConvert.DeserializeObject<TResponse>(responseContentJson);
         }
 
-        public async Task<(TResponse Response, int StatusCode)> TryGetAsync<TResponse>(string path)
+        public async Task<(TResponse Response, int StatusCode)> TryGetAsync<TResponse>(string path, Func<IEnumerable<(string, string)>> headerFactory = null)
```
bgfoshay, line 85: Question: Just out of curiosity, is there a benefit to using a headerFactory as opposed to providing a list of header key-value pairs? Are the headers not available at the time this call is made?

match/tests/Piipan.Match.Func.ResolutionApi.IntegrationTests/GetMatchApiIntegrationTests.cs
dborgen2: Yea, good catch. I'll update the name

shared/src/Piipan.Shared/Http/AuthorizedJsonApiClient.cs
dborgen2: Yea, you know, after looking at it more I don't think there is any benefit. I think I just did it because that's what we did for PostAsync, but looking at it it can be simplified. I'll update it to just take in an IEnumerable rather than a Func

bgfoshay (approved):
👍

PR #3369, opened 2022-06-16
Fixing retrieval of states to not break the page if it cannot be retrieved

dborgen2:
## What’s changing?

Previously if the states API failed the entire app would break. This isn't necessary and should be handled gracefully. This will allow cypress tests to work.

## Why?

Cypress tests started failing after we added the states API

## This PR has:

### DIFF
--- a/query-tool/src/Piipan.QueryTool/Pages/BasePageModel.cs
+++ b/query-tool/src/Piipan.QueryTool/Pages/BasePageModel.cs
@@ -16,6 +16,7 @@ namespace Piipan.QueryTool.Pages
     public class BasePageModel : PageModel
     {
         private const string NotAuthorizedPageName = "/NotAuthorized";
+        private const string StateInfoCacheName = "StateInfo";
         private readonly IClaimsProvider _claimsProvider;
         private readonly ILocationsProvider _locationsProvider;
         private readonly IMemoryCache _memoryCache;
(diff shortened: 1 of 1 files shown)


query-tool/tests/Piipan.QueryTool.Tests/BasePageTest.cs
```diff
@@ -48,16 +58,20 @@ public class BasePageTest
                 }
             };
 
-            statesApiMock.Setup(c => c.GetStates()).ReturnsAsync(stateInfoResponse as StatesInfoResponse);
-
-            var mockMemoryCache = new Mock<IMemoryCache>();
-            mockMemoryCache.Setup(n => n.TryGetValue("StateInfo", out stateInfoResponse))
-                .Returns(true);
+            var statesSetup = statesApiMock.Setup(c => c.GetStates());
+            if (statesResponse != null)
```
bgfoshay, line 62: Nitpick: I had to read through a couple times before I realized that statesResponse and stateInfoResponse were different. The logic block makes a lot more sense when I realized that, but at first glance I was confused.

bgfoshay (approved):
LGTM

query-tool/tests/Piipan.QueryTool.Tests/BasePageTest.cs
dborgen2: Yea, I can definitely see that. I'll update the names for clarity

PR #3424, opened 2022-06-24
States API Documentation and Cleanup

bgfoshay:
## What’s changing?

Adding/Improving documentation for States API

## Why?

NAC-932, NAC-934, NAC-935, NAC-945. Adding documentation for recently introduced States API

## This PR has:


states/docs/openapi/get_states.yaml
```diff
@@ -0,0 +1,13 @@
...
+    - "States"
+  summary: "Retrieve metadata for all States"
+  description: "Retrieves metadata records for every State in the Collaboration database"
+  responses:
+    '200':
+      $ref: './schemas/states.yaml#/components/responses/200_Get_States'
+    '401':
+      $ref: './schemas/states.yaml#/components/responses/401'
+    '429':
+      $ref: './schemas/states.yaml#/components/responses/429'
```
kingcomma, line 13: *Nitpick (non-blocking):* Could probably drop this 429 response for now. Some form of rate-limiting on the internal API may not be a bad idea in the future, but I assume it's not yet implemented?

states/docs/openapi/schemas/state-record.yaml
```diff
@@ -0,0 +1,45 @@
...
+    email: "va_agency_worker@gmail.com"
+    phone: "7031234567"
+    region: "Region1"
+  MD:
+    id: 2
+    state: "Maryland"
+    state_abbreviation: "MD"
+    email: "md_agency_worker@gmail.com"
+    phone: "4101234567"
+    region: "Region1"
```
kingcomma, line 44: *Nitpicks:* A few small things —
* Unless there's a reason not to, let's stick to using the EA, EB, EC example states
* I opened up a ticket for this more generally across the repo, but we should stick to TLDs that are reserved for testing purposes for placeholder emails like this. E.g., `no-reply@ea.example` or something similar.
* Similar to emails, we should probably avoid using potentially real phone numbers. I think numbers in the `555-01xx` range are considered safe (https://en.wikipedia.org/wiki/555_(telephone_number))

states/docs/openapi/index.yaml
```diff
@@ -0,0 +1,20 @@
+# Match Resolution API base spec
```
kingcomma, line 1:
```suggestion
# State Metadata API base spec
```

kingcomma (commented):
Looks great overall — I pulled it up using the [OpenAPI preview extension in VS Code](https://marketplace.visualstudio.com/items?itemName=42Crunch.vscode-openapi) and formatting/structure checks out. Just a few small nitpicks, mostly around example data, if you wouldn't mind. 

Really appreciate this getting picked up 👍

states/docs/openapi/get_states.yaml
bgfoshay: Yes, you're right that we can drop this for now. I removed it.

states/docs/openapi/schemas/state-record.yaml
bgfoshay: 👍These all make sense. I'll adjust the examples.

states/docs/openapi/index.yaml
bgfoshay: Whoops. Good catch on this one.

kingcomma (approved):
👍

ccaprio97 (approved):
LGTM

PR #3465, opened 2022-07-04
Adding Final Disposition Component to Match Details Page

ccaprio97:
## This PR has:


query-tool/src/Piipan.QueryTool.Client/Components/MatchDetail/FinalDisposition.razor
```diff
@@ -0,0 +1,23 @@
+﻿@code {
+    [Parameter] public bool InitiatingState { get; set; }
+}
+<div class="FinalDispositionDiv">
+    <h5 class="FinalDispositionHeader">Final Disposition</h5>
+    <p class="FinalDispositionText">Final Disposition information can be added once the initial Action has been selected.</p>
+    <form class="usa-form FinalDispositionDropdown">
```
dborgen2, line 7: Each of these components should not have their own form, as we only have 1 submit button. The form should be on the outside component (ResolutionFields), and should be a <UsaForm> component instead of just a normal <form> component. You can see QueryForm.razor for an example. The purpose for this will be for showing errors and binding to the datamodel.

query-tool/src/Piipan.QueryTool.Client/Components/MatchDetail/FinalDisposition.razor
```diff
@@ -0,0 +1,23 @@
...
+      <label class="usa-label FinalDispositionLabel" for="options">Final Disposition Taken</label>
+      <select class="usa-select FinalDispositionSelect" name="options" id="options">
```
dborgen2, line 9: We should be utilizing the Blazor Component Library instead of hard-coding the Select. If we were to use this as it is, there is no data binding/error checking. There is also no initial value being set if one has been saved.

That being said, I don't think we have a Select component yet. Let me know if you want to do this or I can throw one up there.

query-tool/src/Piipan.QueryTool.Client/Components/MatchDetail/ResolutionFields.razor
```diff
@@ -7,6 +7,6 @@
     <p class="infoSavedText">After the information is saved, it will appear in the <b>match information</b> panel.</p>
     <VulnerableInvalidStatus></VulnerableInvalidStatus>
     <InitialAction InitiatingState="@InitiatingState"></InitialAction>
-    <h4>FINAL DISPOSITION COMPONENT HERE</h4>
-    <button class="usa-button">Save Changes</button>
+    <FinalDisposition InitiatingState="@InitiatingState"></FinalDisposition>
```
dborgen2, line 10: We need to figure out how to pass the current values of the form to each of these subcomponents. As it stands, neither this nor VulnerableInvalidStatus nor InitialAction takes into account the current saved values.

query-tool/src/Piipan.QueryTool.Client/Components/MatchDetail/FinalDisposition.razor.css
```diff
@@ -0,0 +1,35 @@
+﻿body {
+}
+.FinalDispositionHeader {
```
dborgen2, line 3: We shouldn't need to style each of these headers individually. Do we already have a style for this? Or if not, we should create one and use it for all of the h4s on this page

components/src/Piipan.Components.Demo/Pages/InputRadio.razor
```diff
@@ -0,0 +1,17 @@
+﻿@page "/input-radio"
+@inherits BaseComponent
+@code {
+    private InputTextModel TextModel { get; set; } = new InputTextModel();
+    protected override string[] Files { get; set; } = { "components/InputRadioComponent.razor", "models/InputTextModel.cs" };
+}
+
+<h1>Input Text</h1>
```
dborgen2, line 8: Nitpick: Header page should say Input Radio

components/src/Piipan.Components/Forms/UsaSelect.razor
```diff
@@ -0,0 +1,22 @@
+﻿@using static FormConstants
+@inherits UsaInputBase<string>
+@code {
+    [Parameter] public (string, string)[] DropdownOptions { get; set; }
+}
+
+   <select class="@InputSelect @InputClass @(FormGroup.Status != InputStatus.None ? $"{InputClass}--{FormGroup.Status.ToString().ToLower()}" : "")"
```
dborgen2, line 7: Question: I'm not sure if this needs to add something for when the initial value is set. I think that if the initial value is set, the dropdown will be set to the first item, not the selected one.

If that's the case, simply add this to the `option` tag: `selected="CurrentValue == option.Item1"`

query-tool/src/Piipan.QueryTool.Client/Components/MatchDetail/VulnerableInvalidStatus.razor
```diff
@@ -1,86 +1,59 @@
...
-      <legend class="usa-legend usa-legend">
-          <b><svg class="usa-icon usa-tooltip text-blue"
+
+    <UsaRadioGroup @bind-Value="DispositionData.InvalidMatch">
+        <ChildContent>
+            <UsaRadio Value=true>Yes</UsaRadio>
+            @if (DispositionData.InvalidMatch??false)
+            {
+                <UsaFormGroup>
+                    <UsaSelect DropdownOptions="@DropdownOptions" @bind-Value="DispositionData.InvalidMatchReason" />
```
dborgen2, line 36: We may run into an issue on this. If either this select or the text area below it were filled out and then hidden for some reason, we aren't actually clearing the value.

That being said, I don't think we should clear the value until they try to save. I suggest adding something to OnBeforeSubmit in ResolutionFields.razor to clear out those values if necessary. See QueryForm.razor for an example.

query-tool/src/Piipan.QueryTool.Client/Components/MatchDetail/FinalDisposition.razor.css
ccaprio97: doesnt the header number affect the sizing and thickness of the header? im not sure was just following the figma

components/src/Piipan.Components/Forms/UsaSelect.razor
ccaprio97: seemed to work fine locally

match/src/Piipan.Match/Piipan.Match.Api/Models/Resolution/Disposition.cs
```diff
@@ -7,14 +8,23 @@ namespace Piipan.Match.Api.Models.Resolution
     /// </summary>
     public class Disposition
     {
+        [Display(Name = "Inital Action Date")]
         [JsonProperty("initial_action_at")]
         public DateTime? InitialActionAt { get; set; }
         [JsonProperty("initial_action_taken")]
```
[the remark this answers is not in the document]
ccaprio97: it gets used

query-tool/src/Piipan.QueryTool.Client/Components/MatchDetail/VulnerableInvalidStatus.razor
ccaprio97: will do in following pr

match/src/Piipan.Match/Piipan.Match.Api/Models/Resolution/Disposition.cs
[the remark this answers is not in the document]
dborgen2: Yea I think I clicked the wrong line. I meant Initial Action Taken needs a Display label. But you've added that since, so this can get closed.

components/src/Piipan.Components/Forms/UsaSelect.razor
ccaprio97: and by seemed to work fine i added the value to the db and it correctly populated the dropdown with the first and non first value

dborgen2: Yea, you're right. My bad

query-tool/src/Piipan.QueryTool.Client/Components/MatchDetail/FinalDisposition.razor.css
dborgen2: We talked offline about some CSS simplifying in the next PR, so I'm good resolving this for now.

dborgen2 (approved):
LGTM, thanks for all your hard work on this.

PR #3468, opened 2022-07-07
Eth/nac 908 - Add envionmental support for states.csv file

edwintorres:
## What’s changing?

Setting up states.csv as an environment dependable file, so it can be set with different configurations for different environments. 

## Why?

NAC-908

## This PR has:


bgfoshay (commented):
I think this looks really good and the updated documentation is great! But I was wondering whether you think it'd be worthwhile to add an explicit check near the top of create-resources to verify that the desired environment has a states.csv file and fail with a revealing error message if not?

iac/create-resources.bash
```diff
@@ -93,6 +93,7 @@ main () {
   source "$(dirname "$0")"/iac-common.bash
 
   verify_cloud
+  verify_states_file
```
edwintorres, line 96: @bgfoshay verify_states_file was added to check for states.csv file

iac/iac-common.bash
```diff
@@ -191,6 +191,19 @@ verify_cloud () {
...
 
+# Verify that the states.csv file exist for the current enviroment 
+verify_states_file () {
+
+  local FILE
+  FILE=env/"${azure_env}"/states.csv
+  if [ -f "$FILE" ]; then
+      echo "$FILE exists."
+  else 
+      echo "$FILE does not exist. The states.csv file is required."
```
bgfoshay, line 202: Nitpick: I know this is super picky, but can we change this to something like "$FILE does not exist. A states.csv file is required to be present in env/"${azure_env}"".

edwintorres: This is the last version that follow your suggestion.

https://github.com/18F/piipan/blob/585464d72a5d85903f1d886e84afbe2ee85513b4/iac/iac-common.bash#L194-L211

bgfoshay (approved):
Thank you for accommodating all my nitpicks! This looks good to merge.

PR #3470, opened 2022-07-07
Modal and Navigation blocking behaviors

dborgen2:
## What’s changing?

We are adding the Modal component, as well as Navigation blocking behavior for when forms have been updated but not saved. These "dirty" forms should ask the user if they want to continue filling out the form or if they want to leave the page. This behavior is found in the ConfirmModalWrapper. Simply put an existing UsaForm inside a ConfirmModalWrapper component and the navigation blocking will happen.

The navigation blocking itself is done using a combination of an overridden NavigationManager called PiipanNavigationManager and an overriden Router called PiipanRouter. These two classes are among those [here](https://github.com/dotnet/aspnetcore/tree/v6.0.6/src/Components/Components/src).

Since the [Router](https://github.com/dotnet/aspnetcore/blob/v6.0.6/src/Components/Components/src/Routing/Router.cs) uses classes internal to AspNetCore, we had to recreate those classes. They are simple copy/pastes and a namespace change from the repository. The reason we need the router is to hijack when the location is about to change and perform a logic check on it. The Location Changing event is something [being worked on](https://github.com/dotnet/aspnetcore/issues/14962) by the AspNet team, with it currently slated to go out in .Net7, so at that time we may not need to have a custom NavigationManager/Router. At that time most of the logic in the Piipan.Components.Routing namespace can likely be replaced, and the ConfirmModalWrapper could listen to LocationChanging instead of NavigationBlocked.

## Why?

Closes NAC 446

## This PR has:

### DIFF
--- a/components/src/Piipan.Components.Demo/Components/DemoModalComponent.razor
+++ b/components/src/Piipan.Components.Demo/Components/DemoModalComponent.razor
@@ -1,0 +1,43 @@
+﻿@using Piipan.Components.Modals
+@inject IModalManager ModalManager
+
+@code {
+    [CascadingParameter] public ModalInfo ModalInfo { get; set; }
+    [Parameter] public Action ContinueCallback { get; set; }
+    private void Close()
+    {
+        ModalManager.Close(ModalInfo);
+    }
+    private void Continue()
+    {
+        ModalManager.Close(ModalInfo);
+        if (ContinueCallback != null)
+        {
+            ContinueCallback?.Invoke();    
+        }
+    }
+}
+
+
+<h2 class="usa-modal__heading" id="modal-1-heading">
+    Continue Without Saving?
+</h2>
+<div class="usa-prose">
+    <p id="modal-1-description">
+    Any unsaved changes will be lost.
+    </p>
+</div>
+<div class="usa-modal__footer">
+    <ul class="usa-button-group">
+    <li class="usa-button-group__item">
+        <button type="button" class="usa-button" @onclick="Continue">
+        Continue
+        </button>
+    </li>
+    <li class="usa-button-group__item">
+        <button type="button" class="usa-button usa-button--unstyled padding-105 text-center" @onclick="Close">
+        Cancel and return to form
+        </button>
+    </li>
+    </ul>
+</div>
--- a/components/src/Piipan.Components/Forms/UsaForm.razor.cs
+++ b/components/src/Piipan.Components/Forms/UsaForm.razor.cs
@@ -1,16 +1,18 @@
-﻿using Microsoft.AspNetCore.Components;
-using Microsoft.AspNetCore.Components.Forms;
-using Microsoft.JSInterop;
-using System;
+﻿using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
+using Microsoft.AspNetCore.Components;
+using Microsoft.AspNetCore.Components.Forms;
+using Microsoft.JSInterop;
+using Piipan.Components.Routing;
 
 namespace Piipan.Components.Forms
 {
     public partial class UsaForm
     {
         [Inject] protected IJSRuntime JSRuntime { get; set; } = default!;
+        [Inject] protected PiipanNavigationManager NavigationManager { get; set; }
         private bool HasErrors => currentErrors?.Count() > 0;
         private EditContext editContext;
         private bool showAlertBox = false;
@@ -19,6 +21,18 @@ namespace Piipan.Components.Forms
         public List<UsaFormGroup> FormGroups { get; set; } = new List<UsaFormGroup>();
         private List<(UsaFormGroup FormGroup, IEnumerable<string> Errors)> currentErrors =
             new List<(UsaFormGroup FormGroup, IEnumerable<string> Errors)>();
+
+        private bool _isDirty = false;
+        public bool IsDirty
+        {
+            get => _isDirty;
+            set
+            {
+                _isDirty = value;
+                IsDirtyChanged?.Invoke(_isDirty);
+            }
+        }
+        public Func<bool, Task> IsDirtyChanged { get; set; }
 
         /// <summary>
         /// Set the edit context of this form when it's initialized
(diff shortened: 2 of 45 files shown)


components/src/Piipan.Components.Demo/Components/DemoModalComponent.razor
```diff
@@ -0,0 +1,43 @@
...
+    Continue Without Saving?
+</h2>
+<div class="usa-prose">
+    <p id="modal-1-description">
+    Any unsaved changes will be lost.
+    </p>
+</div>
+<div class="usa-modal__footer">
+    <ul class="usa-button-group">
+    <li class="usa-button-group__item">
```
bgfoshay, line 32: Nitpick (non-blocking): should the list items be idented?

query-tool/src/Piipan.QueryTool/Pages/Shared/_Layout.cshtml
```diff
@@ -15,6 +15,7 @@
     var pageName = Context.Request.Path.ToString().Split('/').Last().ToLower();
 }
 <body>
+    <div id="inert-area">
```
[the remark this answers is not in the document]
dborgen2: Good catch, but actually there was an extra closing div already in the _Layout, so I just used that.

components/src/Piipan.Components/Forms/UsaForm.razor.cs
```diff
@@ -99,6 +115,11 @@ public async Task PresubmitForm()
             if (OnBeforeSubmit != null)
             {
                 await OnBeforeSubmit(currentErrors.Count == 0);
+                if (currentErrors.Count == 0)
+                {
+                    // Form will submit now, so let's mark the form dirty so it doesn't get flagged that we're leaving the page without saving
```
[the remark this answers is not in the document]
dborgen2: Yep, I'll update

components/src/Piipan.Components/Modals/ModalContainer.razor.js
```diff
@@ -0,0 +1,39 @@
...
+
+// Sets the main page to inert, meaning focus cannot go inside it. Only the last modal can have focus go inside it.
+// Also, we add an escape button listener to the ModalContainer, which closes the last opened modal.
+export function SetFocusTrapsAndEscapeListener(callingComponent) {
+    dotNetReference = callingComponent;
+    const modalWrappers = document.querySelectorAll('.usa-modal-wrapper');
+    const inertArea = document.getElementById('inert-area');
+    if (modalWrappers.length === 0) {
+        inertArea?.removeAttribute('inert');
+        if (modalsOpened) {
```
bgfoshay, line 17: Question: Apologies as I'm trying to wrap my head around this block of logic? A modal wrapper is not being used with the ModalContainer but an existing Modal is already open?

dborgen2: Yea, good question, it's confusing. So the ModalContainer Blazor Component is always there, showing modals if they exist. This SetFocusTrapsAndEscapeListener gets called anytime the ModalContainer gets re-rendered (eg. a modal is added or removed).

So if we are calling this method with no modals, we need to do 2 things:

1. Remove the "inert" attribute from the main layout, allowing items inside the main layout to be able to receive focus
2. If last time we called this method we had an open modal, we need to unsubscribe to the escape key since now we have no open modals.

components/src/Piipan.Components/Routing/PiipanRouter.cs
```diff
@@ -0,0 +1,279 @@
...
+
+        private CancellationTokenSource _onNavigateCts;
+
+        private Task _previousOnNavigateTask = Task.CompletedTask;
+
+        private RouteKey _routeTableLastBuiltForRouteKey;
+
+        private bool _onNavigateCalled;
+
+        // Piipan Update: Switch NavigationManager to BlockableNavigationManager
```
bgfoshay, line 48: Question/Nitpick: Did you mean switch to PiipanNavigationManager?

components/src/Piipan.Components/Modals/ModalManager.cs
```diff
@@ -0,0 +1,50 @@
...
+                int sequenceNum = 2;
+                foreach (var prop in properties)
+                {
+                    object value = prop.GetValue(modal);
+                    n.AddAttribute(sequenceNum++, prop.Name, value);
+                }
+
+                n.CloseComponent();
+            });
+            OpenModals.Add(modalInfo);
```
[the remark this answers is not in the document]
dborgen2: Good point. Switching OpenModals to be a ReadOnlyCollection with a private setter. We should not be allowing the code to call Add and Remove, bypassing the rest of the Show/Hide logic.

By doing this, they'll be forced to go through the Show logic, and this logic already has something in it to make sure it's not null.

components/src/Piipan.Components/Routing/PiipanRouter.cs
```diff
@@ -0,0 +1,279 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+// Taken from https://github.com/dotnet/aspnetcore/tree/v6.0.6/src/Components/Components/src with a few modifications
```
bgfoshay, line 4: Question: Are there any other modifications to this class aside from the PiipanNavigationManager? Thank you for calling out that mod explicitly. That was helpful.

components/src/Piipan.Components/Routing/PiipanRouter.cs
dborgen2: I did, yea. Originally called it BlockableNavigationManager and switched the name at one point. Updated

query-tool/src/Piipan.QueryTool.Client/Modals/ConfirmModalWrapper.razor
```diff
@@ -0,0 +1,84 @@
...
+    /// After this is rendered (and thus the Web Assembly & Javascript runtime is available), import the ConfirmModalWrapper javascript.
+    /// This javascript includes a call to checking beforeunload, which is called when a user tries to navigation away from our app
+    /// or close the window entirely.
+    /// </summary>
+    protected override async Task OnAfterRenderAsync(bool firstRender)
+    {
+        await base.OnAfterRenderAsync(firstRender);
+        if (modalJavascriptReference == null)
+        {
+            modalJavascriptReference = await JSRuntime.InvokeAsync<IJSObjectReference>("import", "./Modals/ConfirmModalWrapper.razor.js");
```
bgfoshay, line 32: Question: Assuming there is, but is there a test that will fail if someone goes and renames ConfirmModalWrapper.razor.js without changing it here?

PR #3483, opened 2022-07-11
Created configuration environment script

edwintorres:
## What’s changing?

Added an environment configuration script to setup variables, files and other specific environment configuration parameters. 

## Why?

NAC-943. Add support for environment state metadata inserts.Goal is to support different emails/phone. numbers for notifications in our different environments.
## This PR has:


iac/env/tts/dev/insert-state-info.sql
```diff
@@ -0,0 +1,50 @@
...
+DURING THE IAC PROCESS THIS FILE IS OVERWRITE WITH THE 
+COPY ON THE ENVIRONMENT FOLDER
+
+*/
+
+
+BEGIN;
+
+INSERT INTO state_info(id, state, state_abbreviation, email, phone, region)
+select * from (select '15' as id, 'Iowa' as state, 'IA' as state_abbreviation, 'IA-test@usda.gov' as email, '1234567890' as phone, 'MWRO' AS region) as temp
```
bgfoshay, line 16: Nitpick: While we're refactoring the support for this script, we should change over the fake emails to using a testing TLD rather than usda.gov (https://datatracker.ietf.org/doc/html/rfc2606)

match/dml/insert-state-info.sql
```diff
@@ -1,3 +1,15 @@
...
+
+DURING THE IAC PROCESS THIS FILE IS OVERWRITE WITH THE 
+COPY ON THE ENVIRONMENT FOLDER
+
+*/
+
+
 BEGIN;
 
 INSERT INTO state_info(id, state, state_abbreviation, email, phone, region)
```
bgfoshay, line 15: Question: Are there any scenarios where this script could potentially get run, and if so will this script fail?

match/dml/insert-state-info.sql
```diff
@@ -1,3 +1,15 @@
...
+THIS FILE MUST BE MODIFIED ON THE ENVIRONMENT FOLDER
+/piipan/iac/env/ PLEASE CHECK THE DEVELOPMENT EXAMPLE 
+ON /piipan/iac/env/tts/dev/insert-state-info.sql
+
+DURING THE IAC PROCESS THIS FILE IS OVERWRITE WITH THE 
+COPY ON THE ENVIRONMENT FOLDER
+
+*/
+
+
```
bgfoshay, line 12: Question: Are there any scenarios where this script could potentially get run, and if so will this script fail?

iac/env/tts/dev/config.bash
```diff
@@ -0,0 +1,10 @@
+# ENVIRONMENT CONFIGURATION SCRIPT
+
+main () {
+
+    echo "COPYING... ./env/tts/dev/insert-state-info.sql ../../../match/dml/insert-state-info.sql"
```
bgfoshay, line 5: Question: Similar to the check you added in setup_enviroment in iac-common.bash, should we have a verification check here (and useful error message) if the user has not provided insert-state-info.sql file? Or are we ok proceeding without one?

edwintorres: This is a design decision. The approach in this case is that there is a default file in /piipan/match/dml/ if the process found a file in the environment folder the default file is going to be replace. I need to improve the copy functionality so it manage better the scenario when there is NOT an environment file. The process will run the default script in case that there is not an environment file.  If this approach make sense then it can be replicated with states.csv

bgfoshay: 👍That makes sense. I just wanted to double-check the scenario when the default script gets run.

edwintorres: I added some validation to the copying of the environment files.

https://github.com/18F/piipan/blob/bc86fda6d6e79c7010996e17b308ec97fb6c7b74/iac/env/tts/dev/config.bash#L4-L19

match/dml/insert-state-info.sql
edwintorres: I'm not sure if I address this one with the update to `config.bash` to validate the copying

iac/env/tts/dev/config.bash
```diff
@@ -0,0 +1,26 @@
...
+
+
+safe_copy () {
+    local OLD_FILE
+    local NEW_FILE
+
+    OLD_FILE=$1
+    NEW_FILE=$2
+
+    if [ -f "$OLD_FILE" ] && [ -f "$NEW_FILE" ]; then
```
bgfoshay, line 11: Question: Do you need to check if the new file exists prior to the copy? It should get created as a result of the copy command, right?

edwintorres: My idea was I want to replace a current file, so I have to check that the file really exists and that I really have the new file to replace it. In the case of `insert-state-info.sql` is used by other bash scripts, project definitions, and C# files so if the default file is not there the tests will fail. 

Thinking deeply about how developers are going to use this. I can assume that they will copy the env folder and then rename it and proceed to change the values on the SQL script and CSV files. In that case, there is a problem because the path is hardcoded `/env/tts/dev` on the `config.bash`.  I will change that.

Still not sure if change the `safe_copy` to only validate the NEW_FILE. Let me know if you have more thoughts on this.

bgfoshay: That makes sense. That's a good idea to fix the path because you're right, developers will likely copy an env folder. 

In regards to my comment, I was suggesting maybe only validating the OLD_FILE. If NEW_FILE doesn't exist, it'll get created. If it does exist, it'll get overwritten.

edwintorres: I changed the variables names, to better understanding

https://github.com/18F/piipan/blob/f46477401d862ef9b434d5c7fa2cb870a15da94e/iac/env/tts/dev/config.bash#L4-L18

Now the function receive the environment

https://github.com/18F/piipan/blob/f46477401d862ef9b434d5c7fa2cb870a15da94e/iac/env/tts/dev/config.bash#L20-L26

bgfoshay: Ah, got it now. 👍

PR #3495, opened 2022-07-18
Showing/Hiding Match Detail edit section based off user's role

dborgen2:
## What’s changing?

The edit section of the Match Detail page should only be shown to users with certain roles (see NAC-501).
The roles that are able to edit the Match Detail page will be configurable via the appsettings.json. For development purposes, we are allowing users with the role of "Worker".

## Why?

Closes NAC-501

## This PR has:

### DIFF
--- a/query-tool/src/Piipan.QueryTool/Pages/Match.cshtml.cs
+++ b/query-tool/src/Piipan.QueryTool/Pages/Match.cshtml.cs
@@ -1,13 +1,16 @@
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Mvc;
+using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
 using Piipan.Match.Api;
 using Piipan.Match.Api.Models;
 using Piipan.Match.Api.Models.Resolution;
 using Piipan.QueryTool.Client.Models;
+using Piipan.Shared.Roles;
 
 namespace Piipan.QueryTool.Pages
 {
@@ -16,6 +19,7 @@ namespace Piipan.QueryTool.Pages
 
         private readonly ILogger<MatchModel> _logger;
         private readonly IMatchResolutionApi _matchResolutionApi;
+        private readonly IRolesProvider _rolesProvider;
 
         [BindProperty]
         public MatchSearchRequest Query { get; set; } = new MatchSearchRequest();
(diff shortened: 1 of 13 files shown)


query-tool/src/Piipan.QueryTool/Pages/Match.cshtml.cs
```diff
@@ -125,54 +131,63 @@ public async Task<IActionResult> OnPost([FromRoute] string id)
...
-                foreach (var modelStateKey in ModelState.Keys)
+                if (!_rolesProvider.GetMatchEditRoles().Contains(Role))
                 {
-                    if (!modelStateKey.Contains(nameof(DispositionData)))
-                    {
-                        ModelState[modelStateKey].Errors.Clear();
-                        ModelState[modelStateKey].ValidationState = Microsoft.AspNetCore.Mvc.ModelBinding.ModelValidationState.Valid;
-                    }
+                    _logger.LogError($"User {Email} does not have permissions to edit match details.");
+                    RequestErrors.Add(new("", "You lack the necessary role to save match details."));
```
bgfoshay, line 137: Question: Does this message get sent back to users? If so, we may want to double-check with Andrew/PMs to make sure they're happy with the text.

shared/tests/Piipan.Shared.Tests/Roles/RolesProviderTests.cs
```diff
@@ -0,0 +1,31 @@
+﻿using Microsoft.Extensions.Options;
+using Piipan.Shared.Roles;
+using Xunit;
+
+namespace Piipan.Shared.Tests.Roles
+{
+    public class RolesProviderTests
```
bgfoshay, line 7: Question: Do we need a test for when the roleOptions are empty/null?

query-tool/src/Piipan.QueryTool/Pages/Match.cshtml.cs
dborgen2: It's a very rare case. It would happen only when the client side did NOT hide the fields (maybe you were an eligibility user and somehow you lost that permission, or maybe there was an application error where the form was shown on accident, or maybe you are trying to hack something). In this case, yes this would get sent back to the user. 

After talking with Corey he wanted to switch the wording slightly for the off-chance it could occur.

shared/tests/Piipan.Shared.Tests/Roles/RolesProviderTests.cs
dborgen2: Yea, I think that's a valid test. Updated

bgfoshay (approved):
👍

PR #3501, opened 2022-07-20
Fixing Match Resolution Closure 

ccaprio97:
## What’s changing?

_For External API changes, add the changelog tag to the PR and some CHANGELOG text here_

## Why?

NAC-998

## This PR has:


ccaprio97:
Reviewers, please check over the test update that i made to confirm that it is testing right. I am not 100% that I did it the "correct" way or just a way that makes it pass. 

Originally there was setup and then the act part was adding it I guess? but the functionality was all wrong before so no i am adding both states at the beginning and then when it goes through the "act" section it doesnt actually add anything but it still kicks off the method so the method does a check for closed and closes it. 

TLDR 
the test works now but the Act section is a lie

match/tests/Piipan.Match.Func.ResolutionApi.Tests/AddEventApiTests.cs
```diff
@@ -331,8 +331,12 @@ public async void AddEvent_ChecksIfShouldBeClosedOnSuccess()
             var matchResEventDao = new Mock<IMatchResEventDao>();
             var events = new List<IMatchResEvent>() {
                 new MatchResEventDbo() {
-                    Delta = "{ \"final_disposition\": \"foo\" }",
+                    Delta = "{ \"final_disposition\": \"foo\", \"final_disposition_date\": \"2022-07-20T00:00:02\" }",
                     ActorState = "eb"
+                },
+                new MatchResEventDbo() {
```
dborgen2, line 337: I feel like this shouldn't be added here. events is a list of records already in the database. If we have the one from eb, and then we add the new one, shouldn't that be sufficient enough to close it?

ccaprio97: ya so this is what i made my comment about. It SHOULD be you add one. then in the act section you add the other and then it automatically adds the closed event. 

the problem was that because it wasnt connected to the db or something the addevent method returns a 0. 
then we have that if AddEventReturnedValue is > 0  so it doesnt add the event to the thing so it doesnt close. 

if that doesnt make sense call me

dborgen2: Yea, I think that makes sense. So the root of the problem is that the AddEvent method is returning 0... I'm assuming it's because it was never mocked? A non-mocked method will just return the default return value I believe.

Perhaps adding something like the following would do the trick?
 ```            
 matchResEventDao
  .Setup(r => r.AddEvent(
      It.IsAny<MatchResEventDbo>()
  ))
  .ReturnsAsync(1);

dborgen2 (approved):
👍

PR #3505, opened 2022-07-22
Initial commit for final disposition updates

dborgen2:
## What’s changing?

Added the following checks client side and in the API
- Initial Action Taken cannot be null when Final Disposition is filled out
- Initial Action Date cannot be null when Final Disposition is filled out
- Final Disposition Date is required when Final Disposition is filled out
- Final Disposition is required when Final Disposition Date is filled out
- Final Disposition Date cannot be less than the match date

In addition, the following functionality was added:

- When Initial Action Taken isn't filled out, the Final Disposition area is disabled
- If Initial Action Taken is filled out, and Final Disposition is filled out, and then Initial Action Taken is cleared, Final Disposition will not be disabled. Instead an error will be thrown (see first two bullets above)
- Updated the PatchAsync response to not simply return a generic error. It will display the error returned from the API in a JSON failed response object. Then it's up to the client application to decide what to do with it. In our case, we add it to the server errors on top of the Match Detail page.
- If server-side errors do happen, we no longer throw away the values we tried submitting. We put them back in the form (but not on the right side)
- Default the Final Disposition Date when "Benefits Unchanged" or "Benefits Denied" was selected.

TL;DR
Final Disposition dropdown and date error validation, and disabling the section when Initial Action isn't yet filled in.

## Why?

Closes NAC-74 & NAC-592

## This PR has:

### DIFF
--- a/components/tests/Piipan.Components.Tests/Validation/UsaRequiredIfAttributeTests.cs
+++ b/components/tests/Piipan.Components.Tests/Validation/UsaRequiredIfAttributeTests.cs
@@ -4,54 +4,172 @@ using Xunit;
 namespace Piipan.Components.Tests.Validation
 {
     /// <summary>
-    /// Tests associated with the UsaRequired attribute
+    /// Tests associated with the UsaRequiredIf attribute
     /// </summary>
     public class UsaRequiredIfAttributeTests
     {
         public class DummyModel
         {
-            public string SourceString { get; set; }
+            public string Prop1 { get; set; }
 
-            [UsaRequiredIf(nameof(SourceString))]
-            public string TargetString { get; set; }
+            [UsaRequiredIf(nameof(Prop1), "", "some error")]
+            public string Prop2 { get; set; }
+
+            [UsaRequiredIf(
+                nameof(Prop1), "", "some error",
+                nameof(Prop2), "", "some other error")]
+            public string Prop3 { get; set; }
         }
 
         /// <summary>
         /// Validate a UsaRequiredIf attribute attached to a dummy model using various parameters
         /// </summary>
-        /// <param name="value">The value of the associated field</param>
-        /// <param name="isValidExpected">Whether or not we expect this field to have an error after the value is entered</param>
         [Theory]
-        [InlineData("", "test", null, null)]
-        [InlineData(null, "test", null, null)]
-        [InlineData("", "", null, null)]
-        [InlineData(null, null, null, null)]
-        [InlineData("test", "", null, "Target Value cannot be empty because source value is not null")]
-        [InlineData("test", "test", null, null)]
-        [InlineData("expected", "", "expected", "Target Value cannot be empty because source value is set to the expected value")]
-        [InlineData("expected", "test", "expected", null)]
-        [InlineData("notexpected", "", "expected", null)]
-        [InlineData("notexpected", "test", "expected", null)]
-        public void UsaRequiredIf_ValidateTests(string sourceValue, string targetValue, string requiredIfEqualText, string errorText)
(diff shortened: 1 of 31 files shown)


components/tests/Piipan.Components.Tests/Validation/UsaRequiredIfAttributeTests.cs
```diff
@@ -4,54 +4,172 @@
...
+            Assert.Equal(hasError, !string.IsNullOrEmpty(validationResult?.ErrorMessage));
+        }
+
+        /// <summary>
+        /// Validate a UsaRequiredIf attribute attached to a dummy model using various parameters against an expected value
+        /// </summary>
+        [Theory]
+        [InlineData("expectedvalue", "test", false)]
+        [InlineData("expectedvalue", "", true)] // Error thrown when property 1 has expectedvalue and property 2 is empty
+        [InlineData("notexpectedvalue", "", false)] // Error NOT thrown when property 1 has an unexpected value and property 2 is empty
```
ccaprio97, line 57: does this equate to, we dont throw an error if theres bad data in intial action and no value in final disposition?

match/src/Piipan.Match/Piipan.Match.Core/Parsers/AddEventRequestParser.cs
```diff
@@ -60,6 +60,10 @@ public async Task<AddEventRequest> Parse(Stream stream)
 
                 return request;
             }
+            catch (ValidationException)
+            {
+                throw;
```
ccaprio97, line 65: do we not want to throw a certain type?

components/tests/Piipan.Components.Tests/Validation/UsaRequiredIfAttributeTests.cs
dborgen2: No, good question. This right here is for verifying that if the source property has an expected value then we should be required. This actually isn't used in the app yet, but it is coming. This will be for the Invalid Match error stuff, and making sure that the Invalid Match For Other reason is required when Invalid Match Reason is "Other".

match/tests/Piipan.Match.Func.ResolutionApi.Tests/AddEventApiTests.cs
```diff
@@ -115,13 +115,14 @@ public async void AddEvent_ReturnsUnauthorizedIfMatchClosed()
             var matchRecordDao = new Mock<IMatchRecordDao>();
             matchRecordDao
                 .Setup(r => r.GetRecordByMatchId(It.IsAny<string>()))
-                .ReturnsAsync(new MatchRecordDbo() { States = new string[] { "ea", "bc"}});
+                .ReturnsAsync(new MatchRecordDbo() { States = new string[] { "ea", "bc" } });
```
ccaprio97, line 118: dont know why spacing was added but not an issue

ccaprio97 (approved):
didnt see anything worthy of blocking so LGTM

components/tests/Piipan.Components.Tests/Validation/UsaRequiredIfAttributeTests.cs
ccaprio97: gotcha

match/tests/Piipan.Match.Func.ResolutionApi.Tests/AddEventApiTests.cs
dborgen2: Mm.. yea, probably the code cleanup that happens on save added it. Kind of annoying but should be fine

PR #3509, opened 2022-07-25
Capture Searchi Api calls in Metrics.

kcherukupalli:
## What’s changing?

When a user performs a search, a new entry is added to the Metrics database.

## Why?

@nac-579

## This PR has:

- [x ] **Commented source code** (_required on public classes and methods, and elsewhere as appropriate_)


match/src/Piipan.Match/Piipan.Match.Core/Services/IParticipantPublishSearchMetric.cs
```diff
@@ -0,0 +1,14 @@
+﻿using Piipan.Metrics.Api;
+using System;
```
dborgen2, line 2: Looks like 4 of these using statements are unnecessary, and are not sorted. Do you have Visual Studio set to run code cleanup on save and put System usings first?

match/src/Piipan.Match/Piipan.Match.Core/Services/MatchEventService.cs
```diff
@@ -21,30 +23,33 @@ public class MatchEventService : IMatchEventService
...
-        /// Evaluates each new match in the incoming `matchRespone` against any existing matching records.
-        /// If an open match record for the match exists, it is reused. Else, a new match record is created.
-        /// Each match is subsequently updated to include the resulting `match_id`.
-        /// </summary>
-        /// <param name="request">The OrchMatchRequest instance derived from the incoming match request</param>
-        /// <param name="matchResponse">The OrchMatchResponse instance returned from the match API</param>
-        /// <param name="initiatingState">The two-letter postal abbreviation for the state initiating the match request</param>
-        /// <returns>The updated `matchResponse` object with `match_id`s</returns>
-        public async Task<OrchMatchResponse> ResolveMatches(OrchMatchRequest request, OrchMatchResponse matchResponse, string initiatingState)
+            /// <summary>
```
dborgen2, line 43: Nitpick (non-blocking: Extra spaces here cause things to be mis-aligned in this file. I assume they can be removed.

match/src/Piipan.Match/Piipan.Match.Core/Services/MatchEventService.cs
```diff
@@ -53,7 +58,23 @@ public async Task<OrchMatchResponse> ResolveMatches(OrchMatchRequest request, Or
...
+            //Build Search Metrics
+            ParticipantSearchMetrics participantSearchMetrics = new ParticipantSearchMetrics();
+            participantSearchMetrics.Data = new List<ParticipantSearch>();
+            foreach (OrchMatchResult requestPerson in matchResponse.Data.Results)
+            {
+                var participantUploadMetrics = new ParticipantSearch()
+                {
+                    State = initiatingState,
+                    SearchedAt = DateTime.UtcNow,
+                    SearchFrom = String.Empty,//Need to Identify Website/Api call
```
dborgen2, line 70: Question: Do you need to fill in the SearchFrom on this ticket? Or is there a separate ticket for that?

kcherukupalli: This will be added in future ticket

match/src/Piipan.Match/Piipan.Match.Core/Services/MatchEventService.cs
```diff
@@ -53,7 +58,23 @@ public async Task<OrchMatchResponse> ResolveMatches(OrchMatchRequest request, Or
...
+                    SearchReason = request.Data.ElementAt(requestPerson.Index).SearchReason,
+                    MatchCreation = String.IsNullOrEmpty(String.Join(",", requestPerson.Matches.Select(p => p.MatchCreation))) ? EnumHelper.GetDisplayName(SearchMatchStatus.MATCHNOTFOUND) : EnumHelper.GetDisplayName(SearchMatchStatus.EXISTINGMATCH),
```
dborgen2, line 72: Question: Just want to double check this logic. It looks like MatchCreation is set no matter what in ResolveSingleMatch, either to "New Created Match" or "Already Existing Match". We aren't checking the values here, only checking if it's null or empty, and if it's not empty it gets set to EXISTINGMATCH. In addition, if we have multiple matches we should have both match creation statuses here, correct?

I would think this should say something like:
MatchCreation = string.Join(",", requestPerson.Matches.Select(p => p.MatchCreation)) ?? EnumHelper.GetDisplayName(SearchMatchStatus.MATCHNOTFOUND).

But I could easily be missing something?

match/src/Piipan.Match/Piipan.Match.Core/Services/ParticipantPublishSearchMetric.cs
```diff
@@ -0,0 +1,90 @@
...
+                    {
+                        _client = new EventGridPublisherClient(
+                            new Uri(eventGridUriString),
+                            new AzureKeyCredential(eventGridKeyString),
+                            default
+                        );
+                    }
+                    else
+                    {
+                        _logger.LogWarning("EventGridMetricSearchEndPoint and EventGridMetricSearchKeyString environment variables are not set.");
```
dborgen2, line 43: Question: Just double checking that this shouldn't be a LogError, since it seems like it could cause some big issues.

match/tests/Piipan.Match.Core.Tests/Services/MatchEventServiceTests.cs
```diff
@@ -131,6 +146,12 @@ public async void Resolve_AddsSingleRecord()
                     r.Initiator == record.Initiator &&
                     r.States.SequenceEqual(record.States))),
                 Times.Once);
+
+            participantPublishSearchMetric.Verify(r => r.PublishSearchdMetric(
+              It.Is<ParticipantSearchMetrics>(r => r.Data.FirstOrDefault().MatchCount == search.MatchCount &&
```
dborgen2, line 151: Nitpick (non-blocking): If Data is empty, this will error out with a null reference exception. Looks like it's not possible with this test, so in that case probably just use First(), as FirstOrDefault() implies that it could return null.

match/tests/Piipan.Match.Core.Tests/Services/MatchEventServiceTests.cs
```diff
@@ -389,12 +424,17 @@ public async void Resolve_InsertsNewMatchIdIfMostRecentRecordIsClosed()
                 Status = MatchRecordStatus.Closed
             });
 
+            var participantPublishSearchMetric = new Mock<IParticipantPublishSearchMetric>();
+            participantPublishSearchMetric.Setup(m => m.PublishSearchdMetric(It.IsAny<ParticipantSearchMetrics>()))
+                               .Returns(Task.CompletedTask);
+
```
dborgen2, line 430: Do we need to add an Assert to make sure PublishSearchdMetric was called like you did previously?

match/tests/Piipan.Match.Core.Tests/Services/ParticipantPublishSearchMetricTests.cs
```diff
@@ -0,0 +1,80 @@
...
+                MatchCount = It.IsAny<int>(),
+                SearchedAt = DateTime.UtcNow
+
+            };
+            participantSearchMetrics.Data.Add(metric);
+
+            // Act
+            await Assert.ThrowsAsync<NullReferenceException>(() => participantPublishUploadMetric.PublishSearchdMetric(participantSearchMetrics));
+
+            logger.Verify(x => x.Log(
```
dborgen2, line 72: Do we also need to verify this log that gets thrown?
`_logger.LogWarning("EventGridMetricSearchEndPoint and EventGridMetricSearchKeyString environment variables are not set.");`

metrics/ddl/metrics.sql
```diff
@@ -13,4 +13,15 @@ CREATE TABLE IF NOT EXISTS participant_uploads(
 
 COMMENT ON TABLE participant_uploads IS 'Participant bulk upload event record';
 
+CREATE TABLE IF NOT EXISTS participant_searchs(
```
dborgen2, line 16: I think there's a minor spelling mistake here. Should this be called "participant_searches"

dborgen2 (approved):
👍

PR #3545, opened 2022-08-04
NAC-944: Added third parameter for table name with default value

edwintorres:
## What’s changing?

Added parameter with the table name to two functions on db_common to assign access level. This third parameter is optional, the default value is ALL TABLES. 

## Why?

NAC-944 Add additional functionality to support adding different levels of access at the table-level. For example, authorizing the state metadata API to only have access to the state_info table and no other tables in the collaboration database

## This PR has:

### DIFF
--- a/iac/db-common.bash
+++ b/iac/db-common.bash
@@ -151,26 +151,40 @@ EOF
 }
 
 # Database access controls
+# Example grant access to all tables
+#   db_grant_read "DB_Name" "APP_Name"
+# Example grant access to some tables
+#   db_grant_read "DB_Name" "APP_Name" "table1, table2"
+# Example grant access a table
+#   db_grant_read "DB_Name" "APP_Name" "tableName"
+
 db_grant_read () {
   local db=$1
   local func=$2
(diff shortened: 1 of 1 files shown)


edwintorres:
@ryangeorgeharvey Not sure who I should put as reviewer in this one. I usually put Bill.

dborgen2 (commented):
I think it looks good. Quick question: Do we need to account for when the user wants to give access to multiple tables (but not all of them)? Do they need to call this function more than once, passing in multiple values, or should they be able to pass a list of tables as this third parameter.

edwintorres:
> I think it looks good. Quick question: Do we need to account for when the user wants to give access to multiple tables (but not all of them)? Do they need to call this function more than once, passing in multiple values, or should they be able to pass a list of tables as this third parameter.

@dborgen2  I added some comments to explain how to use it with one table, multiple tables or all tables. There also was a error when used with multiple tables, it is fix.

dborgen2 (approved):
Ok, great. Thanks for adding multiple tables in. LGTM

PR #3591, opened 2022-08-08
Adding vulnerable individual modal, as well as match detail back button

dborgen2:
## What’s changing?

There is a new modal that pops up when the match has a vulnerable individual associated with it. It pops up in these 3 cases:

1. The user performs a duplicate participation query search and clicks the returned match.  This pops it up on the query search screen, and upon clicking continue, the user goes to the match detail screen.
2. The user performs a match ID search and clicks the returned match. This pops it up on the match search screen, and upon clicking continue, the user goes to the match detail screen.
3. The user goes to the match detail screen from a different workflow than above. This pops it up on the match detail screen.

In all 3 cases, the modal has a forced action, meaning it cannot be closed. The Continue button is the only selection.

In addition, the new back button was added to the match detail screen. When coming from the duplicate participation search, clicking this will take you back to that screen. Same for the match ID search. When coming from someplace other than these two screens, the button does not appear.

## Why?

Closes NAC 175 and 701

## This PR has:


kcherukupalli (approved):
Looks good to me.

PR #3645, opened 2022-08-10
Update dependency( dependabot )

kcherukupalli:
## What’s changing?

Update all dependency for the solution. 

## Why?

@nac-1146

## This PR has:

- [x ] **Commented source code** (_required on public classes and methods, and elsewhere as appropriate_)


components/src/Piipan.Components.Demo/wwwroot/components/DemoModalComponent.txt
```diff
@@ -19,11 +19,11 @@
 }
 
 
-<h2 class="usa-modal__heading" id="modal-1-heading">
+<h2 class="usa-modal__heading" id="DemoModalComponent-heading">
```
bgfoshay, line 22: Question: Not sure this file should be included in the updates?

dborgen2: This looks like a missed file when I made PR https://github.com/18F/piipan/pull/3591. I added unique IDs to the modals now that we had more than one modal, and this auto-generated file must never have gotten generated. Although it has nothing to do with this PR, it should be in the codebase so we should keep it.

bgfoshay: 👍Makes sense. Thanks!

bgfoshay (approved):
LGTM

PR #3663, opened 2022-08-15
Adding Unauthorized banner to 4 pages and View Permissions to Match Detail

dborgen2:
## What’s changing?

An unauthorized banner is getting added to the following pages when permissions are not correct:

- Search for Snap Participants (added when you are in charge of more than 1 state)
- Match Search (when your search does not return any values because it either doesn't exist or you're not authorized to see it)
- Match Detail (when your match does not exist or you are not authorized to see it)
- Match List (when you're not a national office user)

In addition, the Match Details page now has explicitly controlled view permissions as well as edit permissions.

Lastly, a system was put in place to easier pass common data to each page by using an AppInitiator that takes in AppData and injects it into each other razor component.

## Why?
Closes NAC-859 and NAC-658

## This PR has:


bgfoshay (approved):
LGTM

PR #3674, opened 2022-08-18
Removing Internal Details from ETL Status API Response

ccaprio97:
## What’s changing?

taking out ID and Publisher from ETL status api respose

## Why?

NAC-1063

## This PR has:


ccaprio97:
did some googling and this seemed like the only way something about it still worries me though because its on the class itself and not done after making the object

dborgen2:
Yea, I've been talking with Conner offline. JsonIgnore is the right thing to use to not pass variables when serialized by Newtonsoft, I wanted to make sure this wouldn't affect the database though, but it looks like it doesn't. We should be good here. Removing from the constructor should be unnecessary, so he's going to try adding it back in and making sure it still works. Assuming so, I can approve.

dborgen2 (approved):
👍

bgfoshay (changes requested):
Question: Can you add a test that serializes the object and verifies that the outputted json does not contain those properties?

bgfoshay (approved):
Thanks for adding the test! 

Nitpick: There might be a little bit of extra overhead in the test but it covers the change you made.

PR #3675, opened 2022-08-19
CSV Parsing does not emit PII on bad data

dborgen2:
## What’s changing?

When a CSV has a row with BadData, it would spill out all information on that row, including potentially PII information.  No longer will that happen, as we are explicitly creating a new exception when BadData is found.

In addition, ExceptionMessagesContainRawData gets set to false, just in case there is a different scenario that could show the raw data in an exception.

## Why?

Closes NAC-1105

## This PR has:


etl/src/Piipan.Etl/Piipan.Etl.Func.BulkUpload/Parsers/ParticipantStreamCsvParser.cs
```diff
@@ -183,7 +183,12 @@ public IEnumerable<IParticipant> Parse(Stream input)
             var config = new CsvConfiguration(CultureInfo.InvariantCulture)
             {
                 HasHeaderRecord = true,
-                TrimOptions = TrimOptions.Trim
+                TrimOptions = TrimOptions.Trim,
+                BadDataFound = (args) =>
+                {
+                    throw new InvalidDataException($"Error parsing the CSV. Bad data found on row {args.Context.Parser.Row}");
```
bgfoshay, line 189: Question: Would it be possible/helpful to also list the invalid column?

bgfoshay: yea, that's a good point

bgfoshay (approved):
LGTM

PR #3689, opened 2022-08-25
Reverting Diagnostic Setting name

ryangeorgeharvey:
## What’s changing?

Reverting Diagnostic Setting name from `stream-diagnostic-logs` to `stream-logs-to-event-hub` .

## Why?

Changing the Diagnostic Settings name for existing resources can cause IaC - ARM errors unless the previous Diagnostic Settings are removed. This results from the `RootManageSharedAccessKey` used across multiple Diagnostic Settings (old name and new name). Thus, the diagnostic setting will remain with the name `` rather than a more generic name like ``.

To create a new Diagnostic Setting for an existing resource using the `RootManageSharedAccessKey`, the old Diagnostic Setting must be removed. 

Normally, engineers would delete the Azure resource containing the Diagnostic Setting, not the Diagnostic Setting itself. However, when deleting resources from Azure, Azure does not always guarantee the Diagnostic Setting will be removed. For example, deleting a Storage Account will delete the Storage Account in Azure, but the underlying Diagnostic Settings will not always be removed. Engineers observed this behavior causing errors in both Commercial and Government Azure clouds.

I created a resource cleanup script to remove the old diagnostic settings before applying the new settings. It works well if all resources are deployed, the script is run, then the new IAC script is run. However, if a resource were previously deleted, the script could not delete the diagnostic settings, and the IAC errors described above would occur.

Rather than removing the old diagnostic settings, and causing headaches for the devs, I am reverting the diagnostic setting name change and adding a note in the ` log-streaming.md`.

## This PR has:


bgfoshay (approved):
👍

PR #3694, opened 2022-08-26
Branching Strategy ADR

bgfoshay:
## What’s changing?

Adding an ADR describing the desired branching strategy we'd like to use as NAC moves to production.

## Why?

NAC-1126 & NAC-1021

## This PR has:


dborgen2 (commented):
It looks good to me. I think it makes more sense for @ryangeorgeharvey to approve this PR, so I'll wait for him to give the green light.

The one thing I did want to note that I talked with Bill offline is that we'll need to update the configurations to kick off all the normal code coverage/unit tests/etc for these release branches that we would have for dev. And also the nightly tests.

PR #3736, opened 2022-09-01
Initial checkin for the new Dashboard UI changes

dborgen2:
## What’s changing?

- Added a new Table component to the common components library
- Added the ability to pass GET request query parameters easily to an API
- Adding a filter to the dashboard main page, which allows you to filter on any combination of start date, end date, state, and status of upload.
- Updating the paging to work correctly and dynamically show pages depending on which page we're on.
- Overall look/feel of the Dashboard metrics page.

## Why?

Closes NAC-1176, 1177, and 1179

## This PR has:


components/src/Piipan.Components/Forms/UsaForm.razor.cs
```diff
@@ -66,7 +73,7 @@ protected override async Task OnAfterRenderAsync(bool firstRender)
                     IsDirty = true;
                     StateHasChanged();
                 }
-                await JSRuntime.InvokeVoidAsync("piipan.utilities.registerFormValidation", Id, DotNetObjectReference.Create(this));
+                await formJavascriptReference.InvokeVoidAsync("RegisterFormValidation", Id, DotNetObjectReference.Create(this));
```
bgfoshay, line 76: Thought: it might be helpful to add a quick comment about why the RegisterFormValidation js method is being called

components/src/Piipan.Components/Forms/UsaInputDate.razor
```diff
@@ -3,11 +3,12 @@
 @inject IJSRuntime JSRuntime
 @code {
     InputDate<DateTime?> InputDateReference { get; set; }
+    IJSObjectReference inputDateJavascriptReference;
```
bgfoshay, line 6: Question: where is this getting set?

components/src/Piipan.Components/Forms/UsaInputDate.razor
```diff
@@ -3,11 +3,12 @@
...
     public override int? Width { get; set; } = 180;
 
     protected override async Task<List<string>> PreverificationChecks()
     {
-        bool hasBadInput = await JSRuntime.InvokeAsync<bool>("piipan.utilities.doesElementHaveInvalidInput", Id);
+        bool hasBadInput = await inputDateJavascriptReference.InvokeAsync<bool>("DoesElementHaveInvalidInput", Id);
```
bgfoshay, line 11: Question: is there any chance this line could be executed with inputDateJavascriptReference still set to null?

dborgen2: It's possible, but only if for some reason the file fails to download, or I guess if it's taking a very long time to download. The OnAfterRenderAsync will get fired off immediately after the component renders, which sets the javascript reference, which should occur well before the form is submitted.

This does have a small chance of being a problem, but I'm not really sure if anything needs to be done, and if so what it should be. Thoughts?

components/src/Piipan.Components/Forms/UsaForm.razor.cs
dborgen2: Sure, I can add that, no problem

components/src/Piipan.Components/Forms/UsaInputDate.razor
bgfoshay: That makes sense and I think it's fine for now. It's a pretty low probability scenario.

components/src/Piipan.Components/Tables/UsaTable.razor
```diff
@@ -0,0 +1,56 @@
...
+        <thead>
+            <tr>
+                @ChildContent
+            </tr>
+        </thead>
+        <tbody>
+            @foreach (var item in Data)
+            {
+                <tr @key="item">
+                    @for (int j = 0; j < Columns.Count; j++)
```
bgfoshay, line 38: Thought (non-blocking): Have you given any thought to how this approach would support column reordering, or column sorting?

components/src/Piipan.Components/Tables/UsaTableColumn.razor
```diff
@@ -0,0 +1,24 @@
+﻿@using System.Linq.Expressions
+@implements IDisposable
+@typeparam T
+
+@code {
+    [Parameter] public string Name { get; set; }
+    [Parameter] public Expression<Func<T, object>> Field { get; set; }
+
+    [Parameter] public RenderFragment<T> Template { get; set; }
+    [CascadingParameter] protected UsaTable<T> DataTable { get; set; }
```
bgfoshay, line 10: Question: I'm still looking through the rest of this PR, but at first glance this parameter seems somewhat circular. UsaTable refers to UsaTableColumn and UsaTableColumn refers to UsaTable?

components/src/Piipan.Components/Tables/UsaTable.razor
dborgen2: I think it shouldn't be too bad. Reordering would be easy... there could be a method similar to AddColumn and RemoveColumn that takes a column current index and their new index. We'd do Columns.RemoveAt(currentIndex) and Columns.Insert(newIndex, column). This method could get called during a drag/drop operation or something else.

Sorting would be easy as well, as long as the data that's being sorted is already in the table. If we need to make a database call to sort the entire data set that gets a little more complicated, but maybe not. We could probably do something like the Paging links, where when they click the column to sort it it navigates to the page with the sort parameter set (and it handles it on the server side)

components/tests/Piipan.Components.Tests/Input/UsaInputSSNTests.razor
```diff
@@ -529,7 +536,7 @@
         Component = componentFragment.FindComponent<UsaInputSSN>();
         inputElement = componentFragment.Find($".{InputClass}");
         JSInterop.SetupVoid("piipan.utilities.setCursorPosition", _ => true);
```
bgfoshay, line 538: Question: Could you have used the moduleInterop for this call as well?

components/src/Piipan.Components/Tables/UsaTable.razor
bgfoshay: 👍

components/tests/Piipan.Components.Tests/Input/UsaInputSSNTests.razor
dborgen2: Yea, this was a mistake. This should be moduleInterop... good catch

dashboard/src/Piipan.Dashboard.Client/Components/ParticipantUploads/ParticipantUploadsResults.razor
```diff
@@ -0,0 +1,127 @@
+﻿@using Piipan.Components.Tables
+@using Piipan.Dashboard.Client.Models
+@using Piipan.Metrics.Api
+@code {
```
bgfoshay, line 4: Question: There's a good deal of logic embedded in the column templates & pagination block within this file. While it's not necessarily complicated logic, I'm wondering if there could be any benefit to moving some of it up to methods within the code block? e.g. CapitalizeFirstLetterOfStatus() or ConvertToLocalDateTime() Would that make readability or testing any easier?

components/src/Piipan.Components/Tables/UsaTableColumn.razor
dborgen2: Discussed offline, should be OK for now but potential to make this CascadingParameter an interface in the future

dashboard/src/Piipan.Dashboard/wwwroot/sass/_uswds-theme-components.scss
```diff
@@ -7,7 +7,7 @@
 ========================================
 ========================================
 ----------------------------------------
-USWDS 2.10.1
+USWDS 2.10.0
```
bgfoshay, line 10: Question: Was the version drop intentional?

dashboard/tests/Piipan.Dashboard.Tests/BasePageTest.cs
```diff
@@ -0,0 +1,108 @@
...
+using Moq;
+using Moq.Language.Flow;
+using Piipan.Dashboard.Pages;
+using Piipan.Shared.Claims;
+using Piipan.States.Api;
+using Piipan.States.Api.Models;
+
+namespace Piipan.Dashboard.Tests
+{
+    public class BasePageTest
```
bgfoshay, line 21: Thought (non-blocking): the contextMock and OnPageHandlerExecutionAsync methods share the same code in BasePageTest.cs in Dashboard and QueryTool test projects. Will they always be the same and if so, can we consider pulling them out into common class somewhere?

dashboard/tests/Piipan.Dashboard.Tests/Components/ParticipantUploads/ParticipantUploadsFilterTests.razor
```diff
@@ -0,0 +1,171 @@
...
+        Assert.Null(Component.Instance.UploadRequest.Status);
+        Assert.Null(Component.Instance.UploadRequest.StartDate);
+        Assert.Null(Component.Instance.UploadRequest.EndDate);
+    }
+
+    /// <summary>
+    /// Verify the Quick Status buttons update the status and visually change
+    /// </summary>
+    [Fact]
+    public void QuickStatusButtons_ShouldChangeStatus()
```
bgfoshay, line 126: Question: Should we also consider a test where QuickStatus button is clicked twice? e.g. click Failed button two times and verify it turns the filter on then off?

query-tool/tests/Piipan.QueryTool.Tests/Helpers/ViewRenderer.cs
```diff
@@ -33,7 +33,7 @@ public class ViewRenderer
             _activator = activator;
         }
 
-        public async Task<(IView, string)> RenderView<TModel>(string name, TModel model) where TModel : PageModel
+        public async Task<(IView, string)> RenderView<TModel>(string name, TModel model)
```
bgfoshay, line 36: Question: Was the constraint removal intentional? What if a random object is provided as a parameter to a call to RenderView?

dborgen2: Yes, actually this whole RenderView function can probably be removed. We only use the RenderPages one, but I left it in there in case anyone needed it. The model should be able to be any type, not just a PageModel, because we are rendering views instead of pages here (it was originally a mistake)

dashboard/tests/Piipan.Dashboard.Tests/BasePageTest.cs
dborgen2: Yes, you're right. I definitely think we should add a common client library, as there have been a couple of instances where we have duplicated code. Maybe a tech debt story?

bgfoshay: Sounds good! I'll add one.

dashboard/tests/Piipan.Dashboard.Tests/Components/ParticipantUploads/ParticipantUploadsFilterTests.razor
dborgen2: Yep, good idea. I've added two new tests (one for failed, one for complete)

dashboard/src/Piipan.Dashboard/wwwroot/sass/_uswds-theme-components.scss
dborgen2: No, I've updated it back. We had to take a lot of styles from QueryTool, and it looks like it was one version behind.

dashboard/src/Piipan.Dashboard.Client/Components/ParticipantUploads/ParticipantUploadsResults.razor
dborgen2: This was a good callout. I created a new UsaPagination component (and UsaInputNumber, which was a side effect of creating this), both with their own unit tests.

I also created a DateTimeExtensions and StringExtensions class for the date and string logic, each with their own unit tests.

bgfoshay: 👍

query-tool/tests/Piipan.QueryTool.Tests/Helpers/ViewRenderer.cs
bgfoshay: 👍

bgfoshay (approved):
Changes look good. Thanks for addressing all of the comments.

PR #3742, opened 2022-09-05
Adding Daily Statistics to the Metrics page

dborgen2:
## What’s changing?

A Daily Statistics area was added to the metrics page, showing how many states have failed uploads today and how many states have successfully completed uploads today.

## Why?

_Link to specific issue (ex: Closes #123)_

## This PR has:


dashboard/src/Piipan.Dashboard/Pages/ParticipantUploads.cshtml.cs
```diff
@@ -63,6 +65,18 @@ public async Task OnGetAsync()
             }
         }
 
+        private async Task GetUploadStatistics()
+        {
+            var offset = TimeZoneInfo.Local.GetUtcOffset(DateTime.UtcNow);
+            ParticipantUploadStatisticsRequest statisticsRequest = new ParticipantUploadStatisticsRequest
+            {
+                StartDate = DateTime.Now.Date,
```
bgfoshay, line 73: Question: Is this right? Retrieving statistics when StartDate and EndDate are the same? Should start date be one day prior?

metrics/src/Piipan.Metrics/Piipan.Metrics.Core/DataAccessObjects/ParticipantUploadDao.cs
```diff
@@ -84,6 +84,64 @@ public async Task<Int64> GetUploadCount(ParticipantUploadRequestFilter filter)
...
+            }
+            return whereClause;
+        }
+
+        private object GenerateStatisticsQueryObject(ParticipantUploadStatisticsRequest request)
+        {
+            return new
+            {
+                dateFrom = request.StartDate?.AddHours(-1 * request.HoursOffset),
+                dateTo = request.EndDate?.AddDays(1).AddHours(-1 * request.HoursOffset),
```
bgfoshay, line 111: Question: Related to my previous comment on ParticipantUploads.cshtml.cs, I see that 1 day is being added to the EndDate here. Is the expectation that EndDate will always be selected automatically so we can safely assume we want to add a day?

dashboard/src/Piipan.Dashboard/Pages/ParticipantUploads.cshtml.cs
dborgen2: Yep, that should be right. If the start date and end date are the same, it will just grab the items from today which is what they wanted.

bgfoshay (commented):
I had a couple questions around EndDate usage/selection but for the most part this looks good. My other comment would be to add a smoke test for the new function, similar to test-metricsapi.bash script, and add a related comment to the testing section (section 10) of iac.md file. We're trying to get better about incorporating smoke tests so this would help.

metrics/src/Piipan.Metrics/Piipan.Metrics.Core/DataAccessObjects/ParticipantUploadDao.cs
dborgen2: So, this area isn't reliant on the filter. It will always (currently) be set to today's date.
The reason we add a day is because the database is also storing time (and we are accounting for time zone). So we add a day to the end date (which is set to midnight today, so adding a day will give us midnight tomorrow), and then we offset by our timezone and find everything less than that. This was the same logic that I did with the filter stuff too, but in this case we are just passing in today's date. However, it's flexible enough that if we want to in the future pass different dates this will work just fine.

bgfoshay: Ok, thanks for the explanation! That makes sense. It might be worth adding those notes as a comment for any future developers who come across the code, just so the logic is clear.

dborgen2: Sounds good, will do!

PR #3750, opened 2022-09-07
Adding Vulnerable Individual to Match List Results

ccaprio97:
## What’s changing?

Adding Vulnerable Individual to match list results, and only showing the other state if the user searching for the match is a state user

## Why?

NAC-1156

## This PR has:

## Anything else?
image
image


query-tool/src/Piipan.QueryTool.Client/Components/MatchResults.razor
```diff
@@ -44,6 +46,64 @@
...
+
+    private string MatchingStateTableResults(MatchResApiResponse result)
+    {
+        if (IsNationalOffice)
+        {
+            return string.Join(", ", result.Data.States.OrderBy(state => state).Select(state => StateHelper.GetStateName(state)));
+        }
+        else
+        {
+            foreach(string state in result.Data.States)
```
dborgen2, line 58: could simplify this whole else clause into one line by doing

`return result.Data.States.FirstOrDefault(n => n != LoggedInUsersState) ?? "";`

query-tool/src/Piipan.QueryTool.Client/Components/MatchResults.razor
```diff
@@ -44,6 +46,64 @@
...
+                    {
+
+                        true => "Yes",
+                        false => "No",
+                        null => "-"
+                    };
+                }
+            }
+        }
+        VulnerableIndividualTableResults  = VulnerableIndividualTableResults.Remove(VulnerableIndividualTableResults.Length - 1);
```
dborgen2, line 103: Wouldn't this take off the last character even on an eligibility worker scenario. Perhaps `VulnerableIndividualTableResults  = VulnerableIndividualTableResults.TrimEnd(',')` would be better

query-tool/src/Piipan.QueryTool.Client/Components/MatchForm.razor
```diff
@@ -6,6 +6,8 @@
     [Parameter] public MatchSearchRequest Query { get; set; }
     [Parameter] public ParameterBase<List<MatchResApiResponse>> QueryResult { get; set; }
     [Parameter] public string Token { get; set; }
+    [Parameter] public string LoggedInUsersState { get; set; }
+    [Parameter] public bool IsNationalOffice { get; set; }
```
dborgen2, line 10: I think we should take in States and Location instead of these two, so we account for regional users.

query-tool/src/Piipan.QueryTool.Client/Components/MatchResults.razor
```diff
@@ -10,6 +10,8 @@
 
 @code {
     [Parameter] public List<MatchResApiResponse> QueryResult { get; set; }
+    [Parameter] public string LoggedInUsersState { get; set; }
```
dborgen2, line 13: Again, taking in States and Location. I would set up a boolean flag for MultipleStatesView.

`private bool MultipleStateView = false;

    protected override void OnInitialized()
    {
        MultipleStateView = Location.Length != 2 || States?.Length != 1;
    }`

query-tool/src/Piipan.QueryTool.Client/Components/MatchResults.razor
```diff
@@ -53,15 +113,17 @@
         <thead>
             <tr>
                 <th scope="col">Match ID</th>
-                <th scope="col">Matching States</th>
+                <th scope="col">Matching State</th>
```
dborgen2, line 116: Label should change if MultipleStateView

query-tool/src/Piipan.QueryTool.Client/Components/MatchResults.razor
```diff
@@ -44,6 +46,64 @@
...
+                    return state;
+                }
+            }
+        }
+        return "";
+    }
+
+    private string VulnerableIndivudualTableResults(MatchResApiResponse result)
+    {
+        string VulnerableIndividualTableResults = "";
```
dborgen2, line 71: Function can be simplified using LINQ and the ValueToDisplayHelper's GetDisplayValue function.
Something like
`private string VulnerableIndivudualTableResults(MatchResApiResponse result)
    {
        if (MultipleStateView) {
            return string.Join(",", result.Data.Dispositions.OrderBy(d => d.State).Select(n => ValueToDisplayHelper.GetDisplayValue(n.VulnerableIndividual)));
        }
        else {
            return result.Data.Dispositions.FirstOrDefault(d => string.Equals(d.State, States[0], StringComparison.OrdinalIgnoreCase)).Select(n => ValueToDisplayHelper.GetDisplayValue(n.VulnerableIndividual));
        }
    }`

query-tool/src/Piipan.QueryTool.Client/Components/MatchResults.razor
dborgen2, line 58: Can be simplified using LINQ, if you want.
`return result.Data.States.FirstOrDefault(s => !string.Equals(s, States[0], StringComparison.OrdinalIgnoreCase));`

query-tool/src/Piipan.QueryTool.Client/Components/MatchResults.razor
ccaprio97: Wasnt sure about that. I know for this ticket its supposed to be matching state but this ticket i dont think was supposed to get into the whole double thing. I should ask andrew

dborgen2: Yea, I mean you're doing all of the other work for showing single/multiple states/VI info, so just changing the Title here would finish that work off instead of having to do it later. Unless there's something else too?

query-tool/src/Piipan.QueryTool.Client/Components/MatchResults.razor
```diff
@@ -44,6 +53,38 @@
...
+    private string VulnerableIndivudualTableResults(MatchResApiResponse result)
+    {
+        if (MultipleStateView)
+        {
+            return string.Join(",", result.Data.Dispositions.OrderBy(d => d.State).Select(n => ValueToDisplayHelper.GetDisplayValue(n.VulnerableIndividual)));
+        }
+        else
+        {
+            Disposition matchingStateDisposition = result.Data.Dispositions.FirstOrDefault(d => !string.Equals(d.State, States[0], StringComparison.OrdinalIgnoreCase));
+            return matchingStateDisposition.VulnerableIndividual switch
```
dborgen2, line 78: Nitpick: You can just return ValueToDisplayHelper.GetDisplayValue(matchingStateDisposition.VulnerableIndividual)

query-tool/src/Piipan.QueryTool.Client/Components/MatchResults.razor
```diff
@@ -44,6 +53,38 @@
...
+
+    private string MatchingStateTableResults(MatchResApiResponse result)
+    {
+        if (MultipleStateView)
+        {
+            return string.Join(", ", result.Data.States.OrderBy(state => state).Select(state => StateHelper.GetStateName(state)));
+        }
+        else
+        {
+            return result.Data.States.FirstOrDefault(s => !string.Equals(s, States[0], StringComparison.OrdinalIgnoreCase));
```
dborgen2, line 65: Sorry, I know I told you to do States[0], but just in case the states API fails and no states are brought back, we should probably do States?.FirstOrDefault() here and below. My bad!

query-tool/src/Piipan.QueryTool.Client/Components/MatchResults.razor
ccaprio97: i dont think that worked because disposition doesnt implement that or something like that

dborgen2: It should... let's talk offline when we both have a chance

query-tool/src/Piipan.QueryTool.Client/Components/MatchResults.razor
dborgen2: Nevermind... in both cases it only looks at States[0] if we're in single state view, which only happens when we have 1 state brought back from the API.

query-tool/src/Piipan.QueryTool.Client/Components/MatchResults.razor
ccaprio97: my bad. was talking about the .select from the last comment

query-tool/tests/Piipan.QueryTool.Tests/Components/MatchFormTests.cs
```diff
@@ -94,6 +96,37 @@ public void Verify_Results_Shown()
             Assert.False(queryForm.HasComponent<UsaAlertBox>());
         }
 
+        /// <summary>
+        /// Verify that when there are results from the query, the results area is shown with appropriate VI and state
+        /// </summary>
+        [Fact]
+        public void Verify_Results_Shown_StateUser()
```
dborgen2, line 103: This is a good start, but I feel like we should also have tests to validate the th text changes, and also (or in the same tests) validate that if there's only 1 state then only 1 value shows up in the states and VI field, and if there are multiple states both values show up, instead of just the Component existing. Honestly this was probably a miss with the original tests, and the values should have been verified, but I think it's more crucial now that we have some additional logic there.

ccaprio97: will do. kinda copied the other test. will add

dborgen2 (approved):
👍

PR #3758, opened 2022-09-08
Radio Button Component Fixes

ccaprio97:
## What’s changing?

Fixing radio buttons:
would have to double click if something else was selected
would need to tabs to get to radio buttons
couldnt go between radio buttons with arrow keys

## Why?

NAC-1242

## This PR has:

## Anything else?
image


query-tool/src/Piipan.QueryTool/package.json
```diff
@@ -14,6 +14,8 @@
     "autoprefixer": "^10.4.8",
     "cypress": "^10.4.0",
     "cypress-audit": "^1.1.0",
+    "cypress-plugin-tab": "^1.0.5",
```
dborgen2, line 17: You can probably drop this now and just use the realPress("Tab"). Even though I like tab() better, not as many packages to import this way

ccaprio97: I can do that but some of the tests are using this package. should i just change all the tests to use the realPress thing?

dborgen2: Yea, just to be consistent and minimalist here, I think so

dborgen2 (approved):
👍

components/src/Piipan.Components/Forms/TemporaryRadioGroup.cs
```diff
@@ -0,0 +1,22 @@
...
+
+namespace Piipan.Components.Forms
+{
+    /*
+     * This is a temporary workaround for a .NET bug. This class can be removed after updating to .NET 7 where the bug will be fixed.
+     * https://github.com/dotnet/aspnetcore/pull/40148
+     */
+    public class TemporaryRadioGroup<TValue> : InputRadioGroup<TValue>
+    {
+        private string? _name;
```
bgfoshay, line 11: Thought: can you add a comment or two documenting why this is necessary and what the bug is you're working around. If you have a link to the .Net 7 fix, that would be great as well

ccaprio97: added dan's comment

PR #3760, opened 2022-09-08
Fixing an issue where Cypress is trying to submit the form before it's fully loaded

dborgen2:
## What’s changing?

It looks like there's a race condition happening in Cypress where the form loads and is submitted before the form's submit listener is hooked up. I believe this was introduced as part of the last sprint where the form's Javascript code was changed to be a module import instead of a utility function (This was done so that Query Tool and Dashboard didn't have duplicate code. Instead it was done in Piipan.Components).

## Why?

Fixes nightly Cypress tests

## This PR has:


bgfoshay (approved):
👍

PR #3765, opened 2022-09-09
Updating tests to use a fake local time zone

dborgen2:
## What’s changing?

Dashboard tests were previously ignoring the datetime text because when people would run in different time zones the test results wouldn't be consistent. Switching to use a fake time zone that's always UTC-5, with a time zone abbreviation of TT.

## Why?

Fixes NAC-1060

## This PR has:


dashboard/tests/Piipan.Dashboard.Tests/Helpers/FakeLocalTimeZone.cs
```diff
@@ -0,0 +1,30 @@
...
+            typeof(TimeZoneInfo).AsDynamicType().s_cachedData._localTimeZone = timeZoneInfo;
+        }
+
+        public FakeLocalTimeZone(TimeZoneInfo timeZoneInfo)
+        {
+            _actualLocalTimeZoneInfo = TimeZoneInfo.Local;
+            SetLocalTimeZone(timeZoneInfo);
+        }
+
+        public void Dispose()
```
bgfoshay, line 25: Question: Should this be in a destructor also? What happens if Dispose is not called or someone uses it without a using statement?

dborgen2: Hm... looks like I learned something new today. For some reason I was under the impression that the finalizer automatically calls Dispose, but it doesn't look like it does. I'll add this.

bgfoshay: 👍

bgfoshay (approved):
LGTM

PR #3807, opened 2022-09-14
Match Creation Status Fix and State To Lower Bug Fix

ccaprio97:
## What’s changing?

Fixing matchCreation statuses
adding get state from header to return a toLower() result to avoid bugs

## Why?

NAC-1151

## This PR has:


match/tests/Piipan.Match.Core.Tests/Services/MatchEventServiceTests.cs
```diff
@@ -559,7 +559,7 @@ public async void Resolve_InsertsNewMatchIdIfMostRecentRecordIsClosed()
                 State = "ea",
                 SearchFrom = String.Empty,
                 SearchReason = null,
-                MatchCreation = "Already Existing Match",
+                MatchCreation = "New Created Match",
```
bgfoshay, line 562: Question: Should we be using the enum here?

bgfoshay (approved):
👍

PR #3811, opened 2022-09-15
Fixing Invalid Match Validations

ccaprio97:
## What’s changing?

Fixing a bug that greys out initial action when it shouldnt be
adding the correct validations and error messages to invalid match on match details page

## Why?

NAC-1243

## This PR has:


match/src/Piipan.Match/Piipan.Match.Core/Validators/AddEventRequestValidator.cs
```diff
@@ -56,6 +56,22 @@ public AddEventRequestValidator()
...
+                });
+
+                RuleFor(r => r.Data.OtherReasoningForInvalidMatch).Custom((val, context) =>
+                {
+                    if (string.IsNullOrEmpty(val) && context.InstanceToValidate.Data.InvalidMatchReason == "Other")
+                    {
+                        context.AddFailure("Reason for Other is required because Invalid Match Reason was set to Other");
+                    }
+                });
             });
```
dborgen2, line 75: Can we add rules here for if Invalid Match Reason is filled in but Invalid Match is false to throw an error? Similar if OtherReasoningForInvalidMatch is filled in but Invalid Match Reason isn't other or Invalid Match is false? Not possible in the UI, but could be possible via API

query-tool/src/Piipan.QueryTool.Client/Models/DispositionModel.cs
```diff
@@ -50,9 +51,16 @@ public string InitialActionTaken
 
         [Display(Name = "Invalid Match Reason")]
         [JsonProperty("invalid_match_reason")]
+        [UsaRequiredIf(
+            nameof(InvalidMatch), "True", "@@@ is required because Invalid Match has been selected"
```
dborgen2, line 55: I know you didn't do this, but since we're adding errors anyway, can we edit all instances of `@@@` to be `{ValidationFieldPlaceholder}` in this file? Might as well use the constant we have.

dborgen2 (approved):
LGTM

PR #3849, opened 2022-09-19
Changes for type 2 notifications

dborgen2:
## What’s changing?

- Added QueryToolUrl environment variable to Match Res function API, so it can be pulled into the template for the clickable link.
- Added a new notified_at column to match_res_events (defaulted to now() so that all existing data doesn't send out notifications). When a new match_res_event is added this gets set to null.
- Type 2 notifications were changed to execute on a 15 minute timer. Any matches that have events at least 30 minutes old that haven't been emailed out will be available for checking to see if they need to be emailed.
- Type 2 notification trigger changed slightly, to send out if VulnerableIndividual was previously true and it is currently false, to send out an email. If InvalidMatch was previously true and is currently false, send out an email. The other notification options still exist: If InvalidMatch or VulnerableIndividual got set to true, or if Initial Action/Final Disposition changes, send out an email.
- No matter whether a notification got send out or not, after all of these checks complete, set the notified_at column to now in the match_res_events table.
- Email templates changed for type 2 notifications, including showcasing all of the current values and which ones changed. Cross email client testing was done in Outlook, Gmail, Apple Mail, Windows Mail, and Outlook 365 webmail.
- Fixed type 1 notifications cross email client issues to be similar to the type 2 ones.
- Added the notifications projects to circleci and build.bash

## Why?

Closes NAC-1149

## This PR has:


match/src/Piipan.Match/Piipan.Match.Core/DataAccessObjects/MatchResEventDao.cs
```diff
@@ -104,6 +106,59 @@ END DESC
...
+        /// </summary>
+        /// <param name="matchId">The given match ID</param>
+        /// <param name="sortByAsc">Boolean indicating ascending sort order, defaults to true. Argument of false is descending order</param>
+        /// <returns>Task of IEnumerable of IMatchResEvents</returns>
+        public async Task<IEnumerable<IMatchResEvent>> GetEventsNotNotified()
+        {
+            // Bring back all of the matches where the last update was greater than 30 minutes ago.
+            // If it has been updated again in the last 30 minutes, skip it.
+            const string sql = @"
+                SELECT
```
bgfoshay, line 120: Question: is the nested sub-query needed? Could it be done in a single query?

SELECT <all fields desired> from match_res_events
                            WHERE notified_at is null AND
                            EXTRACT(EPOCH FROM (NOW() AT TIME ZONE 'utc' - inserted_at)) / 60 > 30)

dborgen2: Unfortunately no, because we need to grab all of the events for these matches (to compare what it was last time it was emailed vs now). But if you can think of a cleaner way to do this, I'm all for it, I realize it's a bit messy

match/src/Piipan.Match/Piipan.Match.Func.ResolutionApi/MatchResEventNotify.cs
```diff
@@ -0,0 +1,190 @@
...
+            _matchResEventDao = matchResEventDao;
+            _matchResAggregator = matchResAggregator;
+            _stateInfoDao = stateInfoDao;
+            _notificationService = notificationService;
+            _notificationRecordBuilder = notificationRecordBuilder;
+
+        }
+
+        [FunctionName("MatchResEventNotify")]
+        public async Task Run([TimerTrigger("0 */5 * * * *")] TimerInfo myTimer, ILogger log)
```
bgfoshay: Question: This is a timer for every 5 minutes, right? Is that the final requirement for this? At one point I heard every 15 minutes?

match/src/Piipan.Match/Piipan.Match.Func.ResolutionApi/MatchResEventNotify.cs
```diff
@@ -0,0 +1,190 @@
...
+                InitState = matchResRecord.Initiator,
+                MatchingState = matchResRecord.States.FirstOrDefault(n => !n.Equals(matchResRecord.Initiator, StringComparison.OrdinalIgnoreCase)),
+                CreatedAt = matchResRecord.CreatedAt,
+                Status = matchResRecord.Status,
+                InitStateDisposition = initiatingStateDisposition,
+                MatchingStateDisposition = matchingStateDisposition
+            };
+        }
+
+        public bool IsValidForNotification(Disposition beforeDisposition, Disposition afterDisposition, HashSet<string> dispositionUpdates)
```
bgfoshay, line 153: Nitpick: does this method need to be public?

match/src/Piipan.Match/Piipan.Match.Func.ResolutionApi/MatchResEventNotify.cs
```diff
@@ -0,0 +1,190 @@
...
+                (dispositionUpdates.Contains(nameof(Disposition.InitialActionAt)) || dispositionUpdates.Contains(nameof(Disposition.InitialActionTaken))) ||
+                (dispositionUpdates.Contains(nameof(Disposition.FinalDispositionDate)) || dispositionUpdates.Contains(nameof(Disposition.FinalDisposition))))
+            {
+                return true;
+            }
+            return false;
+        }
+
+
+        public DispositionUpdatesModel GetDispositionUpdates(DispositionModel dispositionModeBeforeUpdate, DispositionModel dispositionModel)
```
bgfoshay, line 166: Nitpick: does this method need to be public?

match/src/Piipan.Match/Piipan.Match.Func.ResolutionApi/MatchResEventNotify.cs
```diff
@@ -0,0 +1,190 @@
...
+            _matchResEventDao = matchResEventDao;
+            _matchResAggregator = matchResAggregator;
+            _stateInfoDao = stateInfoDao;
+            _notificationService = notificationService;
+            _notificationRecordBuilder = notificationRecordBuilder;
+
+        }
+
+        [FunctionName("MatchResEventNotify")]
+        public async Task Run([TimerTrigger("0 */15 * * * *")] TimerInfo myTimer, ILogger log)
```
bgfoshay, line 46: Thought: Would there be any benefit to splitting out some of the notification logic in this class into a separate service class of some kind?

match/tests/Piipan.Match.Core.IntegrationTests/MatchRecordDaoTests.cs
```diff
@@ -348,5 +348,187 @@ public async Task GetMatches_ReturnsNoRecordsIfNotFound()
                 Assert.Empty(results);
             }
         }
+
+        [Fact]
+        public async Task GetMatchesById_ReturnsMutlipleRecordsIfFound()
```
bgfoshay, line 353: Question/Thought: Could the data insertion/setup be generalized and reused across all these tests?

match/tests/Piipan.Match.Core.IntegrationTests/MatchResEventDaoTests.cs
```diff
@@ -376,5 +376,226 @@ public async void GetEventsByMatchIDs_ReturnsEmptyListWhenNoRecords()
...
+                    ActorState = "ea",
+                    Delta = "{}",
+                };
+
+                // Act
+                int id = await dao.AddEvent(record);
+                UpdateMatchResEventCreateDateTo1HourAgo(id);
+
+
+                // Assert - No results because the events were added within the last 30 minutes
```
bgfoshay, line 540: Question: should this say "// Assert - No results because **NO** events were added within the last 30 minutes"?

shared/src/Piipan.Shared.TestFixtures/MatchesDbFixture.cs
```diff
@@ -123,5 +123,18 @@ public void ClearMatchResEvents()
                 conn.Close();
             }
         }
+
+        public void UpdateMatchResEventCreateDateTo1HourAgo(int id)
```
bgfoshay, line 127: Thought: Could there be any benefit to generalizing this and adding an additional parameter for numHours or numMinutes (or some other unit)?

match/tests/Piipan.Match.Func.ResolutionApi.Tests/MatchResEventNotifyTests.cs
```diff
@@ -0,0 +1,519 @@
...
+                FinalDisposition = null,
+                FinalDispositionDate = null,
+                VulnerableIndividual = false
+            }
+        };
+
+        [Fact]
+        public async void ValidateDispositionUpdatesModeTest()
+        {
+            var beforeUpdate = new DispositionModel()
```
bgfoshay, line 183: Question: Could the BeforeDispositionUpdate and AfterDispositionUpdate member variables be used here instead?

match/src/Piipan.Match/Piipan.Match.Func.ResolutionApi/MatchResEventNotify.cs
bgfoshay: After looking through the PR a little more, it looks like these methods get accessed in tests. We might consider pulling them out into a helper class. It might provide better separation and also, might then allow us to test that class explicitly while mocking the helper in other tests.

notifications/src/Piipan.Notification.Common/Templates/DispositionEmail.cshtml
```diff
@@ -1,96 +1,250 @@
...
-    cursor: pointer;
-    display: inline-block;
-    font-weight: 700;
-    margin-right: .5rem;
-    padding: .75rem 1.25rem;
-    text-align: center;
-    text-decoration: none;
-    width: 100%;
+    string currentProperty;
+    string UpdatedHtml(bool initState, params string[] otherProperties)
```
bgfoshay, line 13: Question: Still looking through the PR so it may be covered already, but wanted to make sure this logic is tested somewhere?

match/tests/Piipan.Match.Core.IntegrationTests/MatchRecordDaoTests.cs
dborgen2: Yea, I think that's a good point. I'll make a Setup function that these tests can all use.

match/tests/Piipan.Match.Core.IntegrationTests/MatchResEventDaoTests.cs
dborgen2: Yea, the comment on this one and the one below it is wrong. Updating now

shared/src/Piipan.Shared.TestFixtures/MatchesDbFixture.cs
dborgen2: Yes, I mean it's only currently used for this test but might as well make it flexible. I'll have it take in a number of hours, defaulted to 1

match/tests/Piipan.Match.Func.ResolutionApi.Tests/MatchResEventNotifyTests.cs
dborgen2: Yea, good call. Just a couple small differences between the two, so they should be used.

match/src/Piipan.Match/Piipan.Match.Func.ResolutionApi/MatchResEventNotify.cs
dborgen2: Yea, I think the overall architecture here can be improved. Will work on it momentarily

notifications/src/Piipan.Notification.Common/Templates/DispositionEmail.cshtml
dborgen2: You're probably right. I'll look into it

match/tests/Piipan.Match.Core.Tests/Services/MatchResNotifyServiceTests.cs
```diff
@@ -0,0 +1,208 @@
...
+using Piipan.Match.Core.Services;
+using Piipan.Notification.Common.Models;
+using Piipan.Notifications.Core.Services;
+using Piipan.States.Core.DataAccessObjects;
+using Piipan.States.Core.Models;
+using Xunit;
+
+namespace Piipan.Match.Core.Tests.Services
+{
+    public class MatchResNotifyServiceTests
```
bgfoshay, line 15: Question: do we need a  test when neither state has valid updates?

dborgen2: Yes, good catch. Adding now

bgfoshay (approved):
Thanks for addressing all the comments I had with this one. I had a suggestion (non-blocking) to add one more test but other than that I think it looks good to go.

PR #3923, opened 2022-09-28
ADR for using Blazor as a Single Page Application

dborgen2:
## What’s changing?

After putting together a [Proof of Concept](https://github.com/18F/piipan/pull/3922), it is clear that we should be able to convert to a SPA relatively quickly, and without running into issues with EasyAuth.

It is recommended that we do this to avoid the downsides that we currently have with server-rendered pages:
1. Every page that we navigate to currently has to fetch all of common server variables, render the HTML code, and wire up Blazor WebAssembly. Because of this the application suffers unnecessary performance issues and could run much faster.
2. Blazor wasn't really meant to work like this and we have had to put in a couple of "hacks" to get the whole process to work. For example, every form hooks up into Javascript's onsubmit event, then when that event fires off it transitions back into Blazor to validate the form. Once the form is validated, it transitions back into Javascript to actually finish submitting the form. Another example is the Match Detail form when a server-side issue occurs in the application. The form should re-render with the typed values in them, even when they do not match the match detail's information. Because of this, we need to save both sets of values and pass them from server to client side when the server renders the page. With a SPA, we just send the values that we are trying to save to the API, and if the request succeeds we update the right side of the application. These issues as well as others we would encounter in the future make an odd developer experience that is prone to errors.
3. It will be much easier to use automated testing, because we will be able to mock API requests. Mocking an API request is much, much easier than mocking an entire HTML response, especially when said response is not requested with Javascript. Using Cypress/BUnit in this way will make our tests a lot more thorough.
4. We ran into an issue on the Dashboard where we did not know the user's timezone, because the page was rendered server side. If it was a SPA, we would know their timezone and could pass that information down to the API.

## This PR has:


docs/adr/0036-use-blazor-as-spa.md
```diff
@@ -0,0 +1,46 @@
...
+
+Proposed
+
+Supercedes [29. Use Blazor WebAssembly for Client Applications](0029-use-blazor-web-assembly.md)
+
+## Context
+
+Ever since the decision was made to use Blazor WebAssembly for client applications, it has fulfilled its purpose by giving us a higher amount of client-side interactivity while having a lower development effort. The decision at the time was to continue using server-side rendered pages to reduce the amount of backend code we'd have to change as well as the uncertainties about how to handle EasyAuth with a Single Page Application (SPA). It was not feasible for MVP to go this route, but now that we have some bandwidth we should consider switching to a SPA.
+
+The downside to the server-side rendered pages include the following:
```
bgfoshay, line 15: Nitpick: Overall I think this looks good. I know it's implied by the Context section, but something minor to consider is to explicitly list out benefits in the Consequences section i.e. better performance, simplified codebase that's easier to develop and maintain, easier and better testing.

bgfoshay (approved):
👍 Thanks for writing this up and for putting together the associated PR with the comprehensive POC.

PR #3982, opened 2022-10-14
Email Notification Support

bgfoshay:
## What’s changing?

Configure Notification App to use Mail Relay

## Why?

NAC-1114

## This PR has:


iac/env/tts/dev.bash
```diff
@@ -43,3 +43,5 @@ STORAGE_SKU="Standard_ZRS" # Standard Zone Redundant Storage
 
 # Azure API Management (APIM) SKU
 APIM_SKU="Developer"
+
+SMTP_SERVER=" smtp.gmail.com"
```
kcherukupalli, line 47: There is an extra space.

notifications/src/Piipan.Notifications.Core/Services/MailService.cs
```diff
@@ -0,0 +1,56 @@
...
+                    IsBodyHtml = true
+                };
+
+                string usdaSymbolImagePath = _usdaImageRetriever.RetrieveUsdaSymbolColorImage(); //Path to images/usda-symbol-color.png"
+                LinkedResource res = new LinkedResource(usdaSymbolImagePath); 
+                res.ContentId = "usda-img";
+                AlternateView alternateView = AlternateView.CreateAlternateViewFromString(emailModel.Body, null, MediaTypeNames.Text.Html);
+                alternateView.LinkedResources.Add(res);
+                mailMessage.AlternateViews.Add(alternateView);
+                mailMessage.To.Add(string.Join(',', emailModel.ToList));
```
dborgen2, line 45: Question: I know we're not currently putting anything in CC and BCC, but should we add that here just in case?

notifications/src/Piipan.Notifications.Func.Api/Api.cs
```diff
@@ -2,31 +2,47 @@
...
 
+        /// <summary>
+        /// Function that processes email requests placed in Azure queue.
+        /// </summary>
+        /// <param name="emailRequest">The message retrieved from the queue for the function to process</param>
+        /// <param name="log">handle to the function log</param>
+        /// <returns></returns>
         [FunctionName("NotificationRequestProcessor")]
-        public async Task Run([QueueTrigger("emailbucket", Connection = "")] string emailQueue, ILogger log)
+        public async Task Run([QueueTrigger("emailbucket", Connection = "")] string emailRequest, ILogger log)
```
dborgen2, line 29: Question: I'm not really sure how this Queue works. Does the email automatically get removed from the Queue when this trigger occurs? Do we need to put it back on the Queue and try again if there is an exception?

notifications/tests/Piipan.Notifications.Core.Tests/Services/MailServiceTest.cs
```diff
@@ -0,0 +1,130 @@
...
+                Subject = "Test Email From Email Service Tests"
+
+            };
+            var emailService = new MailService(smtpClient, retriever);
+            // Act
+            var result = await emailService.SendEmailAsync(emailModel, (ex) =>
+            { });
+
+            // Assert
+            Assert.False(result);
```
dborgen2, line 50: Nitpick (non-blocking): We should probably verify the logger logged the error correctly

notifications/tests/Piipan.Notifications.Core.Tests/Services/MailServiceTest.cs
```diff
@@ -0,0 +1,130 @@
...
+                     "test@email.example"
+                },
+                From = "from.test@email.example",
+                Subject = "Test Email From Email Service Tests"
+
+            };
+            var emailService = new MailService(smtpClient, retriever);
+            // Act
+            var result = await emailService.SendEmailAsync(emailModel, (ex) =>
+            { });
```
dborgen2, line 124: Nitpick (non-blocking): We should probably verify the logger logged correctly here as well

notifications/tests/Piipan.Notifications.Core.Tests/Services/MailServiceTest.cs
dborgen2: Actually, sorry... we don't log anything from here, do we? Maybe we should verify the callback is called.

notifications/tests/Piipan.Notifications.Core.Tests/Services/MailServiceTest.cs
dborgen2: Actually, sorry... we don't log anything from here, do we? Maybe we should verify the callback is called.

notifications/src/Piipan.Notifications.Core/Services/MailService.cs
```diff
@@ -0,0 +1,56 @@
...
+        /// <param name="errorCallback">Callback method to invoke if the email fails to send</param>
+        /// <returns>Returns True if email was sent to smtp successfully. Returns false if not.</returns>
+        public async Task<bool> SendEmailAsync(EmailModel emailModel, Action<Exception> errorCallback)
+        {
+            try
+            {
+                var mailMessage = new MailMessage()
+                {
+                    Subject = emailModel.Subject,
+                    From = new MailAddress(string.IsNullOrEmpty(emailModel.From) ? "noreply@nac.usda.gov" : emailModel.From),
```
dborgen2, line 35: Question: Do we need to update NotificationService.cs's PublishNotifications method to pass in a From address? Right now it looks like we're passing empty. I'm assuming the From address should get pulled from an environment variable or config setting?

iac/env/tts/dev.bash
dborgen2, line 47: Nitpick (non-blocking): I'm assuming this will fail in development environments. Is this OK? If so, should we add some kind of check before the email actually goes out to not send the email if we're in a development environment?

bgfoshay: Yes, this smtp will fail. It's not necessarily the environment itself that's bad. It's the difference in the smtp configuration vs the client smtp configuration.

I'll have to give some thought to this. Rather than checking the environment, it might be better to check for a specific SMTP value. If it's ok, I would prefer to move this off to a separate ticket.

notifications/src/Piipan.Notifications.Core/Services/MailService.cs
bgfoshay: Good catch. It is passing empty currently. This should be moved off to an environment variable or config setting. If ok, I will create a separate ticket to address this.

notifications/tests/Piipan.Notifications.Core.Tests/Services/MailServiceTest.cs
bgfoshay: Yes, I'll add this verification.

notifications/tests/Piipan.Notifications.Core.Tests/Services/MailServiceTest.cs
bgfoshay: Yes, I'll add this verification.

notifications/src/Piipan.Notifications.Func.Api/Api.cs
dborgen2: I see, so do we need to make sure to throw the exception in the callback? Or will just logging an error re-queue it?

iac/env/tts/dev.bash
dborgen2: Sounds good to me

notifications/src/Piipan.Notifications.Core/Services/MailService.cs
dborgen2: Sounds good to me

notifications/src/Piipan.Notifications.Func.Api/Api.cs
bgfoshay: Good point. We need to allow the exception to bubble up. I added a "throw" so that it propagates and I tested and verified that Azure goes into retry mode.

dborgen2 (approved):
LGTM, thanks for addressing the comments

PR #3988, opened 2022-10-18
Made match ID case insensitive

dborgen2:
## What’s changing?

There was an issue where when querying by Match ID or when attempting to navigate to a Match Detail page by ID that the match would not be found if it wasn't uppercase. This is due to PSQL being case sensitive, whereas something like SQL Server is case insensitive in where clauses.

The fix was simple: Just uppercase the Match ID before querying. Since the Match IDs are generated by us as upper case, we only need to upper case the queried value, not the one in the database.

## Why?

Closes NAC-1487

## This PR has:


match/src/Piipan.Match/Piipan.Match.Core/DataAccessObjects/MatchResEventDao.cs
```diff
@@ -96,7 +96,7 @@ END DESC
                 ;";
             var parameters = new
             {
-                MatchId = matchId,
+                MatchId = matchId.ToUpper(),
```
bgfoshay, line 99: Nitpick: might be worth a comment as to why ToUpper is called here

bgfoshay (approved):
LGTM

PR #3994, opened 2022-10-20
Fixing issues found during UX QA

dborgen2:
## What’s changing?

Fixed the following issues that came up during UX QA:

- Line 2: CUI Alert padding
- Line 3: VI column in results table
- Line 5 (partially): Screen reader. Only fixed the dropdown to sign out of the header, not the date pickers. The date pickers are an issue with Chromevox, not the app.
- Line 11: Lowercase "l" in "Last"
- Line 12: Date format in same line
- Line 14: Clear filters should submit the form
- Line 27: Tooltip screen reader issue
- Line 30: Text in success alert should be bold
- Line 35: 24px between each accordion
- Line 39: Hint text for reason for other text box
- Line 40: Duplicate error message should be an info box
- Line 42: Popping up the Close Match confirmation modal. No concurrency checks put in place yet.
- Line 49: Invalid reason is not showing up in the Match Information accordion

## Why?

NAC-1485, NAC-1486, NAC-1488, and other UX QA findings

## This PR has:


query-tool/src/Piipan.QueryTool.Client/Components/MatchDetail/MatchDetail.razor
```diff
@@ -92,18 +93,34 @@
     {
         return IsMatchValid() || IsMatchVulnerable();
     }
-    private Task SubmitForm(bool valid)
+    private Task<bool> SubmitForm(bool valid)
     {
+        var otherState = IsUserInInitiatingState() ? MatchingStateDisposition : InitiatingStateDisposition;
+        if ((!string.IsNullOrEmpty(otherState.FinalDisposition) && otherState.FinalDispositionDate != null)
```
bgfoshay, line 99: Nitpick/Question (non-blocking): Could this logic be pulled out into a simple function/method like IsMatchConsideredClosedByState(state)? This method could be called for both the Initiating and Matching state.

query-tool/src/Piipan.QueryTool/Pages/Match.cshtml.cs
```diff
@@ -232,6 +250,14 @@ public async Task<IActionResult> OnPost([FromRoute] string id)
                 }
 
                 Match = await _matchResolutionApi.GetMatch(id, IsNationalOffice ? "*" : Location);
+                if (matchBeforeSave?.Data.Status == MatchRecordStatus.Open && Match?.Data.Status == MatchRecordStatus.Closed)
```
bgfoshay, line 253: Question: Is there any chance it could reach this code with Request errors and then issue an alert saying the match has been closed?

bgfoshay (approved):
I had a couple minor non-blocking questions/comments but overall this looks good. Thanks for tackling so many of the UX QA findings so quickly.

query-tool/src/Piipan.QueryTool/Pages/Match.cshtml.cs
dborgen2: Hmm... good question. I don't think so, because if there were request errors then the save would have failed, so when re-fetching the status would still be open. But I'll add extra layer of protection here by adding `&& MatchDetailData.SaveSuccess`

query-tool/src/Piipan.QueryTool.Client/Components/MatchDetail/MatchDetail.razor
dborgen2: Unfortunately they aren't the same underlying type, so that wouldn't quite work. However, I do think putting this into it's own function just to make the SubmitForm method a bit cleaner would be a good idea

PR #4038, opened 2022-11-01
Changes after Litmus review

dborgen2:
## What’s changing?

After testing the emails with Litmus, a few accessibility issues came to light. They were updated in the CSHTML templates.

image

The only other problems that I saw when looking through the rendering was some email clients in dark mode had a tough time with the USDA logo. This one was fine (Office 365 Windows):
image

Whereas this one was not: (Office 365 Mac):
image

We may want to consider a different image.

As far as Litmus itself, it was useful to see how it rendered in all of the different clients, but whether or not it's worth the price is a discussion to be had. After large email template changes it may be a good idea to pay for a month, complete some testing, and then cancel.

I did not find a way to automatically test the renderings. I believe you could automatically test the accessibility checks, but it wasn't super intuitive and with the amount of times the emails change at this point in the development process I don't think it would be useful to only automatically test that.

Below you can see an example of the Litmus dashboard after sending an email:
image

## Why?

Closes NAC-1288

## This PR has:


bgfoshay (approved):
The fixes look good to merge. We should consider a separate ticket for utilizing Litmus. We should consider capturing your notes in an ADR and then make recommendation.

PR #4186, opened 2022-11-08
Email body is empty when Enable Email is off

kcherukupalli:
## What’s changing?

Fixed issue with Email body in the log when enable email is set to false.
Fixed issue with empty CC list.  This fix will also be applied for nac-1584.
## Why?

nac-1579
nac-1584

## This PR has:


notifications/src/Piipan.Notifications.Core/Services/MailService.cs
```diff
@@ -12,7 +12,7 @@ namespace Piipan.Notifications.Core.Services
     public class MailService : IMailService
     {
         private readonly SmtpClient smtpClient;
-        private readonly EmailDelivery _emailDelivery;
+        private readonly EmailDelivery _emailDelivery; //Defined in Configuration
```
bgfoshay, line 15: Nitpick: thanks for adding a comment! Maybe detail which configuration you're referring to? Environment variables?

notifications/src/Piipan.Notifications.Core/Services/MailService.cs
```diff
@@ -43,26 +43,23 @@ public async Task<bool> SendEmailAsync(EmailModel emailModel, ILogger logger, Ac
...
                 if (emailModel.ToCCList != null)
                 {
-                    mailMessage.CC.Add(string.Join(',', emailModel.ToCCList));
+                    var ccList = string.Join(',', emailModel.ToCCList.Where(s => !string.IsNullOrEmpty(s.Trim())));
+                    if (!string.IsNullOrEmpty(ccList))
+                        mailMessage.CC.Add(ccList);
                 }
 
-                if (!string.IsNullOrEmpty(_emailDelivery.EmailCc))
+                if (!string.IsNullOrEmpty(_emailDelivery.EmailCc.Trim()))
```
dborgen2, line 55: If EmailCc is null, .Trim() will cause an exception here. I think it can be removed and just keep the .Trim() below.
If you're wanting to make sure it's not null or white space, you can use IsNullOrWhitespace

notifications/src/Piipan.Notifications.Core/Services/MailService.cs
```diff
@@ -43,26 +43,23 @@ public async Task<bool> SendEmailAsync(EmailModel emailModel, ILogger logger, Ac
                 AlternateView alternateView = AlternateView.CreateAlternateViewFromString(emailModel.Body, null, MediaTypeNames.Text.Html);
                 alternateView.LinkedResources.Add(res);
                 mailMessage.AlternateViews.Add(alternateView);
-                mailMessage.To.Add(string.Join(',', emailModel.ToList));
+                mailMessage.To.Add(string.Join(',', emailModel.ToList.Where(s => !string.IsNullOrEmpty(s.Trim()))));
```
dborgen2, line 46: Same thing here. string.IsNullOrWhitespace is probably what you want here instead. If s is null this will throw an exception before calling .Trim()

notifications/src/Piipan.Notifications.Core/Extensions/MailExtensions.cs
```diff
@@ -1,12 +1,39 @@
...
     public static class MailExtensions
     {
         public static string ToStringExtended(this MailMessage mailMessage)
         {
-            return String.Format("Email To:{0} - Email CC: {1} -Email Subject: {2} - Email Body: {3}", mailMessage.To.ToString(), mailMessage.CC.ToString(), mailMessage.Subject, mailMessage.Body);
+            return String.Format("Email To:{0} - Email CC: {1} -Email Subject: {2} - Email Body: {3}", mailMessage.To.ToString(), mailMessage.CC.ToString(), mailMessage.Subject, GetAlternateViewText(mailMessage.AlternateViews[0]));
+        }
+        public static string GetAlternateViewText(AlternateView av)
+        {
+            using (MemoryStream s = new MemoryStream())
```
dborgen2, line 14: Nitpick: I know it's not possible in the current implementation, but if AlternateView is ever null or doesn't have any items in it there will be an exception passing the first AlternateView into this method. I would suggest passing mailMessage.AlternateViews (the whole array), and then in this method check to see if it's null or empty. If so, just return back and empty string (or take the message body?)

bgfoshay (approved):
👍 Thanks for turning this around so quickly and incorporating the logging changes

PR #4195, opened 2022-11-09
Match Detail UX QA fixes

dborgen2:
## What’s changing?

- When using a screen reader, it should say "email link" and not just "link"
- When I try to delete either IA & date and FD & date, it won't save those changes and update in the Match Information accordion
- phone should be formatted like: 123-456-7890 (the hyphens are missing)
- When saving Yes as Invalid, No as Invalid, and then going back to Yes, the previous selection was saved but it should not be
- When going from Benefits Denied (with auto filled current date) to Benefits Approved, it should not keep the current date and default to the date not filled in.

Also fixed a few bugs related to save match event details that I found during testing:
- If a state ever marked a match as invalid or put a final disposition on it, and the new state saved as invalid or put a final disposition on it, the match would be closed, even if the other state ended up removing the invalid/final disposition.
- As a patch request, the AddEventApi endpoint should only overwrite the properties to null if they actually were null in the json. If they were missing from the json, the properties should not be set to null. That way when the ResAggregator runs it will take the earlier value of the property instead of null if it at one point was saved with a value. (this is related to item 2 above)
-  When determining if a saved value is a duplicate action, it should compare the whole object as it would be saved to the currently saved object. Before it was simply comparing the latest match event to the new one, which would be a problem if the latest match event had different properties defined but still resolved to null, or if the latest match event was for a different state.

## Why?

Closes NAC-1483

## This PR has:


match/src/Piipan.Match/Piipan.Match.Api/Models/Resolution/Disposition.cs
```diff
@@ -9,30 +9,26 @@ namespace Piipan.Match.Api.Models.Resolution
     public class Disposition
     {
         [Display(Name = "Inital Action Date")]
-        [JsonProperty("initial_action_at",
-            NullValueHandling = NullValueHandling.Ignore)]
+        [JsonProperty("initial_action_at")]
```
bgfoshay, line 12: Question: Is it ok to remove the NullValueHandling attribute? Is this class reused for different purposes where the attribute is necessary?

match/src/Piipan.Match/Piipan.Match.Func.ResolutionApi/AddEventApi.cs
```diff
@@ -98,36 +99,38 @@ public class AddEventApi : BaseApi
...
-                }
                 // insert event
                 var newEvent = new MatchResEventDbo()
                 {
                     MatchId = matchId,
                     ActorState = state,
                     Actor = req.Headers["From"].ToString() ?? UserActor,
-                    Delta = JsonConvert.SerializeObject(reqObj.Data)
+                    Delta = rawRequest["data"]?.ToString(),
+                    InsertedAt = DateTime.MaxValue
```
bgfoshay, line 109: Question: Is MaxValue really what we want here?

match/tests/Piipan.Match.Core.Tests/Parsers/AddEventRequestParserTests.cs
```diff
@@ -0,0 +1,97 @@
...
+            Assert.NotNull(request.ParsedRequest);
+            Assert.Equal(true, request.ParsedRequest.Data.VulnerableIndividual);
+            Assert.Null(request.ParsedRequest.Data.FinalDispositionDate);
+
+            // Assert raw object contains properties passed into API, and does not contain properties not passed in
+            Assert.False((request.RawRequest["data"] as Newtonsoft.Json.Linq.JObject).ContainsKey("final_disposition_date"));
+            Assert.True((request.RawRequest["data"] as Newtonsoft.Json.Linq.JObject).ContainsKey("vulnerable_individual"));
+        }
+
+        private Stream BuildStream(string s)
```
bgfoshay, line 84: Question: Is there a potential for memory leaks here? Do you need to close the stream everywhere where you're using it?

query-tool/src/Piipan.QueryTool.Client/Components/MatchDetail/MatchDetail.razor
```diff
@@ -177,6 +177,11 @@
             </UsaAlertBox>
         </div>
     }
+    @* TODO: On hold for now, but this logic will be necessary in the near future.
```
bgfoshay, line 180: Question: If we're not doing this work now, do we need to capture it in a JIRA ticket?

query-tool/src/Piipan.QueryTool.Client/Components/MatchDetail/StateInformation.razor
```diff
@@ -19,9 +19,34 @@
 
             if (info.Value != null && info.DisplayType == DisplayInfoType.Email)
             {
-                format = @<a href="mailto:@(displayValue)">@displayValue</a>
+                format = @<a href="mailto:@(displayValue)" aria-label="Email @displayValue">@displayValue</a>
         ;
             }
+            if (info.Value != null && info.DisplayType == DisplayInfoType.Phone)
+            {
+                string phoneFormattedValue = "";
```
bgfoshay, line 27: Question: Should we consider pulling this phone number formatting code into a method that might get reused for phone number displays in other areas of the application?

bgfoshay (commented):
Overall looks great. Just a couple nitpicks/questions.

match/src/Piipan.Match/Piipan.Match.Func.ResolutionApi/AddEventApi.cs
dborgen2: This doesn't get stored into the database, it just is used for sorting in the MatchDetailsAggregator. So it should be fine, I just wanted to make sure it was the latest entry, but open to suggestions.

match/tests/Piipan.Match.Core.Tests/Parsers/AddEventRequestParserTests.cs
dborgen2: Yea, I think you're probably right. I was taking a pattern from a different Test file, but these streams should be closed/disposed after use. I'll update them

query-tool/src/Piipan.QueryTool.Client/Components/MatchDetail/MatchDetail.razor
dborgen2: Yea, we do. I mentioned it to Andrew, but I'll follow up with him to see if he created anything for it yet or if it still needs to be done.

query-tool/src/Piipan.QueryTool.Client/Components/MatchDetail/StateInformation.razor
dborgen2: Yea, probably wouldn't hurt. I can do that

match/src/Piipan.Match/Piipan.Match.Func.ResolutionApi/AddEventApi.cs
bgfoshay: Does the following line from the method not persist it in the db?

var successfulEventAdd = await _matchResEventDao.AddEvent(newEvent);

bgfoshay (approved):
👍 The fixes look good!

PR #4439, opened 2022-12-13
Initial changes for the database restructuring using a common Database manager

dborgen2:
## What’s changing?

A new DatabaseManager was added to help performance issues with the database. All of the changes are reliant on the two new classes: SingleDatabaseManager and MultipleDatabaseManager. Both of them take in the type of the database they are connecting to, either ParticipantsDbManager or CoreDbManager, which correspond to the Participants and Core databases respectively. The IDatabaseManager is registered as a Singleton in the startup to ensure that there aren't more DataSources than necessary floating around the app.

MultipleDatabaseManager should be used when a single DatabaseManager should be handling the connections to multiple databases within the database server. For example, if your app needs to connect to multiple state databases, you would use a MultipleDatabaseManager, passing in the list of databases to the constructor.

In both cases, a SemaphoreSlim is used to make sure only one thread is creating the DataSource at any given time. Once the DataSource is created, any other thread can use it, as long as the Azure Token does not need to be updated. If it does, the old DataSource is disposed, and the new one is created with the new Azure Token.

PerformQuery is used to actually perform a query on the connection. We can expand this method to add retries, logging, and more. Since it is currently impossible to get the DataSource or ConnectionString apart from calling PerformQuery (since PerformQuery is the only public method/property), this ensures that all logic flows through there.

In addition, many tests were updated to use this new architecture.

## Why?

Closes NAC-1727

## This PR has:

- [x] **JIRA ticket(s)** (_Must be listed in PR request and branch name should follow convention NAC-123-description_)<description>

- [x] **Security Related Changes** (_Security Tab questions have been answered on JIRA ticket(s)_)

- [ ] **Documented API changes** (_required on public classes and methods, and elsewhere as appropriate_)
    - [ ] External API documentation - OpenApi docs and Generated Docs (e.g. https://github.com/18F/piipan/tree/dev/docs/openapi) 
    - [ ] Internal API documentation (e.g. https://github.com/18F/piipan/tree/dev/metrics/docs/openapi)   

    - [x] Have IaC scripts run successfully against an existing deployed environment?
    - [ ] Have IaC scripts run successfully against a newly deployed environment?
    - [ ] Do the IaC changes require deleting existing resources?

## Anything else?

_Add other details that are helpful for review here_

## Code Review Checklist:

- [x] Automated checks all pass? (code builds successfully, tests pass, test coverage meets/exceeds threshold, etc.) 
- [x] Have all requirements and Acceptance Criteria been met?
- **If there are code changes**
    - [x] Is there no unnecessary complexity? Code is readable, maintainable, etc.
    - [x] Do the changes follow SOLID principles?
    - [x] Are the changes robust? Concurrency considered? Performance? Security (e.g. no sql injections)?
    - [x] Are newly added dependencies needed and appropriate? Are dependency files (e.g. packages.json & packages.lock.json) updated?
    - [ ] For any new projects/solutions, have they been added to code coverage files?
    - [x] Public classes & methods are documented    
- **If there are API changes**
    - [ ] Does API make sense? Useful and concise?
    - [ ] Are existing API interfaces/contracts intact? i.e. no breaking changes
    - [ ] Are API tests included?
    - [ ] Are API responses appropriate and documented? (200s, 201, 404s, etc.)
    - [ ] Are internals kept private and not leaked into the API?
    - [ ] External API documentation - OpenApi docs and Generated Docs 
    - [ ] Internal API documentation    
- **If there are IAC changes**
    - [ ] IAC Scripts are readable & maintainable
    - [ ] IAC Scripts are documented
    - [ ] IAC scripts pass required checks
    - [ ] IAC scripts were run successfully by author
- **General Logging Questions**
    - [ ] Do new capabilities (or changed capabilities) include logging?
    - [ ] Is PII not logged?
- **General Error Handling Questions**
    - [x] Has exception handling been added for new capabilities (or changed capabilities)?
    - [x] Are errors/exceptions properly communicated to end-users? Were there requirements for user facing error messages?
    - [x] Are errors/exceptions logged?
    - [x] Is PII kept out of error/exception logging?
- **If there are Developer Documentation Changes**
    - [ ] General Project documentation (under piipan\docs. e.g. iac.md, architecture.md, etc.)
    - [ ] Github documentation additions/changes (Readmes)
- **General automated test questions**
    - [x] End-to-End Tests provided (if code implements new user story)
    - [x] Are tests sufficient?
    - [x] Corner cases considered?
    - [x] Are performance tests needed?


match/src/Piipan.Match/Piipan.Match.Core/Services/MatchEventService.cs
```diff
@@ -190,6 +190,7 @@ private async Task<ParticipantMatch> DetermineIfExistingMatchIsOpen(IParticipant
         {
             var latest = existingRecords.OrderBy(r => r.CreatedAt).Last();
 
+            // TODO: Think about adding a "current status" column to the match record so that we don't need to grab all of the events.
```
bgfoshay, line 193: Question: For Match Resolution Events, each record reflects a single event/change, rather than the entire snapshot of the match details at a given point in time. Does a "current status" column still help when there are potentially e.g. 5 events that represent the current status of a match?

match/tests/Piipan.Match.Core.IntegrationTests/MatchDaoTests.cs
```diff
@@ -34,6 +35,11 @@ private IDbConnectionFactory<CollaborationDb> DbConnFactory()
             return factory.Object;
         }
 
+        private IDatabaseManager<CoreDbManager> DbManager()
```
bgfoshay, line 38: Nitpick/Question: Is the DbConnFactory method still needed? I assume no and we can pull it out in a subsequent PR but I was just curious.

shared/src/Piipan.Shared/Database/BaseDatabaseManager.cs
```diff
@@ -0,0 +1,90 @@
...
+        /// Perform a query against the specified database.
+        /// </summary>
+        /// <typeparam name="T">The type to be returned by the function</typeparam>
+        /// <param name="database">The database to perform a query against</param>
+        /// <param name="query">A callback which executes the specified query with a connection string provided by the data source.</param>
+        /// <param name="cancellationToken">Provides a way to cancel the transaction</param>
+        /// <returns></returns>
+        public virtual async Task<T> PerformQuery<T>(Func<NpgsqlConnection, Task<T>> query, string database = null, CancellationToken cancellationToken = default)
+        {
+            // TODO: Perform retries, throttle, log. All queries should flow through here.
```
bgfoshay, line 85: Praise: thanks for calling this retry spot out. This will save some time for whoever works the database retry ticket.

match/tests/Piipan.Match.Core.IntegrationTests/MatchDaoTests.cs
dborgen2: Yep, that's right. All of that DbConnFactory stuff can be removed, but didn't want to do it as part of this PR

match/src/Piipan.Match/Piipan.Match.Core/Services/MatchEventService.cs
dborgen2: I think it would. Right now we use the MatchDetailsAggregator to take in all of the events to figure out the current match details. We also use this same aggregator when populating the metrics match table. I think it would be nice to have an entire table or something dedicated to the current match so we don't have to aggregate it, but at the very least having the latest status would save us from having to do this call here

shared/src/Piipan.Shared/Database/MultipleDatabaseManager.cs
```diff
@@ -0,0 +1,102 @@
...
+using Microsoft.Azure.Services.AppAuthentication;
+using Microsoft.Extensions.Logging;
+using Npgsql;
+
+namespace Piipan.Shared.Database
+{
+    /// <summary>
+    /// The MultipleDatabaseManager handles connections and queries to a set of databases.
+    /// </summary>
+    public class MultipleDatabaseManager<DbType> : BaseDatabaseManager<DbType>
```
bgfoshay, line 17: Question: Do we need to make this class implement IDisposable and dispose of any open datasources when called?

shared/src/Piipan.Shared/Database/MultipleDatabaseManager.cs
```diff
@@ -0,0 +1,102 @@
...
+    {
+
+        Dictionary<string, (NpgsqlDataSource dataSource, string azureToken)> DataSources = new();
+        Dictionary<string, SemaphoreSlim> ThreadLockers = new();
```
bgfoshay, line 21: Thought/Question: Would there be any advantage to using a ConcurrentDictionary for either/both of these variables?

dborgen2: Yea, it's a good point. I originally had them as ConcurrentDictionaries, but after doing some research it seems like the main reason to use ConcurrentDictionaries is for dealing with changes to the dictionary asynchronously. Since the writing to ThreadLockers is only done in the constructor, and the writing to the DataSource is within the ThreadLocker.Wait(), there should never be multiple threads writing at the same time, and the reads would be guaranteed to have the latest value. So I guess it still might not be a bad idea to use a ConcurrentDictionary, I think it's not needed in this case, but let me know if you think otherwise.

shared/src/Piipan.Shared/Database/MultipleDatabaseManager.cs
dborgen2: Yea, probably a good idea. I'll make that change

shared/src/Piipan.Shared/Database/MultipleDatabaseManager.cs
bgfoshay: The more I think about it, I think it's ok. As it stands now, if the class is used outside of dependency injection, it wouldn't be a singleton, so these variables wouldn't have any contention. So even though we're not using it that way, it'd be safe. And for our current usage of it, it should only ever have one instance of it being constructed ever so it should be ok.

shared/src/Piipan.Shared/Database/MultipleDatabaseManager.cs
```diff
@@ -0,0 +1,102 @@
...
+        /// <summary>
+        /// Gets the connection to the specified database.
+        /// </summary>
+        /// <param name="database">The database to retrieve a connection string for</param>
+        /// <param name="cancellationToken">Provides a way to cancel the transaction</param>
+        /// <returns></returns>
+        protected override async Task<NpgsqlConnection> GetConnectionAsync(string database, CancellationToken cancellationToken = default)
+        {
+            // If our password depends on the Azure token, grab the latest one. If it's different than our current one, update our password and create a new data source.
+            if (ShouldFetchAzureToken)
```
bgfoshay, line 52: Question: This flag/method is based off the ConnectionString property. Upon re-fetching a token, do we need to reset the ConnectionString property with pw equal to the latest token?

shared/tests/Piipan.Shared.Tests/Database/MultipleDatabaseManagerTests.cs
```diff
@@ -0,0 +1,210 @@
...
+
+            // Act
+            await PerformDummyQuery(databaseManager);
+
+            // Assert
+            azureTokenProvider.Verify(m =>
+                m.GetAccessTokenAsync(AzurePgConnectionFactory<MockType>.GovermentId, null, default(CancellationToken)), Times.Once);
+
+            // Tear down
+            ClearEnvironmentVariables();
```
bgfoshay, line 97: Question: Does this need to be in a finally block?

shared/src/Piipan.Shared/Database/MultipleDatabaseManager.cs
dborgen2: Although it wouldn't hurt anything, I think leaving the ConnectionString alone is fine. We substitute the new Azure Token on line 79 so it should always use the new token.

shared/tests/Piipan.Shared.Tests/Database/MultipleDatabaseManagerTests.cs
```diff
@@ -0,0 +1,210 @@
...
+        }
+
+        [Fact]
+        public async void PerformQuery_DoesNotFetchToken_WhenNotNeeded()
```
bgfoshay, line 101: Question/Suggestion: Should we consider a test where a token refresh is needed?

shared/tests/Piipan.Shared.Tests/Database/SingleDatabaseManagerTests.cs
```diff
@@ -0,0 +1,180 @@
...
+
+            // Act
+            await PerformDummyQuery(databaseManager);
+
+            // Assert
+            azureTokenProvider.Verify(m =>
+                m.GetAccessTokenAsync(AzurePgConnectionFactory<MockType>.GovermentId, null, default(CancellationToken)), Times.Once);
+
+            // Tear down
+            ClearEnvironmentVariables();
```
bgfoshay, line 96: Question: Same question as above regarding a finally block?

shared/tests/Piipan.Shared.Tests/Database/MultipleDatabaseManagerTests.cs
dborgen2: Yea, probably should. If there was an exception that would cause this method not to get hit, that means the test would have failed, and so we would see that failure in the report. But if this happened the other tests might fail because the environment variable didn't get cleared out, which might cause confusion. I'll update it to be in a try/finally.

states/src/Piipan.States/Piipan.States.Core/DataAccessObjects/StateInfoDao.cs
```diff
@@ -25,15 +27,15 @@ public async Task<IState> GetStateById(string id)
         {
             try
             {
-                using (var connection = await _dbConnectionFactory.Build())
+                return await _databaseManager.PerformQuery(async connection =>
```
bgfoshay, line 30: Question: This pattern is being extensively and I just wanted to double check whether the connection should/does get disposed when the code is implemented this way?

dborgen2: Yep, the connection gets disposed inside PerformQuery. Actually it gets disposed asynchronously (await using, line 86 of BaseDatabaseManager), which should be a tiny performance boost over the previous way of disposing which was synchronous.

shared/tests/Piipan.Shared.Tests/Database/SingleDatabaseManagerTests.cs
dborgen2: Yep, will fix

shared/tests/Piipan.Shared.Tests/Database/MultipleDatabaseManagerTests.cs
dborgen2: Yes, that's a good idea. I'll add a test

bgfoshay (approved):
Great work on this one! I know a lot of coding and testing went into it but I think it'll definitely improve our database performance.

